PR #4, opened 2023-09-29
SDK: Force login if input parameters change.

manugoyal:
It can be confusing to users if they change some of their login information, that running login/init/whatever doesn't actually log them in again, unless they explicitly specify force_login.

Tested: I can remove `force_login=True` in the braintrust internal unittests, which use the SDK to create a new org for each test.

Fixes BRA-403


ankrgyl (approved):
Nice!

PR #6, opened 2023-10-03
Modifications to support tracing.

manugoyal:


py/src/braintrust/logger.py
```diff
@@ -601,105 +686,48 @@ def __init__(
...
-        `prompt`, example's `id`, or anything else that would be useful to slice/dice later. The values in `metadata` can be any
-        JSON-serializable type, but its keys must be strings.
-        :param id: (Optional) a unique identifier for the event. If you don't provide one, Braintrust will generate one for you.
-        :param dataset_record_id: (Optional) the id of the dataset record that this event is associated with. This field is required if and only if the
-        experiment is associated with a dataset.
-        :param inputs: (Deprecated) the same as `input` (will be removed in a future version)
+        Log a single event to the experiment. The event will be batched and
+        uploaded behind the scenes.
+
+        :param **event: Data to be logged. See `braintrust.log` for full details.
```
ankrgyl, line 698: maybe switch the documentation, i.e. keep it here and delete in `braintrust.log()`? I think the latter is rarely (if ever) used

ankrgyl:
Two requests:

* Can you look into context vars in this change?
* Do you want to do the JS changes in this PR or a separate one?

manugoyal:
> Two requests:
> 
> * Can you look into context vars in this change?
> * Do you want to do the JS changes in this PR or a separate one?

Yeah this is currently here just to align with the backend. I will round it out after completing the backend.

py/src/braintrust/logger.py
```diff
@@ -614,105 +699,48 @@ def __init__(
...
+        detailed trace information beyond the scope of a single log event. We
+        recommend using spans within the python ContextManager framework (`with
+        experiment.create_span() as span`) to ensure they are terminated upon
+        completion.
+
+        :param name: The name of the span.
+        :param span_attrs: Optional additional attributes to attach to the span,
+        such as a type name.
+        :param start: Optional start time of the span, as a timestamp in seconds.
+        :param **event: Data to be logged. See `braintrust.log` for full details.
```
ankrgyl, line 732: same comment on `braintrust.log`

py/src/braintrust/logger.py
```diff
@@ -19,13 +23,13 @@
...
 from .gitutil import get_past_n_ancestors, get_repo_status
-from .util import SerializableDataClass, encode_uri_component, response_raise_for_status
+from .util import SerializableDataClass, encode_uri_component, get_caller_location, response_raise_for_status
 
 
 class BraintrustState:
     def __init__(self):
-        self.current_project = None
         self.current_experiment = None
+        self.current_span = contextvars.ContextVar("braintrust_current_span", default=None)
```
ankrgyl, line 32: we should probably return a no-op span here so ppl dont need to do gymnastics in their code to check if this is `None` or not

ankrgyl: Also, should we clear this when a span ends?

py/src/braintrust/logger.py
```diff
@@ -624,95 +810,71 @@ def log(
...
-                if score < 0 or score > 1:
-                    raise ValueError(f"score ({score}) values must be between 0 and 1")
-
-        if metadata:
-            if not isinstance(metadata, dict):
-                raise ValueError("metadata must be a dictionary")
-            for key in metadata.keys():
-                if not isinstance(key, str):
-                    raise ValueError("metadata keys must be strings")
+    def start_span(self, name="root", span_attributes={}, start_time=None, set_current=False, **event):
```
ankrgyl, line 852: can we make `set_current=True` the default :)

py/src/braintrust/logger.py
```diff
@@ -513,15 +551,83 @@ def summarize(summarize_scores=True, comparison_experiment_id=None):
...
+def current_experiment():
+    """Returns the global current experiment (set by `braintrust.init`). Returns
+    None if no experiment has been initialized.
+    """
+
+    with _state.current_experiment.get() as current_experiment_list:
+        return current_experiment_list[0]
+
+
+def current_span():
```
ankrgyl, line 575: should probably type annotate this

py/src/braintrust/framework.py
```diff
@@ -219,8 +219,11 @@ def Eval(
             loop = None
 
         async def run_to_completion():
-            results, summary = await run_evaluator(experiment, evaluator, 0, [])
-            report_evaluator_result(name, results, summary, True)
+            try:
+                results, summary = await run_evaluator(experiment, evaluator, 0, [])
+                report_evaluator_result(name, results, summary, True)
+            finally:
+                experiment.end()
```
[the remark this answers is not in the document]
manugoyal: isn't experiment created above on line 213? When would it be None?

py/src/braintrust/framework.py
```diff
@@ -327,49 +330,52 @@ async def run_evaluator_task(datum):
...
-
-        if experiment and not error:
-            experiment.log(
-                input=datum.input,
-                metadata=metadata,
-                expected=datum.expected,
-                output=output,
-                scores=scores,
-            )
+        with experiment.start_span("eval", input=datum.input, expected=datum.expected) as span:
```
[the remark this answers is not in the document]
manugoyal: Yeah I think we need to make the `current_experiment` a context var that is async safe. Then we should be able to wrap `run_evaluator_task` in the `@traced` decorator. We can also have `braintrust.current_span()` return the `self._current_experiment` if there is no span. I.e. `braintrust.current_span() = braintrust.current_span() or braintrust.current_experiment() or NOOP_EXPERIMENT_SPAN`.

py/src/braintrust/logger.py
manugoyal: Yes :)

py/src/braintrust/logger.py
```diff
@@ -298,6 +370,8 @@ def init_dataset(
     """
     Create a new dataset in a specified project. If the project does not exist, it will be created.
 
+    Remember to end your dataset when it is finished by calling `Dataset.end`. You may also wrap the dataset in a with block (`with braintrust.init_dataset(...) as dataset`) to ensure it is terminated at the end of the block.
```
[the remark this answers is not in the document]
manugoyal: For now it just flushes the log. If we start supporting spans (or have a notion of `current_dataset`) with datasets, it will do span cleanup stuff.

py/src/braintrust/logger.py
```diff
@@ -766,6 +929,189 @@ def summarize(self, summarize_scores=True, comparison_experiment_id=None):
             scores=score_summary,
         )
 
+    def end(self):
```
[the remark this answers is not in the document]
manugoyal: Will rename to `close`.

py/src/braintrust/logger.py
manugoyal: Will do.

PR #7, opened 2023-10-04
Ensure we do not log more than the maximum API gateway can handle

ankrgyl:
Since the logger writes batches of records to the API, we need to ensure we do not send more than the 10 MB limit of API gateway. This change simply predicts the batch length by serializing JSON values before sending them off.

There is a more efficient version of this change that does a streaming serialization of JSON values (and concatenates them together with `,`), but I think we can explore that later if needed.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -137,6 +137,10 @@ class ModelWrapper:
         return self.data[name]
 
 
+# 10 MB (https://docs.aws.amazon.com/apigateway/latest/developerguide/limits.html)
+MAX_REQUEST_SIZE = 10 * 1024 * 1024
+
+
 class _LogThread:
     def __init__(self, name=None):
         self.thread = threading.Thread(target=self._publisher, daemon=True)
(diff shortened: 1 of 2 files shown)


manugoyal (approved):
lgtm. Would it be possible to avoid double-serializing the item, by storing the json-stringified `item` into `curr` and sending that over?

py/src/braintrust/logger.py
```diff
@@ -197,14 +201,33 @@ def flush(self, initial_items=None, batch_size=100):
...
+                if len(curr) > 0:
+                    response_raise_for_status(conn.post("/logs", data=construct_json_array(curr)))
 
             if len(items) < batch_size:
                 break
 
             items.clear()
 
 
+def construct_json_array(items):
```
manugoyal, line 227: nit: maybe for uniformity define this right under `MAX_REQUEST_SIZE` as is done in js?

py/src/braintrust/logger.py
```diff
@@ -197,14 +201,33 @@ def flush(self, initial_items=None, batch_size=100):
                     break
 
             if len(items) > 0:
-                conn.post_json("logs", items)
+                # Construct batches of items that do not exceed the max request size divided by 2
```
manugoyal, line 204: I am implementing exactly the same thing in the AWS lambda publisher. My implementation looks like this:

```
    def flush(self, initial_items=None, batch_size=100):
        initial_items = initial_items or []
        initial_items.reverse()
        items_len = 0
        while True:
            while len(items) < batch_size and items_len < MAX_PUBLISH_SIZE / 2:
                if len(initial_items) > 0:
                    items.append(initial_items.pop())
                else:
                    try:
                        items.append(self.queue.get_nowait())
                    except queue.Empty:
                        break
                items_len += len(json.dumps(items[-1]))
            if len(items) == 0:
                break
        # Do publish
```

While true that we will exceed `MAX_PUBLISH_SIZE / 2` by one element, I feel like it's okay if we're okay with publishing single elements which are larger than the publish size.

manugoyal (approved):

PR #10, opened 2023-10-11
Allow reading a previous (base) experiment as a dataset while running an eval

ankrgyl:
This enables the "hill" climbing workflow.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -880,6 +880,22 @@ class Experiment(ModelWrapper):
             event=event,
         )
 
+    def base_experiment_dataset(self, output_field="auto"):
+        """
+        Get the base experiment's dataset.
+
+        :param output_field: The field to use as the output. If "auto" (the default), will use the expected field if it exists, otherwise the output field.
+
+        :returns: `ExperimentDataset`
+        """
+        if self.base_exp_id is None:
+            return None
+
+        return ExperimentDataset(
+            experiment_id=self.base_exp_id,
+            output_field=output_field,
+        )
+
     def summarize(self, summarize_scores=True, comparison_experiment_id=None):
         """
         Summarize the experiment, including the scores (compared to the closest reference experiment) and metadata.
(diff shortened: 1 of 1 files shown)


py/src/braintrust/logger.py
```diff
@@ -880,6 +880,35 @@ def start_span(self, name="root", span_attributes={}, start_time=None, set_curre
...
+
+        if base_exp_id is None:
+            raise ValueError("No base experiment found")
+
+        return ExperimentDataset(
+            experiment_id=base_exp_id,
+            output_field=output_field,
+        )
+
+    def _fetch_base_exp_id(self):
```
manugoyal, line 905: nit: should this be called `_fetch_base_exp`?

py/src/braintrust/logger.py
```diff
@@ -1121,6 +1147,65 @@ def __exit__(self, type, value, callback):
         self.end()
 
 
+class _ObjectLoader:
+    def __init__(self, object_type: str, object_id: id, version: "str | int" = None):
```
manugoyal, line 1151: should the type of version be `Union[str, int, NoneType]`?

py/src/braintrust/logger.py
```diff
@@ -1121,6 +1147,65 @@ def __exit__(self, type, value, callback):
...
+            except (ValueError, AssertionError):
+                raise ValueError(f"version ({version}) must be a positive integer")
+
+    def fetch(self):
+        self._check_not_finished()
+
+        for record in self.fetched_data:
+            yield {
+                "id": record.get("id"),
+                "input": json.loads(record.get("input") or "null"),
```
manugoyal, line 1172: Why not `None` instead of a "null" string? Also the `or` part is a bit scary, because it will apply to any "zero-valued" python object. If we only expect the value of `record.get("input")` to be None or a non-zero-valued python object (I don't think empty string is valid json?), then it's fine, but at that point, may as well write `record.get("input", "null")` (or just leave it as `None`?).

py/src/braintrust/logger.py
```diff
@@ -1121,6 +1147,65 @@ def __exit__(self, type, value, callback):
...
+                "output": json.loads(record.get("output") or "null"),
+                "metadata": json.loads(record.get("metadata") or "null"),
+            }
+
+        self._clear_cache()
+
+    @property
```
manugoyal, line 1179: Stylistically I'm a bit hesitant to introduce properties that actually do a lot of work behind the scenes. Conventionally "member access" is considered trivial.

py/src/braintrust/logger.py
```diff
@@ -1121,6 +1147,65 @@ def __exit__(self, type, value, callback):
...
+
+        for record in self.fetched_data:
+            yield {
+                "id": record.get("id"),
+                "input": json.loads(record.get("input") or "null"),
+                "output": json.loads(record.get("output") or "null"),
+                "metadata": json.loads(record.get("metadata") or "null"),
+            }
+
+        self._clear_cache()
```
manugoyal, line 1177: I understand this is just moving code around, but I find the two methods `fetch` and `fetched_data` a bit confusing. They both yield similar outputs (not exactly the same), and one clears the cache while the other doesn't. What's the rationale?

py/src/braintrust/logger.py
```diff
@@ -1121,6 +1147,65 @@ def __exit__(self, type, value, callback):
...
+        if self._pinned_version is not None:
+            return self._pinned_version
+        else:
+            return max([int(record.get(TRANSACTION_ID_FIELD, 0)) for record in self.fetched_data] or [0])
+
+    def _check_not_finished(self):
+        if self.finished:
+            raise RuntimeError(f"Cannot invoke method on finished {self.object_type}")
+
+
```
manugoyal, line 1208: Are we missing a method to actually set `finished = True`. Or otherwise is there any need to have a `finished` thing for an object which just reads data?

py/src/braintrust/logger.py
```diff
@@ -1372,6 +1421,58 @@ def __str__(self):
         )
 
 
+class ExperimentDataset:
```
manugoyal, line 1424: Would be cool to have a base class (i.e. an interface) for this and `Dataset` to implement. I imagine we'd have a matching interface on the JS side. Similar to the `Span` abstract base class.

py/src/braintrust/logger.py
```diff
@@ -1372,6 +1421,58 @@ def __str__(self):
...
+        # You can also iterate over the dataset directly.
+        for record in dataset:
+            print(record)
+        ```
+
+        :returns: An iterator over the records in the dataset.
+        """
+        for record in self.fetcher.fetch():
+            if self.output_field == "auto":
+                output = record.get("expected") or record.get("output")
```
manugoyal, line 1454: Don't we need to add an `expected` field to `ObjectLoader.fetch` to make this work?

manugoyal (commented):
generally lgtm. Would be great to add unit tests and the accompanying JS implementation.

PR #12, opened 2023-10-13
JS tracing implementation.

manugoyal:
Also tweaked the python implementation in accordance among other small cleanups.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -1,17 +1,171 @@
 import axios, { AxiosInstance } from "axios";
 import { v4 as uuidv4 } from "uuid";
 
-import iso from "./isomorph";
+import iso, { IsoAsyncLocalStorage, CallerLocation } from "./isomorph";
+import { runFinally } from "./util";
 
-interface BraintrustState {
-  current_project: Project | null;
-  current_experiment: Experiment | null;
+export type SetCurrentArg = { setCurrent?: boolean };
+
+export type StartSpanArgs = {
+  name: string;
+  spanAttributes?: Record<any, any>;
+  startTime?: number;
+  event?: ExperimentLogPartialArgs & Partial<IdField>;
+};
+
+export type StartSpanOptionalNameArgs = Omit<StartSpanArgs, "name"> &
+  Partial<Pick<StartSpanArgs, "name">>;
+
+/**
+ * A Span encapsulates logged data and metrics for a unit of work. This interface is shared by all span implementations.
+ *
+ * We suggest using one of the various `startSpan` methods, instead of creating Spans directly. See `Span.startSpan` for full details.
+ */
+export interface Span {
+  // Row ID of the span.
+  id: string;
+
+  // Span ID of the span. This is used to link spans together.
+  span_id: string;
+
+  // Span ID of the root span in the full trace.
+  root_span_id: string;
+
+  /**
+   * Incrementally update the current span with new data. The event will be batched and uploaded behind the scenes.
+   *
+   * @param event: Data to be logged. See `Experiment.log` for full details.
+   */
+  log(event: ExperimentLogPartialArgs): void;
+
+  /**
+   * Create a new span. This is useful if you want to log more detailed trace information beyond the scope of a single log event. Data logged over several calls to `Span.log` will be merged into one logical row.
+   *
+   * We recommend running spans within a callback (using `startSpanWithinCallback`) to automatically mark them as current and ensure they are terminated. If you wish to start a span outside a callback, be sure to terminate it with `span.end()`.
+   *
+   * @param args.name The name of the span.
+   * @param args.span_attributes Optional additional attributes to attach to the span, such as a type name.
+   * @param args.start_time Optional start time of the span, as a timestamp in seconds.
+   * @param args.event Data to be logged. See `Experiment.log` for full details.
+   * @returns The newly-created `Span`
+   */
+  startSpan(args: StartSpanArgs): Span;
+
+  /**
+   * Wrapper over `Span.startSpan`, which passes the initialized `Span` it to the given callback and ends it afterwards. See `Span.startSpan` for full details.
+   *
+   * @param args.setCurrent If true (the default), the span will be marked as the currently-active span for the duration of the callback. Equivalent to calling `braintrust.withCurrent(span, callback)`.
+   */
+  startSpanWithCallback<R>(
+    args: StartSpanArgs & SetCurrentArg,
+    callback: (span: Span) => R
+  ): R;
+
+  /**
+   * Terminate the span. Returns the end time logged to the row's metrics. After calling end, you may not invoke any further methods on the span object, except for the property accessors.
+   *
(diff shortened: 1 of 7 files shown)


js/src/logger.ts
```diff
@@ -367,21 +515,54 @@ export async function init(
...
+/**
+ * Wrapper over `braintrust.init`, which passes the initialized `Experiment` it to the given callback and closes it afterwards. See `braintrust.init` for full details.
+ *
+ * @param options.setCurrent If true (default), set the currently-active experiment to the newly-created one. Equivalent to calling `braintrust.withCurrent(experiment, callback)`.
+ */
+export async function initWithCallback<R>(
+  project: string,
+  options: Readonly<InitOptions & SetCurrentArg> = {},
+  callback: (experiment: Experiment) => R
+): Promise<R> {
```
ankrgyl, line 537: why not just make the callback an optional 3rd argument to `init`, and do this if it's passed in?

js/src/util.ts
```diff
@@ -0,0 +1,29 @@
+// Given a callback, runs the callback as a "finally" component. If
+// `callback` returns a Promise, `finallyF` is chained to the promise. Otherwise,
+// it is run afterwards synchronously.
+export function runFinally<R>(f: () => R, finallyF: () => void): R {
+  let runSyncCleanup = true;
+  try {
+    const ret = f();
+    // There's no "blessed" way to detect whether an object is a
+    // Promise-like object in javascript. So we use duck typing.
```
ankrgyl, line 9: In `framework.ts` we do:

```
 if (dataResult instanceof Promise) {
    data = await dataResult;
  } else {
    data = dataResult;
  }
```

does this not work?

and if not, should we change those places too?

js/src/logger.ts
manugoyal: I was having trouble figuring out how to do that, because it changes the return type of the function from `Experiment` to `Promise<R>`. And I wasn't having much luck with function overloading in typescript.

js/src/util.ts
manugoyal: I was just scared off by the stack overflow crowd suggesting there are other kinds of "thenable" objects in other promise libraries: https://stackoverflow.com/questions/27746304/how-to-check-if-an-object-is-a-promise. Unclear if we need to really care about that.

js/src/logger.ts
ankrgyl: Ah ok let’s leave as is

py/src/braintrust/util.py
```diff
@@ -47,5 +47,9 @@ def get_caller_location():
         if this_dir is None:
             this_dir = os.path.dirname(caller.filename)
         if os.path.dirname(caller.filename) != this_dir:
-            return dict(caller_filename=caller.filename, caller_lineno=caller.lineno)
-    return {}
+            return dict(
+                caller_function_name=caller.function,
+                caller_file_name=caller.filename,
+                caller_line_num=caller.lineno,
```
ankrgyl, line 53: I don't think we shold rename these -- we already have people who have logged them, and we'll be in a permanently inconsistent state. They're hardcoded to be ignored in the trace UI implementation right now.

js/src/logger.ts
```diff
@@ -1,18 +1,172 @@
...
+   *
+   * We recommend running spans within a callback (using `startSpanWithinCallback`) to automatically mark them as current and ensure they are terminated. If you wish to start a span outside a callback, be sure to terminate it with `span.end()`.
+   *
+   * @param args.name The name of the span.
+   * @param args.span_attributes Optional additional attributes to attach to the span, such as a type name.
+   * @param args.start_time Optional start time of the span, as a timestamp in seconds.
+   * @param args.event Data to be logged. See `Experiment.log` for full details.
+   * @returns The newly-created `Span`
+   */
+  startSpan(args: StartSpanArgs): Span;
```
ankrgyl, line 52: I think it's more consistent (and ergonomic) to make this `startSpan(name: string, args?: StartSpanArgs)`

js/src/logger.ts
```diff
@@ -1,18 +1,172 @@
...
+
+  /**
+   * Wrapper over `Span.startSpan`, which passes the initialized `Span` it to the given callback and ends it afterwards. See `Span.startSpan` for full details.
+   *
+   * @param args.setCurrent If true (the default), the span will be marked as the currently-active span for the duration of the callback. Equivalent to calling `braintrust.withCurrent(span, callback)`.
+   */
+  startSpanWithCallback<R>(
```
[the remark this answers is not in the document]
manugoyal: yeah I think traced makes sense

js/src/util.ts
manugoyal: I'll just use `instanceof Promise` for consistency.

js/src/framework.ts
```diff
@@ -145,74 +152,104 @@ export async function runEvaluator(
...
+                  ...resultRest
+                } = result;
+                currentSpan().log({
+                  output: resultRest,
+                  metadata: resultMetadata,
+                });
+                return result;
+              },
+              {
+                name: score.name || "<anonymous>",
```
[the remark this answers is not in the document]
ankrgyl: Screenshot 2023-10-16 at 4 59 26 PM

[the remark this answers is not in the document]
ankrgyl: We should also probably make `anonymous` be something like `scorer_<idx>`

js/src/framework.ts
```diff
@@ -145,74 +152,104 @@ export async function runEvaluator(
...
-        output = outputResult;
-      }
+    const callback = async () => {
+      try {
+        const meta = (o: Record<string, unknown>) =>
+          (metadata = { ...metadata, ...o });
+
+        await traced(
+          async () => {
+            const outputResult = evaluator.task(datum.input, { meta });
```
ankrgyl, line 165: i think it's worth passing the span explicitly in here still

ankrgyl (commented):
looks great!

js/src/framework.ts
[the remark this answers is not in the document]
manugoyal: I can make the scorer name nicer. But I do think `score.name` works for named functions (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name). E.g.
Screenshot 2023-10-16 at 7 58 37 PM

py/src/braintrust/util.py
```diff
@@ -47,5 +47,9 @@ def get_caller_location():
         if this_dir is None:
             this_dir = os.path.dirname(caller.filename)
         if os.path.dirname(caller.filename) != this_dir:
-            return dict(caller_filename=caller.filename, caller_lineno=caller.lineno)
-    return {}
+            return dict(
+                caller_functionname=caller.function,
+                caller_filename=caller.filename,
+                caller_lineno=caller.lineno,
+            )
```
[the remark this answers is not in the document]
manugoyal: Separate. Filed https://linear.app/braintrustdata/issue/BRA-516/move-caller-out-of-metrics-and-into-another-field-like-context

PR #16, opened 2023-10-21
Python SDK: Concurrent logging.

manugoyal:
Tested with existing unit tests. Noticed a significant improvement in the `fake_experiment.py` benchmark (14s to 4s for 1000 rows), though a lot of the win there is probably coming from merging span rows, rather than the concurrent logging. Now we will want to implement some sort of "distributed logging" functionality to actually exercise the row merging on the lambda side.


py/src/braintrust/logger.py
```diff
@@ -334,14 +347,23 @@ def __init__(self, name=None):
...
         self._start()
         for event in args:
-            self.queue.put(event)
-        self.queue_filled_event.set()
+            try:
+                self.queue.put_nowait(event)
+            except queue.Full:
+                # Notify consumers to start draining the queue.
+                with self.queue_filled_condition:
+                    self.queue_filled_condition.notify()
```
ankrgyl, line 364: do we lose the event in this case?

py/src/braintrust/logger.py
```diff
@@ -368,14 +392,24 @@ def flush(self, batch_size=100):
         # We cannot have multiple threads flushing in parallel, because the
         # order of published elements would be undefined.
         with self.flush_lock:
+            # Drain the queue.
+            all_items = []
+            try:
+                while True:
+                    all_items.append(self.queue.get_nowait())
```
ankrgyl, line 399: this is a scary piece of code. i assume we rely on the fact that the function will either return something or throw an exception. i'd still feel more comfortable if we capped/batched it somehow, e.g. to `num_threads*batch_size`?

py/src/braintrust/merge_row_batch.py
```diff
@@ -0,0 +1,45 @@
+# Mirrors the implementation of merge_row_batch in api/chalicelib/util.py.
```
ankrgyl, line 1: maybe remove references to `api` etc since they're inaccessible from this codebase?

py/src/braintrust/util.py
```diff
@@ -53,3 +56,20 @@ def get_caller_location():
                 caller_lineno=caller.lineno,
             )
     return None
+
+
+# TODO(manu): Share with chalicelib.
```
ankrgyl, line 61: remove?

py/src/braintrust/logger.py
manugoyal: Good catch!

py/src/braintrust/logger.py
manugoyal: Why don't I cap it to `self.queue.qsize()` called at the beginning. That way we don't get into some weird game of pulling things off the queue on one thread and pushing them on another forever.

ankrgyl: 👍

py/src/braintrust/logger.py
```diff
@@ -334,14 +347,25 @@ def __init__(self, name=None):
...
-            self.queue.put(event)
-        self.queue_filled_event.set()
+            while True:
+                try:
+                    self.queue.put_nowait(event)
+                    break
+                except queue.Full:
+                    # Notify consumers to start draining the queue.
+                    with self.queue_filled_condition:
+                        self.queue_filled_condition.notify()
```
ankrgyl, line 366: this seems a little dicey -- if i'm reading this correctly, we'll spin on one CPU until the queue clears up a bit. Is that correct?

ankrgyl (approved):
lgtm feel free to land but happy to discuss the spinning comment a bit further

py/src/braintrust/logger.py
manugoyal: Yeah that's true. Maybe it makes sense to block until the queue clears up. I will adjust.

PR #18, opened 2023-10-23
Remove axios

ankrgyl:
Switch to using (isomorphic) fetch. This requires node >= 17, but we haven't tested on earlier versions (and doubt we support them).

With this change, the SDK should work just fine in edge runtimes (https://github.com/axios/axios/issues/5523)

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -1,4 +1,5 @@
-import axios, { AxiosInstance, AxiosError } from "axios";
+/// <reference lib="dom" />
+
 import { v4 as uuidv4 } from "uuid";
 
 import iso, { IsoAsyncLocalStorage, CallerLocation } from "./isomorph";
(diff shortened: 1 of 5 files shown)


js/src/logger.ts
```diff
@@ -289,40 +314,71 @@ class HTTPConnection {
...
-    this.session = iso.makeAxios({ headers });
   }
 
-  async get(path: string, params: unknown | undefined = undefined) {
-    return await this.session!.get(_urljoin(this.base_url, path), { params });
+  async get(
+    path: string,
+    params: Record<string, string | undefined> | undefined = undefined
+  ) {
+    const url = _urljoin(this.base_url, path);
```
manugoyal, line 327: alternatively:
```
const url = new URL(_urljoin(...));
url.search = new URLSearchParams(...);
await fetch(url)
```

manugoyal (approved):

js/src/logger.ts
ankrgyl: Cool

PR #23, opened 2023-10-24
Split expected into its own type

ankrgyl:
`Output` and `Expected` are often, but not always, the same type. This change allows them to be different. As a user, if you want to enforce them to be the same, you can simply specify their types in the `Eval<...>(...)` definition as such.


js/src/framework.ts
```diff
@@ -15,15 +15,15 @@ import pluralize from "pluralize";
 
 export type Metadata = Record<string, unknown>;
 
-export interface EvalCase<Input, Output> {
+export interface EvalCase<Input, Output, Expected> {
```
manugoyal, line 18: Do we need the `Output` template param here if it isn't even part of the type? Should it only need to be part of `EvalScorerArgs`?

ankrgyl: we do not. fixed

manugoyal (approved):

PR #24, opened 2023-10-25
Minor tweaks to SDKs

manugoyal:
Python:
    - `braintrust.start_span` should have an optional name, as in the JS
      version.

Javascript:
    - Mirror the atexit flushing in the python SDK in JS.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -493,6 +493,15 @@ class LogThread {
   private active_flush: Promise<string[]> = Promise.resolve([]);
   private active_flush_resolved = true;
 
+  constructor() {
+    // Note that this will not run for explicit termination events, such as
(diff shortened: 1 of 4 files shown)


js/src/logger.ts
```diff
@@ -493,6 +493,15 @@ class LogThread {
   private active_flush: Promise<string[]> = Promise.resolve([]);
   private active_flush_resolved = true;
 
+  constructor() {
+    // Note that this will not run for explicit termination events, such as
+    // calls to `process.exit()` or uncaught exceptions. Thus it is a
+    // "best-effort" flush.
+    iso.processOn("beforeExit", async () => {
```
ankrgyl, line 500: should we use this in the `UnterminatedObjectsHandler` too?

ankrgyl (approved):
lgtm

PR #25, opened 2023-10-25
Add metadata field to eval definition.

manugoyal:
Sometimes it's useful to specify additional metadata for the experiment, such as the name of the experiment. We also add support for running multiple evals with the same project and different experiment names.

### DIFF
--- a/js/src/framework.ts
+++ b/js/src/framework.ts
@@ -2,6 +2,7 @@ import chalk from "chalk";
 import {
   Experiment,
   ExperimentSummary,
+  InitOptions,
   Span,
   currentSpan,
   noopSpan,
@@ -43,17 +44,29 @@ export type EvalScorer<Input, Output> =
   | ((args: EvalScorerArgs<Input, Output>) => Score)
   | ((args: EvalScorerArgs<Input, Output>) => Promise<Score>);
 
+export type EvalMetadata = {
+  experimentName?: string;
+};
+
+export function evalMetadataToInitOptions(
+  metadata: EvalMetadata | undefined
+): InitOptions {
+  return { experiment: metadata?.experimentName };
+}
+
 /**
  * An evaluator is a collection of functions that can be used to evaluate a model.
  * It consists of:
  * - `data`, a function that returns a list of inputs, expected outputs, and metadata
  * - `task`, a function that takes an input and returns an output
  * - `scores`, a set of functions that take an input, output, and expected value and return a score
+ * - `metadata`, optional additional metadata for the eval definition, such as experiment name.
  */
 export interface Evaluator<Input, Output> {
   data: EvalData<Input, Output>;
   task: EvalTask<Input, Output>;
   scores: EvalScorer<Input, Output>[];
+  metadata?: EvalMetadata;
 }
 
 export type EvaluatorDef<Input, Output> = {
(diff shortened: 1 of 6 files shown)


js/src/framework.ts
```diff
@@ -75,26 +97,36 @@ export async function Eval<Input, Output>(
...
 ): Promise<void | ExperimentSummary> {
-  if (_evals[name]) {
-    throw new Error(`Evaluator ${name} already exists`);
+  const evalName = makeEvalName(name, evaluator.metadata);
+  if (_evals[evalName]) {
+    throw new Error(`Evaluator ${evalName} already exists`);
   }
   if (globalThis._lazy_load) {
-    _evals[name] = { name, ...evaluator };
+    _evals[evalName] = { evalName, projectName: name, ...evaluator };
```
ankrgyl, line 105: should we throw an error here if the name already exists? (I understand this is an existing bug/wart), not something new in your change

py/src/braintrust/framework.py
```diff
@@ -87,6 +87,21 @@ class EvalScorerArgs(SerializableDataClass):
 ]
 
 
+@dataclasses.dataclass
+class EvalMetadata(SerializableDataClass):
+    """
+    Additional metadata for the eval definition, such as experiment name.
+    """
+
+    experiment_name: Optional[str] = None
```
ankrgyl, line 96: you can write a docstring on this field itself (see how this is done in the summary) too

js/src/framework.ts
```diff
@@ -43,25 +44,46 @@ export type EvalScorer<Input, Output> =
...
+  if (metadata?.experimentName) {
+    out += ` [experimentName=${metadata.experimentName}]`;
+  }
+  return out;
 }
 
 export type EvaluatorDef<Input, Output> = {
-  name: string;
+  evalName: string;
+  projectName: string;
```
ankrgyl, line 82: nit: put `projectName` first since it's the first/primary name the user specifies

py/src/braintrust/framework.py
```diff
@@ -99,9 +114,14 @@ class Evaluator:
     """
 
     """
-    The name of the evaluator. This corresponds to a project name in Braintrust.
+    A unique name for the eval.
```
ankrgyl, line 117: Maybe something like `A name that uniquely defines this type of experiment. You do not need to change it each time the experiment runs, but you should not have other experiments in your code with the same name.`

js/src/framework.ts
```diff
@@ -43,25 +44,46 @@ export type EvalScorer<Input, Output> =
   | ((args: EvalScorerArgs<Input, Output>) => Score)
   | ((args: EvalScorerArgs<Input, Output>) => Promise<Score>);
 
+export type EvalMetadata = {
```
ankrgyl, line 47: should we document these too?

js/src/framework.ts
manugoyal: Isn't this checked right above?

js/src/framework.ts
ankrgyl: i meant with a docstring

js/src/framework.ts
ankrgyl: oops 🤦

ankrgyl (approved):

PR #27, opened 2023-10-30
Fix deadlock in python logger.

manugoyal:
The condition variable approach was flawed because `condition_variable.wait()` will only respond to notify signals sent after the `wait` runs. So if the logger thread was in the middle of a flush while the producer thread fills up the queue and sends a notify signal, the consumer thread will miss the notify signal on its next `wait` call and we have a deadlock.

The solution is to replace the condition variable with a semaphore. So that the producer can increment the semaphore as a "persistent" way to trigger the consumer thread, which blocks on the semaphore being 0.

One possible concern is that the background thread can spin for a while on a filled-up semaphore after draining the queue (since we call `release` each time we log). I ran a quick benchmark for the cost to drain a semaphore:
```
In [1]: def drain_semaphore(sem):
   ...:     while sem.acquire(blocking=False):
   ...:         pass
   ...:

In [2]: import threading

In [3]: sem = threading.Semaphore(value=0)
...
In [9]: %timeit sem.release(100); drain_semaphore(sem)
43.2 µs ± 1.2 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [10]: %timeit sem.release(1000); drain_semaphore(sem)
420 µs ± 5.51 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [11]: %timeit sem.release(10000); drain_semaphore(sem)
4.19 ms ± 115 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Seems to be about 400 ns per acquire call, and 420us for a completely filled-up queue, so not too worried.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -349,9 +349,9 @@ class _LogThread:
         except Exception:
             queue_size = 1000
         self.queue = queue.Queue(maxsize=queue_size)
-        # Each time we put items in the queue, we notify any waiting consumer
-        # threads to attempt a flush.
-        self.queue_filled_condition = threading.Condition()
+        # Each time we put items in the queue, we increment a semaphore to
+        # indicate to any consumer thread that it should attempt a flush.
+        self.queue_filled_semaphore = threading.Semaphore(value=0)
 
         atexit.register(self._finalize)
 
@@ -362,11 +362,9 @@ class _LogThread:
                 self.queue.put_nowait(event)
             except queue.Full:
                 # Notify consumers to start draining the queue.
-                with self.queue_filled_condition:
-                    self.queue_filled_condition.notify()
+                self.queue_filled_semaphore.release()
                 self.queue.put(event)
-        with self.queue_filled_condition:
-            self.queue_filled_condition.notify()
+        self.queue_filled_semaphore.release()
 
     def _start(self):
         if not self.started:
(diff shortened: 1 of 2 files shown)


py/src/braintrust/logger.py
```diff
@@ -362,11 +362,9 @@ def log(self, *args):
                 self.queue.put_nowait(event)
             except queue.Full:
                 # Notify consumers to start draining the queue.
-                with self.queue_filled_condition:
-                    self.queue_filled_condition.notify()
+                self.queue_filled_semaphore.release()
```
ankrgyl, line 365: I'm not sure I follow the logic here. Once you call `release()` a consumer thread _can_ wake up, but it's not guaranteed that (a) you'll context switch to it and (b) that you'll only context switch here _after_ it has pulled something off the queue. Therefore, I think you can still lose an event here.

In traditional concurrency land, you'd have a condition variable here that's "paired" with a mutex, and the mutex is _released_ when you call `wait`. Then, the thread that dequeues would, after clearing up the queue, acquire hte mutex and notify the condition variable, allowing this thread to proceed.

manugoyal: From offline discussion, we shouldn't be at risk of losing data because the subsequent `put` is blocking until the queue has space.

ankrgyl (approved):

PR #28, opened 2023-10-31
Remove ping from the logger

ankrgyl:
This both reduces a roundtrip and also keeps the logger alive if `ping()` fails.

We still need to make the logger return a no-op object if it cannot connect to braintrust at all.

Fixes BRA-527

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -158,7 +158,6 @@ class BraintrustState {
 
   private _apiConn: HTTPConnection | null;
   private _logConn: HTTPConnection | null;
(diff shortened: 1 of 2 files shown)


manugoyal (approved):
exciting! Shall we remove `user_id` from the javascript SDK logging as well?

ankrgyl:
> exciting! Shall we remove `user_id` from the javascript SDK logging as well?

Mind taking a quick look? Just removed it from the JS side too

manugoyal (approved):
lgtm. Assuming unit tests pass.

PR #30, opened 2023-11-01
OpenAI wrappers for JS and PY

ankrgyl:
A really thin wrapper for the OpenAI SDK (v4) that:

* Preserves the type information of the original OpenAI SDK object
* Wraps the chat completion methods

Fixes BRA-559, BRA-560


ankrgyl:
@manugoyal one thing I've been thinking about is whether we should wrap the Python module into a wrapper instead of a class (as I have). E.g. using [__getattr__](https://peps.python.org/pep-0562/) at the module level.

py/src/braintrust/oai.py
```diff
@@ -1,41 +1,128 @@
...
+                                )
+                                first = False
+                            all_results.append(item)
+                            yield item
+                        span.log(output=all_results)
+                    finally:
+                        span.end()
 
+                should_end = False
+                return (x for x in gen())
```
manugoyal, line 40: does `return gen()` not work for some reason?

py/src/braintrust/oai.py
```diff
@@ -1,41 +1,128 @@
...
+import time
 
-import json
-import sqlite3
+from .logger import current_span
 
-from .cache import CACHE_PATH
 
-OAI_CACHE = CACHE_PATH / "oai.sqlite"
+class ChatCompletionWrapper:
```
manugoyal, line 6: I wonder if this wrapper could be used for more than just `ChatCompletion`. What about any object that derives from `EngineAPIResource`?

manugoyal (commented):
Cool. Only looked at python. Yeah I think wrapping the module could be cool. So that people just have to change their imports from `import openai` to `import braintrust_openai` or whatever?

py/src/braintrust/oai.py
```diff
@@ -1,41 +1,128 @@
-# TODO MOVE TO AUTOEVALS IMPORT
-# https://github.com/braintrustdata/braintrust/issues/218
+import time
```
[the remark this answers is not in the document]
ankrgyl: done

py/src/braintrust/oai.py
ankrgyl: Each one has distinct things we'd want to log I think

py/src/braintrust/oai.py
ankrgyl: It does

py/src/braintrust/oai.py
```diff
@@ -1,41 +1,128 @@
...
+            "input": messages,
+            "metadata": params,
+        }
 
-    return resp
+    def __getattr__(self, name):
+        return getattr(self.chat, name)
+
+
+class OpenAIWrapper:
```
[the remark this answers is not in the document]
ankrgyl: We should detect the various versions, and wrap them separately. I'm about to do this to handle the new version

py/src/braintrust/logger.py
```diff
@@ -333,7 +333,7 @@ def construct_json_array(items):
 
 
 class _LogThread:
-    def __init__(self, name=None):
+    def __init__(self, name=None, lazy_login=None):
```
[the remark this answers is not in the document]
manugoyal: Can we add a unit test?

py/src/braintrust/logger.py
```diff
@@ -791,13 +794,17 @@ def decorator(span_args, span_kwargs, f):
 
         @wraps(f)
         def wrapper_sync(*f_args, **f_kwargs):
-            with start_span(*span_args, **span_kwargs):
-                return f(*f_args, **f_kwargs)
+            with start_span(*span_args, **span_kwargs) as span:
```
manugoyal, line 797: Can we put these changes in a separate PR? I am worried about having substantially different behavior between the python and javascript SDKs, and I don't want to pollute this PR with that conversation.

py/src/braintrust/oai.py
manugoyal: Is that true? I think both [`Completion`](https://platform.openai.com/docs/api-reference/completions/create) and [`Edit`](https://platform.openai.com/docs/api-reference/edits) have the same schema, as far as we are drilling down.

py/src/braintrust/oai.py
```diff
@@ -1,41 +1,186 @@
...
+
+
+class ChatV1Wrapper(NamedWrapper):
+    def __init__(self, chat):
+        super().__init__(chat)
 
-def run_cached_request(Completion=None, **kwargs):
-    if Completion is None:
-        # OpenAI is very slow to import, so we only do it if we need it
         import openai
```
manugoyal, line 162: we don't need this import anymore?

py/src/braintrust/oai.py
```diff
@@ -1,41 +1,186 @@
...
-    resp = Completion.create(**kwargs).to_dict()
 
-    cursor.execute("""INSERT INTO "cache" VALUES (?, ?)""", [param_key, json.dumps(resp)])
-    conn.commit()
+def wrap_openai(openai):
+    """
+    Wrap the openai module to add tracing. If Braintrust is not configured, this is a no-op.
 
-    return resp
+    :param openai: The openai module
```
manugoyal, line 181: For v1 they should be passing in the client object, not the module, right?

js/src/logger.ts
```diff
@@ -1231,7 +1231,7 @@ function _urljoin(...parts: string[]): string {
   return parts.map((x) => x.replace(/^\//, "")).join("/");
 }
 
-function getCurrentUnixTimestamp(): number {
+export function getCurrentUnixTimestamp(): number {
```
manugoyal, line 1234: Could this go in `util`?

js/src/oai.ts
```diff
@@ -1,24 +1,154 @@
...
+        prompt_tokens: number;
+        completion_tokens: number;
+      }
+    | undefined;
+}
+
+// TODO: Mock this up better
+type StreamingChatResponse = any;
+
+function wrapChatCompletionNonStreaming<
```
manugoyal, line 77: Should we just call this `wrapChatCompletion` if it also wraps streaming?

js/src/oai.ts
```diff
@@ -1,24 +1,154 @@
...
+  P extends ChatParams,
+  C extends NonStreamingChatResponse | StreamingChatResponse
+>(completion: (params: P) => Promise<C>): (params: P) => Promise<any> {
+  return async (params: P) => {
+    if (params.stream) {
+      const { messages, ...rest } = params;
+      const span = currentSpan().startSpan("OpenAI Chat Completion", {
```
manugoyal, line 84: should we use the global method `startSpan`? `currentSpan` will not work if your active object is an experiment/logger. Also could we share the event content across both branches?
```
const { messages, ...rest } = params;
const initEvent = {
  input: messages,
  metadata: { ...rest }
};
...
```

js/src/oai.ts
```diff
@@ -1,24 +1,154 @@
...
+        first = false;
+      }
+
+      allResults.push(item);
+      yield item;
+    }
+    this.span.log({
+      output: allResults,
+    });
+    this.span.end();
```
manugoyal, line 152: should we do try/finally here (as in python) just for safety?

manugoyal (commented):
overall lgtm. Are there examples we can add to test this out and make sure things still work over time?

js/src/logger.ts
ankrgyl: yep

js/src/oai.ts
ankrgyl: yep

js/src/oai.ts
ankrgyl: Yeah good call

js/src/oai.ts
ankrgyl: Yep

py/src/braintrust/logger.py
ankrgyl: Yep

ankrgyl: https://github.com/braintrustdata/braintrust-sdk/pull/33

py/src/braintrust/oai.py
ankrgyl: I haven't looked at the return types -- in general i think we should proceed with caution until we thoroughly test.

py/src/braintrust/oai.py
ankrgyl: we do on the RHS because we do a typecheck

py/src/braintrust/logger.py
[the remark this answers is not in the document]
ankrgyl: added

manugoyal (approved):

PR #31, opened 2023-11-01
Fix currentSpan() in no-op case

ankrgyl:
If there's no span set, then return the no-op span.


ankrgyl:
#32 is replacing this

PR #36, opened 2023-11-09
SDK fixes for github actions.

manugoyal:
`get_past_n_ancestors` would crash in python if `git merge-base` was failing. Also made the TS and PY implementations more aligned.

### DIFF
--- a/py/src/braintrust/gitutil.py
+++ b/py/src/braintrust/gitutil.py
@@ -82,7 +82,7 @@ def _get_base_branch_ancestor(remote=None):
     try:
         return subprocess.check_output(["git", "merge-base", head, f"{remote_name}/{base_branch}"]).decode().strip()
     except subprocess.CalledProcessError as e:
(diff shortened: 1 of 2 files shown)


py/src/braintrust/gitutil.py
```diff
@@ -82,7 +82,7 @@ def _get_base_branch_ancestor(remote=None):
     try:
         return subprocess.check_output(["git", "merge-base", head, f"{remote_name}/{base_branch}"]).decode().strip()
     except subprocess.CalledProcessError as e:
-        _logger.warning(f"Could not find a common ancestor with {remote_name}/{base_branch}", e)
+        # _logger.warning(f"Could not find a common ancestor with {remote_name}/{base_branch}")
```
manugoyal, line 85: This was commented out in TS so I figured we could do it here too.

ankrgyl (approved):

PR #43, opened 2023-11-16
Fix OpenAI wrapping in edge environments

ankrgyl:
There were two issues here:

* We weren't exporting `oai.ts` from the `browser.ts` entrypoint, which next.js's edge runtime uses
* Next.js seems to stub / support `AsyncLocalStorage` in edge environments, even though it's not supported in the browser. I copied a small snippet of code that adds it to global scope and attempt to initialize it, which fixes it in edge environments.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -1147,6 +1147,13 @@ export function startSpan(args?: StartSpanOptionalNameArgs): Span {
   const name =
     (nameOpt ?? iso.getCallerLocation()?.caller_functionname) || "root";
   const parentSpan = currentSpan();
+
+  if (!parentSpan) {
+    throw new Error(
+      "Cannot call startSpan() from outside a trace. Please wrap this code in a traced() callback."
+    );
+  }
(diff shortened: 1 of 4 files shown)


manugoyal (approved):

js/src/logger.ts
```diff
@@ -1147,6 +1147,13 @@ export function startSpan(args?: StartSpanOptionalNameArgs): Span {
   const name =
     (nameOpt ?? iso.getCallerLocation()?.caller_functionname) || "root";
   const parentSpan = currentSpan();
+
+  if (!parentSpan) {
+    throw new Error(
+      "Cannot call startSpan() from outside a trace. Please wrap this code in a traced() callback."
```
manugoyal, line 1153: I think it's always a bug if `parentSpan` is undefined. The ISO should be initializing it to a noop span.

ankrgyl: discussed offline -- we need to dig into this further

PR #48, opened 2023-11-17
Fix race condition when logging from multiple threads.

manugoyal:
Fixes BRA-618

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -371,10 +371,12 @@ class _LogThread:
         self.queue_filled_semaphore.release()
 
     def _start(self):
-        with self.start_thread_lock:
-            if not self.started:
-                self.thread.start()
-                self.started = True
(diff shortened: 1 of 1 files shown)


py/src/braintrust/logger.py
```diff
@@ -370,9 +371,10 @@ def log(self, *args):
         self.queue_filled_semaphore.release()
 
     def _start(self):
-        if not self.started:
-            self.thread.start()
-            self.started = True
+        with self.start_thread_lock:
+            if not self.started:
+                self.thread.start()
+                self.started = True
```
ankrgyl, line 377: This is going to add contention in the common case too. You can tweak this slightly to:

```
    if not self.started:
        with self.start_thread_lock:
            if not self.started:
                self.thread.start()
                self.started = True
```

ankrgyl (approved):

PR #55, opened 2023-12-04
Properly validate that logged metrics are shallow dicts

aphinx:
* Validate that metrics are shallow dicts (one level deep, values are numbers).

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -1276,9 +1276,13 @@ function validateAndSanitizeExperimentLogPartialArgs(
   }
 
   if (event.metrics) {
-    for (const key of Object.keys(event.metrics)) {
+    for (const [key, value] of Object.entries(event.metrics)) {
       if (typeof key !== "string") {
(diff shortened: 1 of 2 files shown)


js/src/logger.ts
```diff
@@ -1276,10 +1276,14 @@ function validateAndSanitizeExperimentLogPartialArgs(
   }
 
   if (event.metrics) {
-    for (const key of Object.keys(event.metrics)) {
+    for (let [key, value] of Object.entries(event.metrics)) {
```
ankrgyl, line 1279: why use `let` here instead of `const`?

manugoyal: Note that we only use `let` in the scores loop above because we're re-assigning `score`. But we don't need to do that here, so `const` would be better.

aphinx: updated

manugoyal (approved):
lgtm!

aphinx:
Fixes BRA-537

PR #56, opened 2023-12-05
Implement `COMMON_BASE_BRANCHES` in JS to speed up common branch lookup

aphinx:
* Try to match common branch names (`main`, `master`, `develop`) before resorting to a full `git remote show` to find the base branch.

This change mostly copies the existing logic in `gitutil.py`.

Fixes BRA-677.

### DIFF
--- a/js/src/gitutil.ts
+++ b/js/src/gitutil.ts
@@ -1,5 +1,7 @@
 import { simpleGit } from "simple-git";
 
+const COMMON_BASE_BRANCHES = ["main", "master", "develop"];
+
 /**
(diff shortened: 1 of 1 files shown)


manugoyal (approved):
lgtm! Were you able to manually verify that this works as intended? For instance, you could try adding some print statements to trace the codepath and then run an example script like
```
npx tsx tests/expect_tests/experiment_span.ts
```

aphinx:
yup, I ran that test and checked that the new codepath was used with some `console.log`s :)

PR #63, opened 2023-12-15
Add support for trial count parameter

ankrgyl:
Now that we support visualizing trials in the product, this change makes them really easy to run in the eval framework.

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -158,6 +158,13 @@ class Evaluator:
     """
     metadata: Optional[EvalMetadata]
 
+    """
+    The number of times to run the evaluator per input. This is useful for evaluating applications that
+    have non-deterministic behavior and gives you both a stronger aggregate measure and a sense of the
+    variance in the results.
+    """
+    trial_count: int = 1
+
 
 _evals = {}
 _lazy_load = False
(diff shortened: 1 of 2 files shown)


py/src/braintrust/framework.py
```diff
@@ -158,6 +158,13 @@ class Evaluator:
     """
     metadata: Optional[EvalMetadata]
 
+    """
+    The number of times to run the evaluator per input. This is useful for evaluating applications that
+    have non-deterministic behavior and gives you both a stronger aggregate measure and a sense of the
+    variance in the results.
+    """
+    trial_count: int = 1
```
manugoyal, line 166: nit: for uniformity with TS maybe we could keep this None and default it to 1 in the implementation? It's sad that python doesn't provide a `coalesce` function, but I have defined one here https://github.com/braintrustdata/braintrust/blob/e7f1ea81c43443c2c63e364d6dd8542e3de57fac/api/chalicelib/util.py#L38.

manugoyal (approved):

py/src/braintrust/framework.py
ankrgyl: I think I prefer this because it'll be nicer if someone jumps into the definition and is poking around

PR #65, opened 2023-12-17
Create separate proxy-based wrapper for openai.

manugoyal:
This is the new default for when people call `wrap_openai`, though they can opt-into using the non-proxy wrappers by passing `proxy_base_url=None` (or equivalent in javascript).

Other changes to the SDK to facilitate this:

- Add logic for serializing spans, so we can send the root span over to the proxy.

- Add logic for serializing login info, so we can send that over to the proxy and instantiate a BackgroundLogger for the span using the serialized state. This requires parameterizing the BackgroundLogger on a given `logConn`, rather than using the global state connection.

- Change the `get_caller_location` logic to ignore files in the stacktrace that in a subdirectory of the first stacktrace entry, not just the same directory. This seems to slightly change a couple of the internal expect test stack traces, but the changes are not really important (getting rid of meaningless, internal nodeJS callsites).


js/src/oai_wrappers/non_proxy_wrappers.ts
```diff
@@ -0,0 +1,139 @@
+import { Span, startSpan } from "../logger";
```
manugoyal, line 1: This is purely a cut-and-paste from `oai.ts`, except for changing the wrapper function name.

py/src/braintrust/oai_wrappers/non_proxy_wrappers.py
```diff
@@ -0,0 +1,180 @@
+"""Wrappers for the openAI client which do not go through the Braintrust
```
manugoyal, line 1: This is purely a cut-and-paste from `oai.py`, except for changing the names of the wrapper classes.

js/src/logger.ts
```diff
@@ -216,46 +249,6 @@ export function _internalSetInitialState() {
...
-      }
-      warningMessage += msg;
-    });
-    console.warn(warningMessage);
-  }
-}
-
-let unterminatedObjects = new UnterminatedObjectsHandler();
-
 class FailedHTTPResponse extends Error {
```
ankrgyl, line 252: is this carried over from the other change?

js/src/logger.ts
```diff
@@ -202,6 +218,23 @@ class BraintrustState {
...
+  // Note: the format should be the same as in logger.py.
+  public serializeLoginInfo(): string {
+    return JSON.stringify({
+      api_url: this.apiUrl,
+      login_token: this.loginToken,
+      org_id: this.orgId,
+      org_name: this.orgName,
+      log_url: this.logUrl,
+    });
+  }
```
ankrgyl, line 231: I think this is redundant, and if you use your `BRAINTRUST_API_KEY` we already re-derive this info in the logger. Perhaps we should just send `x-bt-org-name`, and let the logic in the proxy handle the rest?

ankrgyl (commented):
Discussed offline: can we simplify the state we're sending to the proxy?

js/src/logger.ts
manugoyal: It was done independently, because I didn't want a warning from resuming the serialized root span and not ending it in the proxy. But now that we've decided to only create subspans out of a serialized span, I can leave the removal to the other PR.

js/src/logger.ts
```diff
@@ -979,6 +984,71 @@ export function initLogger(options: Readonly<InitLoggerOptions> = {}) {
...
+}
+
+/**
+ * Wrapper over `braintrust.resumeSpan` which passes the resumed `Span` to the
+ * given callback and ends it afterwards. See `braintrust.resumeSpan` for full
+ * details.
+ *
+ * @param args.setCurrent If true (the default), the span will be marked as the currently-active span for the duration of the callback. Equivalent to calling `braintrust.withCurrent(span, callback)`.
+ */
+export function withResumedSpan<R>(
```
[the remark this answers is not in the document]
manugoyal: Will just remove this for now.

js/src/logger.ts
```diff
@@ -1628,6 +1723,100 @@ export class Experiment {
...
+  | { parent_span_kind: "root_span"; span_id: string; root_span_id?: undefined }
+  | { parent_span_kind: "none"; span_id?: undefined; root_span_id?: undefined };
+
+type SerializedSpanInfo = SpanObjectIds & SpanParentSpanIds;
+
+function SerializedSpanInfoToString(info: SerializedSpanInfo): string {
+  const objectKindIds = (() => {
+    if (info.object_kind === "e") {
+      return [info.project_id, info.experiment_id];
+    } else if (info.object_kind === "pl") {
```
ankrgyl, line 1743: this seems like good logic to write some simple unit tests for

js/src/oai.ts
```diff
@@ -1,157 +1,28 @@
...
 /**
- * Wrap an `OpenAI` object (created with `new OpenAI(...)`) to add tracing. If Braintrust is
- * not configured, this is a no-op
+ * Wrap an `OpenAI` object (created with `new OpenAI(...)`) to add tracing.
  *
  * Currently, this only supports the `v4` API.
  *
- * @param openai
+ * @param openai The `OpenAI` object.
+ * @param proxyBaseUrl Use the Braintrust proxy (https://github.com/braintrustdata/braintrust-proxy) as the base URL. When left `undefined`, the URL is obtained from the environment variable `BRAINTRUST_PROXY_URL`, defaulting to `https://braintrustproxy.com/v[api_version]`. Pass `null` to leave the original base URL intact and use the non-proxy wrapper, or pass a string to use a custom URL.
```
ankrgyl, line 10: This seems like a fairly subtle distinction (especially between `undefined` and `null`). I think it'd be more intuitive to:

* rename `args` to `options`
* make the first argument `useProxy` and make it a `string | bool`. If `true`, it uses the default proxy url. If a string, it uses that URL. If we want to keep the proxy off by default, then maybe we create a flag like `disableProxy`.
* make the second argument `apiKey`.

ankrgyl, line 10: I think it's fine to just mention `v1` instead of the api version

js/src/oai_wrappers/proxy_wrappers.ts
```diff
@@ -0,0 +1,69 @@
...
+function getProxyBaseUrlDefault({ version }: { version: number }): string {
+  return (
+    process.env["BRAINTRUST_PROXY_URL"] ??
+    `https://braintrustproxy.com/v${version}`
+  );
+}
+
+function getProxyApiKeyDefault(): string | undefined {
+  return process.env["BRAINTRUST_API_KEY"];
+}
```
ankrgyl, line 12: We probably need to check whether process is defined or not. In browser/edge environments, it's not

js/src/logger.ts
```diff
@@ -1628,6 +1723,100 @@ export class Experiment {
...
+  | { object_kind: "pl"; org_id: string; project_id: string; log_id: "g" };
+
+type SpanParentSpanIds =
+  | { parent_span_kind: "sub_span"; span_id: string; root_span_id: string }
+  | { parent_span_kind: "root_span"; span_id: string; root_span_id?: undefined }
+  | { parent_span_kind: "none"; span_id?: undefined; root_span_id?: undefined };
+
+type SerializedSpanInfo = SpanObjectIds & SpanParentSpanIds;
+
+function SerializedSpanInfoToString(info: SerializedSpanInfo): string {
```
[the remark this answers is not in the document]
manugoyal: un-capitalized it

js/src/logger.ts
manugoyal: done!

js/src/oai.ts
```diff
@@ -7,19 +7,24 @@ import { openAIV4ProxyWrapper } from "./oai_wrappers/proxy_wrappers";
...
  *
  * @param openai The `OpenAI` object.
- * @param proxyBaseUrl Use the Braintrust proxy (https://github.com/braintrustdata/braintrust-proxy) as the base URL. When left `undefined`, the URL is obtained from the environment variable `BRAINTRUST_PROXY_URL`, defaulting to `https://braintrustproxy.com/v[api_version]`. Pass `null` to leave the original base URL intact and use the non-proxy wrapper, or pass a string to use a custom URL.
- * @param proxyApiKey For clients using the proxy wrapper (`proxyBaseUrl` is not `null`). When left `undefined`, the API key is set from `BRAINTRUST_API_KEY` if available. Pass `null` to leave the original API key intact, or pass a string to use a custom API key.
+ * @param options.useProxy By default (or if `false`), the wrapper does not trace through the proxy. Pass `true` to use the Braintrust proxy as the base URL. The URL is obtained from the environment variable `BRAINTRUST_PROXY_URL`, defaulting to `https://braintrustproxy.com/v1. Pass a string to use a custom proxy URL.
+ * @param options.apiKey Only used when `useProxy` is set. By default, the API key is set from `BRAINTRUST_API_KEY` if available. Pass a string to use a custom API key.
```
ankrgyl, line 11: I don't quite follow the `apiKey` setting. Why would someone need to set this, instead of setting the api key in the OpenAI client? I'm imagining the code like

```
wrapOpenAI(new OpenAI({ apiKey: process.env.BRAINTRUST_API_KEY }), { useProxy: true })
```

py/src/braintrust/oai.py
```diff
@@ -1,20 +1,22 @@
+from typing import Optional, Union
+
 from braintrust.oai_wrappers.non_proxy_wrappers import OpenAIV0NonProxyWrapper, OpenAIV1NonProxyWrapper
-from braintrust.oai_wrappers.proxy_wrappers import PROXY_SENTINEL, openai_v1_proxy_wrapper
+from braintrust.oai_wrappers.proxy_wrappers import openai_v1_proxy_wrapper
 
 
-def wrap_openai(openai, proxy_base_url=PROXY_SENTINEL, proxy_api_key=PROXY_SENTINEL):
+def wrap_openai(openai, use_proxy: Union[None, str, bool] = None, api_key: Optional[str] = None):
```
ankrgyl, line 7: super nit: slightly easier to read as `Union[None, bool, str]`

ankrgyl (approved):

js/src/oai.ts
manugoyal: I guess there are two ways someone can initialize the OpenAI client
```
// Uses OPENAI_API_KEY
new OpenAI();

// Uses custom api key
new OpenAI({apiKey: [custom_api_key]});
```

I was thinking that in the first case (which I imagine is the most common), they don't have to change anything about their code (just add the wrapper) and it'll change the API key to `BRAINTRUST_API_KEY`. In the second case, they can move the `apiKey` argument to the wrapper to tell it to set a custom API key instead of `BRAINTRUST_API_KEY`. But that wouldn't work if they hadn't defined `OPENAI_API_KEY` at all.

We could change the API key functionality so that if `openai.apiKey === process.env.OPENAI_API_KEY`, only then do we change it to `BRAINTRUST_API_KEY`, otherwise we leave it intact? And we could add a flag `keepApiKey?: boolean`, which if true, will not alter the API key at all?

PR #33, opened 2023-11-07
Capture input/output args automatically in traced decorator

ankrgyl:

PR #67, opened 2023-12-18
Simplify SDK tracing API

manugoyal:
We make the following improvements/simplifications to the SDK:

- Globally, we maintain a single `currentExperiment` and `currentLogger`, and a nested `AsyncLocalStorage` `currentSpan`. Now there is no need for wrapper functions / context managers in experiments or loggers. But I kept the functions around with a deprecation warning.

- The `traced` decorator in python will auto-log the function's input/output.

- In javascript, the global `traced` method will now consider `loggers`, at the cost of making the whole function async.

- Add a `flush` function in place of `close` for experiments and loggers. Deprecate the `close` method.


js/src/framework.ts
```diff
@@ -220,27 +218,25 @@ export async function runEvaluator(
...
-              if (outputResult instanceof Promise) {
-                output = await outputResult;
-              } else {
-                output = outputResult;
-              }
-              currentSpan().log({ input: datum.input, output });
-            },
-            { name: "task" }
-          );
+          await currentSpan().traced("task", async () => {
```
ankrgyl, line 221: Per discussion RE deno issues, I think this will fail if async local storage doesn't work. Can we thread the relevant `span` here explicitly?

js/src/logger.ts
```diff
@@ -816,51 +759,8 @@ export async function init(
...
-    () => experiment.close()
-  );
-}
-
-/**
- * Wrapper over `braintrust.initLogger`, which passes the initialized `Logger` it to the given callback and closes it afterwards. See `braintrust.initLogger` for full details.
- *
- * @param options.setCurrent If true (default), set the currently-active logger to the newly-created one. Equivalent to calling `braintrust.withCurrent(logger, callback)`.
- */
-export async function withLogger<R>(
```
ankrgyl, line 849: maybe we can deprecate this, and just assign to the new globla variable if someone does it?

ankrgyl (commented):
Is there a way we can stage this change so that it doesn't break people's code? Like maybe deprecate the functions?

js/src/logger.ts
```diff
@@ -59,33 +60,36 @@ export interface Span {
   log(event: ExperimentLogPartialArgs): void;
 
   /**
-   * Create a new span. This is useful if you want to log more detailed trace information beyond the scope of a single log event. Data logged over several calls to `Span.log` will be merged into one logical row.
+   * Create a new span and run the provided callback. This is useful if you want to log more detailed trace information beyond the scope of a single log event. Data logged over several calls to `Span.log` will be merged into one logical row.
```
manugoyal, line 63: I shifted the bulk of the documentation to `traced` rather than `startSpan`, because most of the times people should be calling `traced` in javascript. I also made one breaking change to make the `name` an optional parameter instead of required. This is only breaking in JS, because in python it retains the same position in the arglist, but in JS we move it into the object.

I felt like optional name everywhere was the simplest DX and hopefully we can ask folks to make the slight modification to anywhere they are tracing with spans in JS.

js/src/logger.ts
```diff
@@ -1108,123 +1024,98 @@ export async function summarize(
...
+export async function startSpan(args?: StartSpanArgs): Promise<Span> {
+  const parentObject = getSpanParentObject();
+  const { name: nameOpt, ...argsRest } = args ?? {};
+  const name = parentObject.kind === "span" ? nameOpt : nameOpt ?? "root";
+  return parentObject.startSpan({ name, ...argsRest });
+}
+
+// Set the given span as current within the given callback and any asynchronous
+// operations created within the callback.
+function withCurrent<R>(span: Span, callback: (span: Span) => R): R {
```
manugoyal, line 1117: Un-exposing this function is technically a breaking change. If we want to more-softly deprecate it with a warning, we could also do that.

py/src/braintrust/framework.py
```diff
@@ -280,9 +280,12 @@ def Eval(
             loop = None
 
         async def run_to_completion():
-            with init_experiment(evaluator.project_name, evaluator.experiment_name, evaluator.metadata) as experiment:
+            experiment = init_experiment(evaluator.project_name, evaluator.experiment_name, evaluator.metadata)
```
ankrgyl, line 283: are we removing the context manager altogether? What happens if someone uses it?

py/src/braintrust/logger.py
```diff
@@ -710,121 +669,103 @@ def summarize(summarize_scores=True, comparison_experiment_id=None):
...
-
-    def __enter__(self):
-        self.context_token = self.object_cvar.set(self.object)
-
-    def __exit__(self, type, value, callback):
-        del type, value, callback
-
-        self.object_cvar.reset(self.context_token)
+def _try_log_input_output(span, f_args, f_kwargs, output):
+    log_kwargs = dict(input=dict(args=f_args, kwargs=f_kwargs), output=output)
```
ankrgyl, line 724: is there a way we can format it so that it's just always a dict with the arg names? I.e. a function like

```
def foo(a, b, **kwargs):
  return a + b + kwargs["foo"]
```

`foo(1, b=2, foo=3)` would log:

```
{
  "input": {"a": 1, "b": 2, "foo": 3},
  "output": 6
}
```

py/src/braintrust/logger.py
```diff
@@ -1536,42 +1447,33 @@ def fetched_data(self):
...
 
-        :returns: The dataset id.
-        """
-        self._check_not_finished()
-
-        self.logger.flush()
-
-        self.finished = True
-        _unterminated_objects.remove_unterminated(self)
+        print(
```
ankrgyl, line 1462: should we make these print statements to stderr?

js/src/logger.ts
ankrgyl: So now does `experiment.traced` have the same structure/arguments as `span.traced`?

js/src/logger.ts
ankrgyl: Let's just leave it as is

js/src/logger.ts
```diff
@@ -1539,6 +1407,21 @@ export class Experiment {
     );
   }
 
+  /**
+   * Lower-level alternative to `traced`, which does not automatically end the span or mark it as current.
```
ankrgyl, line 1411: i like framing this as a lower-level alternative

js/src/logger.ts
```diff
@@ -1108,123 +1024,98 @@ export async function summarize(
...
+ *  * Currently-active span
+ *  * Currently-active experiment
+ *  * Currently-active logger
+ *
+ * and creates a span under the first one that is active. If none of these are active, it returns a no-op span object.
+ *
+ * See `Span.traced` for full details.
  */
-export function traced<R>(
+export async function traced<R>(
```
ankrgyl, line 1089: I know we've gone back & forth on this, but I'm mildly worried about making this function `async`, since I think it graph colors any uses of `traced` to also have to be `async` functions.

I know it's a lot to ask, but I _think_ we could solve this by making the `rootProject: RegisteredProject` arg in `SpanImpl`'s constructor a `Promise<RegisteredProject>`. Then, we can make the subfields of `objectInfo` also promises. And in the flush thread, we can check the values of the dictionary, and if are any are promises await them.

WDYT?

ankrgyl (commented):
* Should we close https://github.com/braintrustdata/braintrust-sdk/pull/33?

py/src/braintrust/framework.py
manugoyal: It's not removed. Just a no-op now.

js/src/logger.ts
manugoyal: Yes now all of the `traced` methods have the same signature.

py/src/braintrust/framework.py
ankrgyl: Ah cool

js/src/logger.ts
manugoyal: Yeah I think pushing all the async work into the background logger could be a good design. As discussed offline, I separated out the python changes into https://github.com/braintrustdata/braintrust-sdk/pull/74, and we'll fix the JS API separately.

manugoyal:
Whoops I merged the wrong PR! Reverted in https://github.com/braintrustdata/braintrust-sdk/commit/aa64beaac06e1cd9b7351a90bf6938925f183dce

PR #69, opened 2023-12-19
Check if AsyncLocalStorage has enterWith function, to make Deno happy

cngarrison:
Deno doesn't support the `enterWith` function in `AsyncLocalStorage`:

- https://github.com/denoland/deno/issues/18127

The function is still marked experimental.

- https://nodejs.org/api/async_context.html#asynclocalstorageenterwithstore

This PR checks the `enterWith` function exists before calling it.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -169,7 +169,9 @@ class BraintrustState {
     this.currentExperiment = iso.newAsyncLocalStorage();
     this.currentLogger = iso.newAsyncLocalStorage();
     this.currentSpan = iso.newAsyncLocalStorage();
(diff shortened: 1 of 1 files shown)


manugoyal (commented):
thanks @cngarrison! Can we also return `noopSpan` if the local storage value is undefined here? https://github.com/braintrustdata/braintrust-sdk/blob/f2eacfe8657dfc1c08d037ecd4f4927969566ffa/js/src/logger.ts#L1137

Basically
```
return _state.currentSpan.getStore() ?? noopSpan;
```

manugoyal (approved):
thanks, lgtm!

PR #71, opened 2023-12-20
Add support in wrapOpenAI for the beta.chat.completions.stream API

vitorbal:
This seems to work, but probably needs more testing.

This change adds support for `wrapOpenAI` to also trace `openai.beta.chat.completions.stream` calls:
https://github.com/openai/openai-node?tab=readme-ov-file#streaming-responses-1


manugoyal (approved):
looks great, thanks @vitorbal!

I added a test to our internal repo which exercises this new API. Looks like the tracing works!

PR #75, opened 2023-12-22
Add option to not set new experiment/logger as current.

manugoyal:
Motivation was for the Eval framework to not leave the global experiment lying around afterwards for people to accidentally write to.

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -368,7 +368,7 @@ class DictEvalHooks(EvalHooks):
 
 
 def init_experiment(project_name, experiment_name: Optional[str] = None, metadata: Optional[Metadata] = None):
-    ret = _init_experiment(project_name, experiment=experiment_name, metadata=metadata)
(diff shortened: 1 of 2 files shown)


py/src/braintrust/framework.py
```diff
@@ -368,7 +368,7 @@ def meta(self, **info):
 
 
 def init_experiment(project_name, experiment_name: Optional[str] = None, metadata: Optional[Metadata] = None):
-    ret = _init_experiment(project_name, experiment=experiment_name, metadata=metadata)
+    ret = _init_experiment(project_name, experiment=experiment_name, metadata=metadata, set_current=False)
```
ankrgyl, line 371: do we need to do similar stuff in JS?

manugoyal: As of yet the JS SDK still uses async local storage for experiments/loggers, so this isn't a problem.

ankrgyl (approved):

PR #76, opened 2023-12-22
Small improvements to eval framework (jsonl, public experiments)

ankrgyl:
* Add support for public experiments through `Eval(...)`
* Add support for `--jsonl` flag to make parsing outputs easy (eg in CI)

### DIFF
--- a/js/src/framework.ts
+++ b/js/src/framework.ts
@@ -45,25 +45,43 @@ export type EvalScorer<Input, Output, Expected> = (
   args: EvalScorerArgs<Input, Output, Expected>
 ) => Score | Promise<Score>;
 
-/**
- * An evaluator is a collection of functions that can be used to evaluate a model.
- * It consists of:
- * - `data`, a function that returns a list of inputs, expected outputs, and metadata
- * - `task`, a function that takes an input and returns an output
- * - `scores`, a set of functions that take an input, output, and expected value and return a score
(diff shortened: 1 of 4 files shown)


js/src/framework.ts
```diff
@@ -118,12 +136,13 @@ export async function Eval<Input, Output, Expected>(
           progressReporter,
           []
         );
-        reportEvaluatorResult(name, ret, true);
+        reportEvaluatorResult(name, ret, true, false);
```
manugoyal, line 139: Can we change these to named arguments? Or if backwards-compatibility is a must, add any new arguments as named arguments?

py/src/braintrust/framework.py
```diff
@@ -177,32 +184,42 @@ def pluralize(n, singular, plural):
         return plural
 
 
-def report_evaluator_result(eval_name, results, summary, verbose):
+def report_evaluator_result(eval_name, results, summary, verbose, jsonl):
```
manugoyal, line 187: same here, would be nice if we passed by keyword name or made these keyword-only args.

manugoyal (approved):

js/src/framework.ts
ankrgyl: fixed

py/src/braintrust/framework.py
ankrgyl: fixed

PR #77, opened 2023-12-22
Support reading experiments

ankrgyl:
Similar to datasets, support reading experiments through the SDK


core/py/src/braintrust_core/util.py
```diff
@@ -12,6 +12,13 @@ def as_json(self, **kwargs):
         """Serialize the object to JSON."""
         return json.dumps(self.as_dict(), **kwargs)
 
+    @classmethod
+    def from_dict(cls, d: Dict):
+        """Deserialize the object from a dictionary."""
+        fields = set(f.name for f in dataclasses.fields(cls))
+        filtered = {k: v for k, v in d.items() if k in fields}
```
manugoyal, line 19: I'm guessing this doesn't work for nested dataclasses? I think it won't even crash at runtime because it's just the types that are different. Perhaps we could:
- assert `type(v) == field.type`
- optionally, recursively invoke `from_dict` if `isinstance(field.type, SerializableDataClass)`

js/src/framework.ts
```diff
@@ -10,16 +10,52 @@ import {
...
 
 export interface EvalCase<Input, Expected> {
   input: Input;
   expected?: Expected;
   metadata?: Metadata;
 }
 
+export type BaseExperiment<Input, Expected> = {
+  _type: "BaseExperiment";
+  _phantom?: [Input, Expected];
```
manugoyal, line 23: is this used?

js/src/framework.ts
```diff
@@ -108,14 +144,23 @@ globalThis._evals = {};
...
-): Promise<void | ExperimentSummary> {
+): Promise<ExperimentSummary> {
   const evalName = makeEvalName(name, evaluator.experimentName);
   if (_evals[evalName]) {
     throw new Error(`Evaluator ${evalName} already exists`);
   }
   if (globalThis._lazy_load) {
     _evals[evalName] = { evalName, projectName: name, ...evaluator };
-    return;
+    // Better to return this empty object than have an annoying-to-use signature
```
manugoyal, line 154: Isn't the standard solution here to return `undefined` or `null`?

js/src/framework.ts
```diff
@@ -210,10 +255,43 @@ export async function runEvaluator(
...
+      }
+      name = baseExperiment.name;
+    }
+    dataResult = init(evaluator.projectName, {
+      experiment: name,
+      open: true,
+    }).asDataset();
+  }
+
+  let data: EvalCase<unknown, unknown>[] = [];
```
manugoyal, line 285: unrelated nit on the line below: we should just be able to write `let dataResult = await (typeof ...)` on line 258 and remove the `if` here.

js/src/framework.ts
```diff
@@ -210,10 +255,43 @@ export async function runEvaluator(
...
     throw new Error("Unimplemented: string data paths");
   }
-  const dataResult = evaluator.data();
-  let data = null;
+  let dataResult =
+    typeof evaluator.data === "function" ? evaluator.data() : evaluator.data;
+
+  if ("_type" in dataResult) {
+    if (dataResult._type !== "BaseExperiment") {
+      // For some reason, the typesystem won't let me check if dataResult._type === "BaseExperiment"
```
manugoyal, line 263: Not sure I follow. The TS compiler already thinks it's a `BaseExperiment` at this point, so this is just an extra-cautious check that they didn't put something undesirable in `_type` somehow. So why would we rather check `dataResult._type === "BaseExperiment"`?

js/src/logger.ts
```diff
@@ -24,6 +24,7 @@ import {
   isEmpty,
   LazyValue,
 } from "./util";
+import { EvalCase } from "./framework";
```
manugoyal, line 27: yikes, now there's a circular dependency between `logger` and `framework`. I'm surprised javascript doesn't complain about this. Should we factor this out into `util`?

js/src/logger.ts
```diff
@@ -943,7 +945,7 @@ export type InitOptions = {
  * @param options.description An optional description of the experiment.
  * @param options.dataset (Optional) A dataset to associate with the experiment. You can pass in the name of the dataset (in the same project) or a
  * dataset object (from any project).
- * @param options.update If the experiment already exists, continue logging to it.
+ * @param options.open If the experiment already exists, open it instead of creating a new one.
```
manugoyal, line 948: Hmm I'm not sure I love the idea of conflating `open` and `update`. I was hoping that this would serve as the implementation of the `GET /v1/experiment/{experiment_id}` REST endpoint, which just lets you read an experiment, without possibly mutating it. And `update` remains the implementation of `PUT /v1/experiment`.

Now if someone explicitly specifies a `BaseExperiment` which doesn't exist, it'll just create a random empty experiment, which seems like a weird unintended side-effect. And if the experiment does exist, won't we overwrite existing fields like `isPublic`?

js/src/logger.ts
```diff
@@ -1642,6 +1655,87 @@ function validateAndSanitizeExperimentLogFullArgs(
...
+  [Symbol.iterator]() {
+    return this.fetch();
+  }
+
+  async fetchedData() {
+    if (this._fetchedData === undefined) {
+      const state = await this.getState();
+      const resp = await state.logConn().get(`object/${this.objectType}`, {
+        id: await this.id,
+        fmt: "json",
```
manugoyal, line 1700: If you fetch the `json2` format, it'll do all this deep-json-deserialization for you and exclude `_object_delete` automatically.

js/src/logger.ts
```diff
@@ -1642,6 +1655,87 @@ function validateAndSanitizeExperimentLogFullArgs(
   return event;
 }
 
+class ObjectFetcher<RecordType> {
+  private excludeFields: Set<string> = new Set(["_object_delete"]);
+  private _fetchedData: any[] | undefined = undefined;
```
manugoyal, line 1660: Why isn't the type `RecordType[] | undefined`?

js/src/logger.ts
```diff
@@ -1642,6 +1655,87 @@ function validateAndSanitizeExperimentLogFullArgs(
...
+      yield Object.fromEntries(
+        Object.entries(record)
+          .filter(([k, _]) => !this.excludeFields.has(k))
+          .map(([k, v]) => [
+            k,
+            this.jsonFields.has(k) && typeof v === "string" ? JSON.parse(v) : v,
+          ])
+      ) as RecordType;
+    }
+    this.clearCache();
```
manugoyal, line 1688: Why does using the fetch iterator trigger a cache clear?

js/src/logger.ts
```diff
@@ -1642,6 +1655,87 @@ function validateAndSanitizeExperimentLogFullArgs(
...
+  clearCache() {
+    this._fetchedData = undefined;
+  }
+
+  public async version() {
+    if (this.pinnedVersion !== undefined) {
+      return this.pinnedVersion;
+    } else {
+      const fetchedData = await this.fetchedData();
+      let maxVersion = undefined;
```
manugoyal, line 1722: Maybe more tersely:
```
const maxVersion = Math.max(...fetchedData.map((r) => r[TRANSACTION_ID_FIELD] ?? -Infinity));
return (maxVersion === -Infinity) ? undefined : maxVersion;
```

js/src/logger.ts
```diff
@@ -1642,6 +1655,87 @@ function validateAndSanitizeExperimentLogFullArgs(
...
+        .map((x: string) => JSON.parse(x));
+    }
+    return this._fetchedData || [];
+  }
+
+  clearCache() {
+    this._fetchedData = undefined;
+  }
+
+  public async version() {
```
manugoyal, line 1717: super nit: we're being a bit inconsistent about marking methods public here.

js/src/logger.ts
```diff
@@ -1642,6 +1655,87 @@ function validateAndSanitizeExperimentLogFullArgs(
...
+        if (maxVersion === undefined || (xactId ?? xactId > maxVersion)) {
+          maxVersion = xactId;
+        }
+      }
+      return maxVersion;
+    }
+  }
+}
+
+export interface ExperimentIdentifier {
```
manugoyal, line 1734: unused

manugoyal (commented):
nice. maybe the main points are about this `open` vs `update` interface thing, and whether we can use the new `json2` format in the object fetcher to cut down on some logic.

js/src/logger.ts
manugoyal: BTW if you are looking for an endpoint to retrieve an existing experiment by name, look no further than `GET /v1/experiment`, which lets you specify a query param `experiment_name`.

js/src/framework.ts
ankrgyl: Yeah it lets typescript believe the types are being used

js/src/framework.ts
ankrgyl: Generally, yes, but our users _never_ expect this to be the case (`_lazy_load` is not something they can ever trigger). So I don't think it makes sense to pollute the function's type.

js/src/framework.ts
ankrgyl: I had originally written `if ("_type" in dataResult && dataResult._type === "BaseExperiment")` which wasn't enough to convince the compiler that we've covered all cases where `dataResult` can be a `BaseExperiment`

js/src/framework.ts
ankrgyl: will let you refactor this at some point :)

js/src/logger.ts
ankrgyl: I think javascript doesn't complain because it's just a type. The simpler thing to do would be to define it in logger.ts and import it in framework.ts

js/src/logger.ts
ankrgyl: it has to be post-processed in `fetch()` before it's `RecordType`

js/src/logger.ts
ankrgyl: Hmm, I copied this over from the last thing, but it's likely an incorrect relic of debugging code. Will remove.

js/src/logger.ts
ankrgyl: we can go through this later (i just copy/pasted as they currently are)

js/src/logger.ts
ankrgyl: I'm going to leave as is, in the spirit of not introducing unnecessary refactoring

js/src/logger.ts
ankrgyl: removed

core/py/src/braintrust_core/util.py
ankrgyl: I'll just add a comment that it's shallow

js/src/logger.ts
ankrgyl: Ok NVM, I hit a bug with my approach and am convinced :). Fixing

manugoyal: Yeah I feel strongly about this one because I think it would make it more confusing and difficult to cleanly expose those two functions separately without making a backwards-incompatible change, or creating a function which sounds similar to `open` but does something different. I'm just thinking from the perspective of creating parity between the existing SDKs and the new REST API.

core/py/src/braintrust_core/util.py
manugoyal: Seems harmless to also assert that `isinstance(type(v), field.type)`. When would someone want to invoke this with a type that's not the same as the declared dataclass type.

js/src/framework.ts
```diff
@@ -210,10 +243,44 @@ export async function runEvaluator(
...
+    }
+    let name = dataResult.name;
+    if (isEmpty(name)) {
+      const baseExperiment = await experiment.fetchBaseExperiment();
+      if (!baseExperiment) {
+        throw new Error("BaseExperiment() failed to fetch base experiment");
+      }
+      name = baseExperiment.name;
+    }
+    dataResult = init(evaluator.projectName, {
```
manugoyal, line 267: Maybe for future-proofing we should import `init` as `_init_raw`, and create a helper inside `framework.ts` which opens it with `setCurrent: false`.

manugoyal (approved):

core/py/src/braintrust_core/util.py
ankrgyl: ok will add

js/src/framework.ts
ankrgyl: will do

core/py/src/braintrust_core/util.py
ankrgyl: Unfortunately `type` doesn't really seem to be what we want:


```
In [4]: dataclasses.fields(EvalCase)[0]
Out[4]: Field(name='input',type=~Input,default=<dataclasses._MISSING_TYPE object at 0x102c87cd0>,default_factory=<dataclasses._MISSING_TYPE object at 0x102c87cd0>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),kw_only=False,_field_type=_FIELD)

In [5]: dataclasses.fields(EvalCase)[0].type
Out[5]: ~Input

In [6]: isinstance(None, dataclasses.fields(EvalCase)[0].type)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[6], line 1
----> 1 isinstance(None, dataclasses.fields(EvalCase)[0].type)

TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

PR #78, opened 2023-12-23
JS API: Make all data objects lazily-initialized.

manugoyal:
Essentially we accomplish this by logging promises instead of materialized objects. The logger resolves all the promises before logging. Intermediate steps share promises, which should just resolve to the same value once they are awaited, so we're basically using them as a "lazily-computed value".

We also make the async part of the Logger API optional by introducing a generic parameter `IsAsyncFlush`. By default, it is false, so all logging operations will flush and are thus async. But if the user passes the argument `asyncFlush: true` (which should cause typescript to infer the type of `IsAsyncFlush` as `true` as long as they provide a literal), it will stop flushing automatically and the type signatures will no longer be wrapped in promises.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -151,7 +151,9 @@ declare global {
 class BraintrustState {
   public id: string;
   public currentExperiment: IsoAsyncLocalStorage<Experiment | undefined>;
-  public currentLogger: IsoAsyncLocalStorage<Logger | undefined>;
+  // Note: the value of IsAsyncFlush doesn't really matter here, since we
+  // (safely) dynamically cast it whenever retrieving the logger.
+  public currentLogger: IsoAsyncLocalStorage<Logger<false> | undefined>;
   public currentSpan: IsoAsyncLocalStorage<Span>;
 
   public apiUrl: string | null;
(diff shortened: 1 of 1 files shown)


js/src/logger.ts
```diff
@@ -1460,23 +1478,43 @@ export class Experiment {
...
-    this.name = name;
+    this.lazyMetadata = lazyMetadata;
     this.dataset = dataset;
     this.bgLogger = new BackgroundLogger();
     this.lastStartTime = getCurrentUnixTimestamp();
 
     unterminatedObjects.addUnterminated(this, iso.getCallerLocation());
   }
 
+  public get id(): Promise<string> {
```
manugoyal, line 1494: Technically this is a breaking change as we have changed these from bare values to promises. Not sure there's a good way around this unless we make the lazy init opt-in for all of these objects and do template parameterize everything to let people turn it on. But I feel like lazy init is the correct default so maybe we just eat the one-time cost?

ankrgyl: I think it's likely fine, but I'm not sure there's much value in lazily initializing these objects (specifically, experiments and datasets), either, since the `init` functions are already async?

ankrgyl: If we keep this, then when we land this change, we should just send a notification to customers in case they rely on this (although AFAIK, most, if not all, have moved to the `Eval` framework).

ankrgyl (commented):
looks great, especially for the logger. Left a reply on the proposed breaking change for experiments and datasets

js/src/logger.ts
manugoyal: Yeah I think I'm still leaning towards landing it. Seems like a nice feature, and I imagine the breakage is relatively benign. We do have https://linear.app/braintrustdata/issue/BRA-119/experiment-and-dataset-initialization-time-should-be-instant, which this solves (I can do a similar thing for python without changing the API at all).

ankrgyl: SGTM!

ankrgyl (approved):

PR #81, opened 2023-12-24
Implement lazy-login for python SDK

manugoyal:
Analogous to https://github.com/braintrustdata/braintrust-sdk/pull/78.

### DIFF
--- a/py/src/braintrust/util.py
+++ b/py/src/braintrust/util.py
@@ -76,7 +76,7 @@ class LazyValue(Generic[T]):
         if self.has_computed:
             return self.value
         if self.mutex:
-            self.mutex.lock()
+            self.mutex.acquire()
         try:
             if not self.has_computed:
                 res = self.callable()
@@ -85,4 +85,4 @@ class LazyValue(Generic[T]):
             return self.value
         finally:
             if self.mutex:
-                self.mutex.unlock()
+                self.mutex.release()
--- a/py/src/braintrust/util.py
+++ b/py/src/braintrust/util.py
@@ -1,7 +1,9 @@
 import inspect
 import os.path
 import sys
+import threading
 import urllib.parse
+from typing import Any, Callable, Generic, TypeVar
 
 from requests import HTTPError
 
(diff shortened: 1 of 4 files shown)


py/src/braintrust/util.py
```diff
@@ -50,3 +52,31 @@ def get_caller_location():
 # https://stackoverflow.com/questions/5574702/how-do-i-print-to-stderr-in-python.
 def eprint(*args, **kwargs):
     print(*args, file=sys.stderr, **kwargs)
+
+
+T = TypeVar("T")
+
+
+class LazyValue(Generic[T]):
```
ankrgyl, line 60: why not just use `cached_property` from functools?

https://docs.python.org/3/library/functools.html#functools.cached_property

Since our minimum python version is 3.8, we should have it everywhere

py/src/braintrust/util.py
```diff
@@ -50,3 +52,31 @@ def get_caller_location():
...
+    """A simple wrapper around a callable object which computes the value
+    on-demand and saves it for future retrievals.
+    """
+
+    def __init__(self, callable: Callable[[], T], use_mutex: bool):
+        self.callable = callable
+        self.mutex = threading.RLock() if use_mutex else None
```
ankrgyl, line 67: is there potentially some benefit to using an `RWLock` here? Also why does it need to be re-entrant? In case a lazy value depends on itself?

ankrgyl (commented):
Looks like a good approach overall to me

py/src/braintrust/util.py
manugoyal: Yeah there is some good stuff in functools. If I understand correctly, `@cached_property` seems specifically intended for "getter/setter" methods in classes (not really what we're doing). It's also not thread-safe, so we'd need to include that ourselves. Maybe `@cached`/`@lru_cache` is closer to what we want, but I feel like they are a bit overkill (we don't need arguments, and we don't always need thread safety).

I thought that this was a simple-enough util that I'd rather roll it myself.

py/src/braintrust/util.py
manugoyal: Hmm, you're right that re-entrancy is unnecessary overhead here. I'm not sure RWLock is the right solution to reduce overhead (would we acquire a read lock and then "upgrade" it if the value hasn't been computed yet?)

I'm thinking the right thing to do is a short-circuit check, like
```
if self.has_computed:
  return self.value
with self.lock:
  if not self.has_computed:
    ...
return self.value
```

Pretty sure we're not going to run into sequential consistency issues in python (i.e if I read `self.has_computed == True` I'm guaranteed to pick up the new value of `self.value`)? Couldn't find docs but I think the GIL should ensure that?

py/src/braintrust/util.py
ankrgyl: Yeah that makes sense / works for me

py/src/braintrust/util.py
ankrgyl: Yes it will ensure sequential consistency. I think that approach, with a plain mutex (or even re-entrant), is fine, since now the average overhead is much much lower.

ankrgyl (approved):

PR #82, opened 2023-12-30
Undo infer `force_login`.

manugoyal:
We introduced logic in
https://github.com/braintrustdata/braintrust-sdk/pull/4 to infer `force_login=True` if the login params you provide are different from what we have currently logged in as.

But this logic ended up creating a bug in the eval framework, where if you login initially with custom params here
(https://github.com/braintrustdata/braintrust-sdk/blob/844d83b7df2de4b09dd03176391964fa25cae046/py/src/braintrust/cli/eval.py#L202), then we would try to login again whenever creating an experiment here (https://github.com/braintrustdata/braintrust-sdk/blob/844d83b7df2de4b09dd03176391964fa25cae046/py/src/braintrust/framework.py#L396), and the second login could trigger a `force_login` because it uses whatever is in the environment variables.

While there are other ways to work around this, we felt like just asking users to pass `force_login=True` explicitly to login if their params changed should suffice and is simple.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -979,6 +979,7 @@ type InitLoggerOptions<IsAsyncFlush> = {
   apiUrl?: string;
   apiKey?: string;
   orgName?: string;
+  forceLogin?: boolean;
   setCurrent?: boolean;
 } & AsyncFlushArg<IsAsyncFlush>;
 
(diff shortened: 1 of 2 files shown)


ankrgyl (approved):

js/src/logger.ts
```diff
@@ -979,6 +979,7 @@ type InitLoggerOptions<IsAsyncFlush> = {
   apiUrl?: string;
   apiKey?: string;
   orgName?: string;
+  forceLogin?: boolean;
```
manugoyal, line 982: Shoved this in too, so that our unit tests which exercise the lazy-login functionality could force a login.

PR #86, opened 2024-01-09
Small updates to core lib to include additional system fields.

ankrgyl:


core/js/src/db_fields.ts
```diff
@@ -2,3 +2,12 @@ export const TRANSACTION_ID_FIELD = "_xact_id";
 export const OBJECT_DELETE_FIELD = "_object_delete";
 export const CREATED_FIELD = "created";
 export const IS_MERGE_FIELD = "_is_merge";
+export const ID_FIELD = "id";
+
+export const ALL_ROW_ID_FIELDS = [
+  TRANSACTION_ID_FIELD,
+  OBJECT_DELETE_FIELD,
```
manugoyal, line 9: this one isn't in py? Also the ordering is different, if that matters.

Also maybe a comment to explain what `ALL` means in this case? I would think if it's all the IDs necessary to identify a row in the DB, then we're missing some.

manugoyal: oh I guess it's just all the fields that we have individually listed out above.

manugoyal (approved):

PR #87, opened 2024-01-10
Fix null scores

ankrgyl:
Fix typedefs and a few checks

### DIFF
--- a/core/py/src/braintrust_core/score.py
+++ b/core/py/src/braintrust_core/score.py
@@ -1,6 +1,6 @@
 import dataclasses
 from abc import ABC, abstractmethod
-from typing import Dict
+from typing import Dict, Optional
 
 from .util import SerializableDataClass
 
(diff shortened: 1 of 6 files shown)


core/py/src/braintrust_core/score.py
```diff
@@ -20,7 +20,7 @@ def as_dict(self):
         }
 
     def __post_init__(self):
-        if self.score < 0 or self.score > 1:
+        if self.score is not None and (self.score < 0 or self.score > 1):
```
manugoyal, line 23: Can we change the type signature of score to `Optional[float]`?

manugoyal (approved):
let's add some unit tests too.

core/py/src/braintrust_core/score.py
ankrgyl: yep

PR #90, opened 2024-01-17
Trace view editing

ankrgyl:
Small updates to the core library to include merge paths and configurable (by path) merges.


core/py/src/braintrust_core/util.py
```diff
@@ -24,19 +24,27 @@ def coalesce(*args):
...
+    """Merges merge_from into merge_into, destructively updating merge_into. Does not merge any further than
+    merge_paths."""
 
     if not isinstance(merge_into, dict):
         raise ValueError("merge_into must be a dictionary")
     if not isinstance(merge_from, dict):
         raise ValueError("merge_from must be a dictionary")
 
     for k, merge_from_v in merge_from.items():
+        full_path = path + [k]
```
manugoyal, line 37: nit: this could also be a tuple because we are never mutating the underlying list. So you could avoid the`tuple` conversion on line 39.

manugoyal (approved):

core/py/src/braintrust_core/util.py
ankrgyl: done

PR #92, opened 2024-01-23
Track span types

ankrgyl:
If the `type` attribute is set, we'll now print some useful information in the UI

### DIFF
--- a/core/py/src/braintrust_core/span_types.py
+++ b/core/py/src/braintrust_core/span_types.py
@@ -1,0 +1,10 @@
+from enum import Enum
+
+
+class SpanTypeAttribute(str, Enum):
+    LLM = "llm"
+    SCORE = "score"
+    FUNCTION = "function"
+    EVAL = "eval"
+    TASK = "task"
+    TOOL = "tool"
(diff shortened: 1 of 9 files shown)


core/py/src/braintrust_core/span_types.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum
+
+
+class SpanTypeAttribute(str, Enum):
```
manugoyal, line 4: Isn't there StrEnum to make this easier? https://docs.python.org/3/library/enum.html#enum.StrEnum

js/src/oai.ts
```diff
@@ -16,6 +17,10 @@ interface OpenAILike {
   beta?: BetaLike;
 }
 
+declare global {
```
manugoyal, line 20: What is this used for?

py/src/braintrust/logger.py
```diff
@@ -888,6 +889,11 @@ def decorator(span_args, span_kwargs, f):
 
         f_sig = inspect.signature(f)
 
+        if "span_attributes" not in span_kwargs:
```
manugoyal, line 892: What if they pass `span_attributes` as a positional argument? If we wanted to be very strict about making it a keyword-only argument, we could add an asterisk to the function signature, so it becomes `def start_span(name=None, *, span_attributes={}, ...)`.

manugoyal (approved):

manugoyal (approved):

PR #96, opened 2024-01-26
Only send git metadata that's configured as allowed

ankrgyl:
Now that we support configuring which git metadata you can collect at an org-wide level, this change updates the SDK to respect that configuration, and further allows users to optionally override it (merging the two settings into the _least_ permissive setting).


ankrgyl:
Optional field changes LGTM (I'll let you do the final "approve" on the PR)

PR #100, opened 2024-02-01
Update "output" field on datasets to "expected"

aphinx:
Updates the SDK to use `expected` instead of `output` for dataset records.

The object fetcher and logger now attempt to use new API endpoints (`logs3/` and `object3/`) that use `expected` instead of `output` for datasets. If the request errors, we fall back to the old endpoints, patching the data to use `output` accordingly.


py/src/braintrust/logger.py
```diff
@@ -1163,12 +1173,29 @@ def fetched_data(self):
...
-            response_raise_for_status(resp)
+            data = None
+            try:
+                resp = _state.log_conn().get(
+                        f"object3/{self.object_type}", params={"id": self.id, "fmt": "json2", "version": self._pinned_version, "api_version": 2}
+                )
+                response_raise_for_status(resp)
+                data = resp.json()
+            except Exception as e:
+                # When hitting old versions of the API where the "object3/" endpoint isn't available, fall back to
```
ankrgyl, line 1184: add DEPRECATION_NOTICE to this comment. It's not a perfect convention but it's convenient to grep for to find stuff like this.

py/src/braintrust/logger.py
```diff
@@ -1752,7 +1784,7 @@ class Dataset(ObjectFetcher):
     You should not create `Dataset` objects directly. Instead, use the `braintrust.init_dataset()` method.
     """
 
-    def __init__(self, lazy_metadata: LazyValue[ProjectDatasetMetadata], version: Union[None, int, str] = None):
+    def __init__(self, lazy_metadata: LazyValue[ProjectDatasetMetadata], version: Union[None, int, str] = None, include_output_alias=True):
```
ankrgyl, line 1787: if this is true, should we swap `expected` -> `output`? Otherwise if they `json.dumps` it'll have both fields.

py/src/braintrust/logger.py
```diff
@@ -1743,6 +1770,11 @@ def __exit__(self, type, value, callback):
         self.end()
 
 
+def add_dataset_output_alias(r):
+    r["output"] = r.get("expected")
```
ankrgyl, line 1774: If you happened to receive data via the legacy endpoint, then wouldn't `output` be set and this would override it to `None`?

core/py/src/braintrust_core/legacy_api.py
```diff
@@ -0,0 +1,29 @@
...
+def make_legacy_dataset(r):
+    row = {**r}
+    expected = row.pop("expected", None)
+    if "output" not in row:
+        row["output"] = expected
+    return row
+
+
+def make_legacy_record(r):
+    if "dataset_id" not in r:
```
ankrgyl, line 27: nice, very clever

py/src/braintrust/logger.py
ankrgyl: Should we just use

```
def patch_legacy_dataset(r):
    row = {**r}
    output = row.pop("output", None)
    if "expected" not in row:
        row["expected"] = output
    return row
```

here?

core/py/src/braintrust_core/legacy_api.py
```diff
@@ -0,0 +1,29 @@
+import json
+
+
+def patch_legacy_dataset(r):
+    row = {**r}
+    output = row.pop("output", None)
+    if "expected" not in row:
```
manugoyal, line 7: nit: it might read more clearly to write it as
```
if "output" in row and "expected" not in row:
    row["expected"] = row.pop("output")
```

just so the reader doesn't have to understand the implicit assertion that the row cannot contain both output and expected

core/py/src/braintrust_core/legacy_api.py
```diff
@@ -0,0 +1,29 @@
...
+
+def patch_legacy_record(r):
+    if "dataset_id" not in r:
+        return r
+    return patch_legacy_dataset(r)
+
+
+def make_legacy_dataset(r):
+    row = {**r}
+    expected = row.pop("expected", None)
```
manugoyal, line 20: similarly here

js/src/logger.ts
```diff
@@ -779,8 +780,27 @@ type BackgroundLogEvent =
 export interface DatasetRecord {
   id: string;
   input: any;
-  output: any;
+  expected?: any;
   metadata: any;
+  output?: any;
```
manugoyal, line 785: Maybe helpful to be super strict about types here and define it as
```
export type DatasetRecord =
    & { id, input, metadata }
    & ({ expected?: any } | { output?: any })
```

js/src/logger.ts
```diff
@@ -779,8 +780,27 @@ type BackgroundLogEvent =
...
+}
+
+function patchLegacyDataset(r: DatasetRecord): DatasetRecord {
```
manugoyal, line 788: Should we stick these in `core` as well similar to python?

js/src/logger.ts
```diff
@@ -855,9 +875,8 @@ class BackgroundLogger {
           break;
         }
 
-        const itemS = JSON.stringify(item);
-        items.push(itemS);
-        itemsLen += itemS.length;
+        items.push(item);
```
manugoyal, line 878: Hmm so now we're stringifying the JSON objects twice: once for computing `itemsLen` and then another time when creating `dataS`. It might be preferable to fake the construction of `dataS` as
```
dataS = `{"rows": ${constructJsonArray(itemsS)}, "api_version": 2}`
```

manugoyal: I recognize that we have to manipulate the object itself for the legacy endpoint, but maybe we can only incur the extra work there. Like we could keep track of both `items` and `itemsS`, and fallback to patching and re-stringifying `items` if they are on the old endpoint.

js/src/logger.ts
```diff
@@ -2301,9 +2350,12 @@ export class Dataset extends ObjectFetcher<DatasetRecord> {
 
   constructor(
     lazyMetadata: LazyValue<ProjectDatasetMetadata>,
-    pinnedVersion?: string
+    pinnedVersion?: string,
+    outputInsteadOfExpected?: boolean,
   ) {
-    super("dataset", pinnedVersion);
+    const outputInsteadOfExpectedBool = outputInsteadOfExpected || outputInsteadOfExpected === undefined;
```
manugoyal, line 2356: Alternatively, you can use the "null-ish" coalescing operator `outputInsteadOfExpected ?? true`, which will only use the right hand operand if the left is null or undefined.

manugoyal (approved):
lgtm!

js/src/logger.ts
aphinx: Explicitly typed both versions instead

js/src/logger.ts
```diff
@@ -925,7 +849,7 @@ type InitOpenOption<IsOpen extends boolean> = {
 export type InitOptions<IsOpen extends boolean> = {
   experiment?: string;
   description?: string;
-  dataset?: Dataset;
+  dataset?: Dataset<boolean>;
```
manugoyal, line 852: Could use `AnyDataset` here and elsewhere?

js/src/logger.ts
```diff
@@ -1198,13 +1126,14 @@ type InitDatasetOptions = {
  * @param options.apiKey The API key to use. If the parameter is not specified, will try to use the `BRAINTRUST_API_KEY` environment variable. If no API
  * key is specified, will prompt the user to login.
  * @param options.orgName (Optional) The name of a specific organization to connect to. This is useful if you belong to multiple.
+ * @param options.outputInsteadOfExpected Unless set to false, records will be fetched from this dataset in the legacy record format, with "output" as an alias for the "expected" field. This will default to false in a future version of Braintrust.
```
manugoyal, line 1129: "as an alias" kind of makes it sound like we'll have both "output" and "expected". Maybe `with the "expected" field renamed to "output"`?

manugoyal (approved):
lgtm!

core/js/src/object.ts
```diff
@@ -0,0 +1,181 @@
...
+  const row = {
+    ...r,
+    expected: r.output,
+  };
+  delete row.output;
+  return row;
+}
+
+export function makeLegacyEvent(e: BackgroundLogEvent): BackgroundLogEvent {
+  if (!("dataset_id" in e) || !("expected" in e)) {
```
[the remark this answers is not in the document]
aphinx: I didn't do that because `ensureLegacyDatasetRecord` is typed to handle only dataset records, but these logged "events" also have `<object>_id`, `created`, and other stuff. I suppose there's some type foo I could do to share the functions, but not sure if worth doing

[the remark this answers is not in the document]
manugoyal: ah gotcha. Yeah the way you have it is fine then.

PR #105, opened 2024-02-05
Don't log full embedding vector from EmbeddingWrapper

aphinx:
Just remove the vector and replace it with `{"embedding_length": <length>}` for now. We can add a flag for it later if it turns out to be useful.


ankrgyl:
Make sure to update the typescript SDK as well

ankrgyl (approved):

PR #108, opened 2024-02-08
Improve type ergonomics for metadata type in scorer

ankrgyl:
* Now the metadata type is inferred from the data
* If it's always specified, you don't have to make it `?` in scorer functions


js/src/framework.ts
```diff
@@ -376,7 +402,9 @@ export async function runEvaluator(
           meta({ scores: scoreMetadata });
         }
 
-        rootSpan.log({ scores, metadata });
+        // Note: We're asserting here that any object can be cast as Record<string, unknown>, which should generally
+        // be true, but if we discover that it's not, we may want to update the definition of BaseMetadata.
+        rootSpan.log({ scores, metadata: metadata as Record<string, unknown> });
```
vitorbal, line 407: Seems like a fair assumption to me 👍

js/src/framework.ts
```diff
@@ -140,9 +161,14 @@ declare global {
 
 globalThis._evals = {};
 
-export async function Eval<Input, Output, Expected>(
+export async function Eval<
+  Input,
+  Output,
+  Expected,
+  Metadata extends BaseMetadata,
```
manugoyal, line 168: I think this breaks existing evals which don't specify anything for metadata. When I build this branch and compile one of our internal tests, I see the following error:
```
tests/expect_tests/basic.eval.ts:5:3 - error TS2322: Type '() => Promise<{ input: string; expected: string; }[]>' is not assignable to type 'EvalData<string, string, object>'.
  Type '() => Promise<{ input: string; expected: string; }[]>' is not assignable to type '() => EvalCase<string, string, object>[]'.
    Type 'Promise<{ input: string; expected: string; }[]>' is missing the following properties from type 'EvalCase<string, string, object>[]': length, pop, push, concat, and 29 more.

5   data: async () => {
    ~~~~

  sdk/js/dist/framework.d.ts:37:5
    37     data: EvalData<Input, Expected, Metadata>;
           ~~~~
    The expected type comes from property 'data' which is declared here on type 'Evaluator<string, string, string, object>'
```

The chunk it's complaining about is here:
```
  Eval("basic", {
E   data: async () => {     ■ Type '() => Promise<{ input: string; expected: string; }[]>' is not assignable to type 'EvalData<string, string, object>'.    Type '() => Promise<{ input: string; expected: string; }[]>' is not assignable to type '() => EvalCase<string, string, object>[]'.      Type 'Promise<{ input: string; expected: string; }[]>' is missing the
      return [
        {
          input: "c",
          expected: "d",
        },
      ];
    },
```

My guess is that restricting `Metadata` to be a subtype of `BaseMetadata` prevents it from being nothing (`void`). It seems this can be fixed by changing `BaseMetadata = object` to `BaseMetadata = object | void`.

manugoyal: Also I think this will have to be defaulted to `Metadata extends BaseMetadata = void`, otherwise existing callsites which have type specs for eval as `Eval<something, something, something>` will complain that there is no fourth argument.

js/src/logger.ts
```diff
@@ -1764,11 +1762,11 @@ class ObjectFetcher<RecordType> {
   }
 }
 
-export interface EvalCase<Input, Expected> {
+export type BaseMetadata = object;
```
manugoyal, line 1765: out of curiosity, why `object` instead of `Record<string, unknown>`? The latter seems to compile for me.

js/src/framework.ts
ankrgyl: Updated the type signature and updated  `BaseExperiment` and `Eval` to set a default value of void. Do you think we need to update other places too?

js/src/logger.ts
ankrgyl: so that the right half (i.e. `unknown`) can be typed differently for each key

manugoyal: I think `Record<string, unknown>` allows this: https://www.typescriptlang.org/play?ssl=5&ssc=2&pln=1&pc=1#code/MYewdgzgLgBAZiEAuGAlApqATgEwDzRYCWYA5gDQwCuYA1mCAO5gB8MAvDAN4BQM-MAIYoAnCPJ8BAIxQAiQVOCyJAmMBRQsVdBIC+QA

js/src/framework.ts
manugoyal: In theory, to avoid any breakage, we should include defaults for every publicly-exported type/function signature which adds a `Metadata` param. But if we don't expect much breakage outside of `BaseExperiment` and `Eval`, maybe it's not worth the clutter.

js/src/logger.ts
ankrgyl: Ok will update!

js/src/framework.ts
ankrgyl: ok will update!

js/src/framework.ts
```diff
@@ -376,7 +411,9 @@ export async function runEvaluator(
           meta({ scores: scoreMetadata });
         }
 
-        rootSpan.log({ scores, metadata });
+        // Note: We're asserting here that any object can be cast as Record<string, unknown>, which should generally
```
manugoyal, line 414: This whole thing should not be necessary if we type `metadata` as `Record<string, unknown>` on line 341.

manugoyal (approved):
lgtm! Let's make sure CI passes on our main internal repo as well before merging this.

PR #110, opened 2024-02-09
[Python] Support logging of Pydantic models and dataclasses

aphinx:
Also handles nested tuples/lists/dicts containing dataclasses and Pydantic `BaseModel`s.

Fixes BRA-823.


core/py/src/braintrust_core/json.py
```diff
@@ -0,0 +1,60 @@
+import dataclasses
+import json
+from typing import Any, Dict, Optional
+
+from pydantic import BaseModel
```
ankrgyl, line 5: I don't think we can assume the user's environment has `pydantic` (nor do we depend on it). We should probably make this a dynamic import and gracefully fail.

core/py/src/braintrust_core/json.py
```diff
@@ -0,0 +1,60 @@
...
+    except JSONSerializationError:
+        return False
+
+
+def dump_object(obj: Any) -> Any:
+    """Convert an object to a JSON-serializable object if it is an instance of `dataclasses.dataclass` or `pydantic.BaseModel`."""
+    if dataclasses.is_dataclass(obj) and not isinstance(obj, type):
+        return dataclasses.asdict(obj)
+    if isinstance(obj, BaseModel):
+        return obj.model_dump()
```
ankrgyl, line 32: I know that they changed this b/w version 1 and 2. Many environments still use pydantic 1, which has a `dict()` method.

We can potentially just avoid the pydantic stuff altogether and just check if it has a `model_dump()` and then a `dict()` method?

core/py/src/braintrust_core/json.py
```diff
@@ -0,0 +1,69 @@
...
+    # Attempt to dump a Pydantic v1 `BaseModel`.
+    try:
+        return obj.dict()
+    except (AttributeError, TypeError):
+        pass
+
+    return obj
+
+
+def make_json_serializable(obj: Any) -> Any:
```
manugoyal, line 45: I'm a little worried about the efficiency of this function, because it's trying to JSON serialize the object at the very start to check if it's JSON serializable, and then again at the very end. We should be careful (and hopefully very clear) about how many times we try JSON-serializing the objects they give us.

I wonder if an easier way to do all of this is to utilize the `default` method of the json encoding library: https://docs.python.org/3/library/json.html#json.JSONEncoder.

Maybe we could create our own JSON encoder as
```
class BraintrustJSONEncoder(json.JSONEncoder):
    def default(self, o):
        # Your dump_obj implementation
        ...
        # In case none of the cases match
        return super().default(o)
```

And use that everywhere instead of the standard one?

core/py/src/braintrust_core/json.py
```diff
@@ -0,0 +1,69 @@
...
+        check_json_serializable(obj)
+        return True
+    except JSONSerializationError:
+        return False
+
+
+def dump_object(obj: Any) -> Any:
+    """Convert an object to a JSON-serializable object if it is an instance of `dataclasses.dataclass` or `pydantic.BaseModel`."""
+    if dataclasses.is_dataclass(obj) and not isinstance(obj, type):
+        return dataclasses.asdict(obj)
```
manugoyal, line 28: I wonder if we want to recursively invoke `dump_object` on the result of this, in case the dataclass contains a pydantic model or something. Maybe that's too crazy.

core/py/src/braintrust_core/json.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def check_json_serializable(obj: Any) -> None:
+    try:
+        _ = json.dumps(obj)
+    except Exception as e:
+        raise JSONSerializationError from e
+
+
+def is_json_serializable(obj: Any) -> bool:
```
manugoyal, line 17: If the function is only used internally, I'd preface with an underscore.

core/py/src/braintrust_core/json.py
```diff
@@ -0,0 +1,69 @@
...
+    try:
+        check_json_serializable(obj)
+        return True
+    except JSONSerializationError:
+        return False
+
+
+def dump_object(obj: Any) -> Any:
```
manugoyal, line 25: same here about underscore

core/py/src/braintrust_core/json.py
```diff
@@ -0,0 +1,69 @@
...
+        return {
+            make_json_serializable(k): make_json_serializable(v)
+            for k, v in obj.items()
+        }
+    obj = dump_object(obj)
+    check_json_serializable(obj)
+    return obj
+
+
+def make_dict_values_json_serializable(obj: Dict[str, Any]) -> Dict[str, Any]:
```
manugoyal, line 60: Is this a big-enough efficiency gain that it's worth exposing separately from `make_json_serializable`?

aphinx:
OK, simplified this thing a ton. Tests are at https://github.com/braintrustdata/braintrust/pull/1002

core/py/src/braintrust_core/bt_json.py
```diff
@@ -0,0 +1,26 @@
...
+        except (AttributeError, TypeError):
+            pass
+
+        # Attempt to dump a Pydantic v1 `BaseModel`.
+        try:
+            return o.dict()
+        except (AttributeError, TypeError):
+            pass
+
+        return json.JSONEncoder().default(self, o)
```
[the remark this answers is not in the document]
aphinx: hah I copied this line directly from the python docs but I do agree that `super()` is nicer. updated

PR #112, opened 2024-02-12
Move typespecs into `@braintrust/core`

manugoyal:
This way we can use them in more places, such as the SDK itself.


core/js/src/typespecs/app_types.ts
```diff
@@ -0,0 +1,362 @@
+// Type definitions for operating on the app database.
```
manugoyal, line 1: The stuff in `typespecs` can be ignored, as we are just moving the files unmodified from the internal repo.

ankrgyl (approved):

PR #121, opened 2024-02-16
Deprecate error field from `Score` type.

manugoyal:
Aligned with the autoevals change
(https://github.com/braintrustdata/autoevals/pull/56), which bubbles errors back up to the user, rather than swallowing them into the score.

We also change the eval framework to capture errors from individual scorers (rather than fail the whole scoring loop), save them to the printed eval result, and also surface them in the toplevel row's metadata.


js/src/framework.ts
```diff
@@ -512,7 +512,8 @@ export function reportEvaluatorResult(
     if (!verbose && !jsonl) {
       console.error(warning("Add --verbose to see full stack traces."));
     }
-  } else if (summary) {
+  }
+  if (summary) {
```
manugoyal, line 516: This is now aligned with how we do it in the python framework, where even if there are errors in the eval, we still print out the summary.

core/js/src/score.ts
```diff
@@ -2,6 +2,7 @@ export interface Score {
   name: string;
   score: number | null;
   metadata?: Record<string, unknown>;
+  // Note: this field is deprecated, as errors are propagated up to the caller.
```
ankrgyl, line 5: can you add a DEPRECATION_NOTICE to the comment, so when we grep for it, we can find this message?

core/py/src/braintrust_core/score.py
```diff
@@ -10,32 +10,30 @@ class Score(SerializableDataClass):
...
             "score": self.score,
             "metadata": self.metadata,
-            "error": repr(self.error) if self.error else None,
         }
 
     def __post_init__(self):
         if self.score is not None and (self.score < 0 or self.score > 1):
             raise ValueError(f"score ({self.score}) must be between 0 and 1")
+        if self.error is not None:
+            print(
```
ankrgyl, line 26: eprint?

ankrgyl (approved):

PR #125, opened 2024-02-19
Robustness and debuggability improvements to background loggers

manugoyal:
- Add retry logic to the LazyValue resolution when logging.

- Introduce a `sync_flush` parameter to the loggers. If false (the current behavior), any lazy value resolution and logging is run effectively in the background, so that any errors are just printed out after all retries have completed. If true, the background thread is inactive. We still print out the errors when retrying, but after the retries have been exhausted, we throw an exception. This should in theory give users fine-grained control over which/when statements are sent to the network, and gives them the ability to implement their own control flow.

- While we're at it, make the other logger arguments controllable via environment variable.

- While implementing the same functionality in the JS logger, I took a pass at trying to more closely mirror the logic between python and JS.


py/src/braintrust/logger.py
```diff
@@ -1191,8 +1226,6 @@ def __next__(self):
 
 class ObjectFetcher:
     def __init__(self, object_type, pinned_version=None, mutate_record=None):
-        assert hasattr(self, "id"), "ObjectFetcher subclasses must have an 'id' attribute"
```
manugoyal, line 1194: Unrelated bugfix. Removing this because querying the object `id` inadvertently activates the lazy-login logic, making the lazy-login futile.

py/src/braintrust/logger.py
```diff
@@ -1226,10 +1259,11 @@ def fetched_data(self):
         return self._refetch()
 
     def _refetch(self):
+        state = self._get_state()
```
manugoyal, line 1262: This is necessary to make sure we lazy-login before accessing the state.

js/src/logger.ts
```diff
@@ -719,22 +719,40 @@ function constructLogs3Data(items: string[]) {
...
+  private activeFlushResolved = true;
+
+  public syncFlush: boolean = false;
+  public defaultBatchSize: number = 100;
+  public numRetries: number = 3;
 
   constructor(logConn: LazyValue<HTTPConnection>) {
     this.logConn = logConn;
 
+    const syncFlushEnv = Number(iso.getEnv("BRAINTRUST_SYNC_FLUSH"));
```
ankrgyl, line 739: should we document these variables somewhere? Maybe in a troubleshooting section of the eval guide?

ankrgyl (approved):
Seems like a good idea to me

js/src/logger.ts
manugoyal: Yes. Will add in the alignment PR.

PR #127, opened 2024-02-21
Change the type of TransactionId to a string.

manugoyal:
See comment in code for justification (also aligns with our internal codebase's type alteration).

I think this should be backwards compatible with the API and vercel endpoints, because both of the postgres clients seem to accept integer-or-string query arguments and cast them to the appropriate DB type.


ankrgyl (approved):
This is surprisingly surgical. Nice

PR #130, opened 2024-02-27
Tags sdk

ankrgyl:


js/src/logger.ts
```diff
@@ -460,16 +461,22 @@ function logFeedbackImpl(
     throw new Error(`source must be one of ${VALID_SOURCES}`);
   }
 
-  if (isEmpty(scores) && isEmpty(expected) && isEmpty(comment)) {
+  if (
```
manugoyal, line 464: Should we also support logging tags through the rest API? I think we'd need to add them to the typespecs here: https://github.com/braintrustdata/braintrust-sdk/blob/28b5b2b6e88d6eeff6ed047c8a7959a78bb6d8f6/core/js/src/typespecs/api_types.ts#L574

js/src/logger.ts
```diff
@@ -1809,6 +1816,22 @@ function _urljoin(...parts: string[]): string {
   return parts.map((x) => x.replace(/^\//, "")).join("/");
 }
 
+function validateTags(tags: readonly string[]) {
+  if (!tags) {
```
manugoyal, line 1820: This should not be necessary (or use `tags.length`), because an empty array evaluates to true. But also I feel like it's just unnecessary short-circuiting.

manugoyal (approved):

js/src/logger.ts
ankrgyl: will do

js/src/logger.ts
ankrgyl: will just remove

PR #135, opened 2024-03-03
Support reading prompts through the SDK

ankrgyl:
This change includes typespecs (for the REST API) and the python implementation of prompt loading.


ankrgyl:
@manugoyal could you take a look at the typespec changes?

core/js/src/typespecs/app_types.ts
```diff
@@ -158,6 +159,16 @@ export const datasetSchema = z
   .openapi("Dataset");
 export type Dataset = z.infer<typeof datasetSchema>;
 
+export const promptSchema = z.object({
+  id: z.string().uuid(),
```
manugoyal, line 163: Should we include documentation for these fields, as we do in the other API objects? That is one benefit to using the `generateBaseTableSchema` helper function, because it autogenerates docs for the common fields.

core/js/src/typespecs/app_types.ts
```diff
@@ -341,6 +352,28 @@ const patchDatasetSchema = createDatasetSchema
   .strict()
   .openapi("PatchDataset");
 
+const createPromptSchema = z
+  .object({
```
manugoyal, line 356: nit: a shorter-hand way to write this would be `createPromptSchema = promptSchema.omit({ id: true })`

core/js/src/typespecs/prompt.ts
```diff
@@ -0,0 +1,122 @@
+import { z } from "zod";
+
+export const roleSchema = z.enum([
+  "system",
```
manugoyal, line 4: Are these roles supposed to align with the RBAC roles? If it's prompt-specific, maybe `promptRole` to avoid naming confusion?

core/js/src/typespecs/prompt.ts
```diff
@@ -0,0 +1,122 @@
...
+  "user",
+  "assistant",
+  "function",
+  "tool",
+  "model",
+]);
+export type Role = z.infer<typeof roleSchema>;
+
+export const functionCallSchema = z.object({
+  name: z.string(),
```
manugoyal, line 14: Should we add docs to all of these fields? Or if they mirror some other documented schema elsewhere on the internet, we could link to those docs in the toplevel prompt schema.

core/js/src/typespecs/prompt.ts
```diff
@@ -0,0 +1,122 @@
...
+  role: roleSchema,
+  name: z.string().optional(),
+  function_call: z.union([z.string(), functionCallSchema]).optional(),
+  tool_calls: z.array(toolCallSchema).optional(),
+});
+
+export const promptBlockDataSchema = z.union([
+  z.object({
+    type: z.literal("completion"),
+    content: z.string().default(""),
```
manugoyal, line 38: Are you sure these defaults will actually get "applied" if someone sends the data through a REST API? If not, maybe we should apply the defaulting logic in one place ubiquitously.

core/js/src/typespecs/prompt.ts
```diff
@@ -0,0 +1,122 @@
...
+    arguments: z.string(),
+    name: z.string(),
+  }),
+  type: z.literal("function"),
+});
+
+export const messageSchema = z.object({
+  content: z.string().default(""),
+  role: roleSchema,
+  name: z.string().optional(),
```
manugoyal, line 30: Usually I prefer `nullish` to `optional`, because it lets people pass in `null` as the value. In fact a lot of auto-generated SDKs (such as python) will explicitly pass in `None`, which gets converted to `null`, and then rejected by an API spec that only allows for `optional`.

core/js/src/typespecs/prompt.ts
```diff
@@ -0,0 +1,122 @@
...
+export type ModelParams = z.infer<typeof modelParamsSchema>;
+export const promptOptionsSchema = z.object({
+  model: z.string().optional(),
+  params: modelParamsSchema.optional(),
+  position: z.string().optional(),
+});
+
+export type PromptOptions = z.infer<typeof promptOptionsSchema>;
+
+export const promptDataSchema = z.object({
```
manugoyal, line 111: For large objects that may get used in multiple places in an openAPI spec, I like to add a `.openapi("[TypeName]")` annotation, so that the generated spec can define a JSON schema type just for this object, and reference it in other places. Otherwise it'll get copied throughout everywhere it's used.

core/js/src/typespecs/api_types.ts
```diff
@@ -288,7 +289,7 @@ export const fetchEventsRequestSchema = z
   .strict()
   .openapi("FetchEventsRequest");
 
-function makeFetchEventsResponseSchema<T extends z.AnyZodObject>(
+export function makeFetchEventsResponseSchema<T extends z.AnyZodObject>(
```
manugoyal, line 292: Is this still used externally?

core/js/src/typespecs/api_types.ts
```diff
@@ -401,6 +402,26 @@ const projectLogsEventSchema = z
...
+  .object({
+    id: promptEventBaseSchema.shape.id,
+    [TRANSACTION_ID_FIELD]: promptEventBaseSchema.shape[TRANSACTION_ID_FIELD],
+    created: promptEventBaseSchema.shape.created,
+    org_id: projectSchema.shape.org_id,
+    project_id: projectSchema.shape.id,
+    log_id: z
+      .literal("p")
+      .describe("A literal 'p' which identifies the log as a prompt entry"),
+    name: z.string().describe("The name of the prompt"),
```
manugoyal, line 416: nit: might be a good idea to share these fields/docs with the `app_types` definitions.

manugoyal (commented):
typespecs look fine. Just nits

PR #138, opened 2024-03-07
Factor out batch update logic

ankrgyl:
This change simply factors out the request batching logic so it can be used in more places.


manugoyal (approved):
Let's merge this because I'm about to overcomplicate this batching logic in order to fix all the annoying parent_id cross-row dependency issues.

PR #139, opened 2024-03-07
Fix nullable and numeric scores

ankrgyl:
* Update typedefs to support nullable scores in TS and Python
* Support numeric scores in TS and handle nullable scores correctly.

### DIFF
--- a/js/src/framework.ts
+++ b/js/src/framework.ts
@@ -79,6 +79,8 @@ export type EvalScorerArgs<
   output: Output;
 };
 
+type ScoreValue = Score | number | null;
+
 export type EvalScorer<
   Input,
   Output,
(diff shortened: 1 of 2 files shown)


js/src/framework.ts
```diff
@@ -383,10 +385,18 @@ export async function runEvaluator(
...
+                  const scoreValue =
                     scoreResult instanceof Promise
                       ? await scoreResult
                       : scoreResult;
+
+                  if (scoreValue === null) {
+                    return null;
+                  }
+                  const result: Score =
+                    scoreValue === null || typeof scoreValue === "object"
```
manugoyal, line 397: We're checking for null above, so the first case should be unnecessary? Maybe just `typeof scoreValue === 'number'`?

py/src/braintrust/framework.py
```diff
@@ -91,8 +91,8 @@ class EvalScorerArgs(SerializableDataClass):
 
 EvalScorer = Union[
     Scorer,
-    Callable[[Input, Output, Output], Score],
-    Callable[[Input, Output, Output], Awaitable[Score]],
+    Callable[[Input, Output, Output], Optional[Score]],
```
manugoyal, line 94: Should we have symmetrical logic here for allowing raw number outputs for the scorers?

manugoyal (approved):

PR #140, opened 2024-03-08
Fix bug in JS LazyValue

manugoyal:
Since the `.get` method was async, it was possible that multiple callers could invoke it concurrently. Also it was awaiting the result of the callable in the same statement as setting the callable to computed, which allows multiple callers to generate and await the callable concurrently.

The observed result is that the LazyValue function could be invoked multiple times, and we might end up registering multiple experiments. The new implementation should guarantee we invoke the callable at most once.

### DIFF
--- a/js/src/util.ts
+++ b/js/src/util.ts
@@ -49,6 +49,11 @@ export class LazyValue<T> {
     if (this.value.hasComputed) {
       return this.value.val;
     }
+    // Note that we do not want to await the Promise returned by the callable
+    // inside `get` before setting `hasComputed` to true, because that would
+    // allow multiple async tasks to invoke `.get` concurrently and potentially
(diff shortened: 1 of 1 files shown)


ankrgyl (approved):

js/src/util.ts
```diff
@@ -35,19 +35,21 @@ export function isEmpty(a: unknown): a is null | undefined {
...
     this.callable = callable;
   }
 
-  async get(): Promise<T> {
+  get(): Promise<T> {
     if (this.value.hasComputed) {
       return this.value.val;
     }
-    this.value = { hasComputed: true, val: await this.callable() };
+    this.value = { hasComputed: true, val: this.callable() };
```
ankrgyl, line 52: can youa dd a comment here along the lines of the PR description

PR #145, opened 2024-03-11
Make it easier to specify span type

ankrgyl:
Just like `name`, you can now specify `type` at the top-level.


core/js/src/span_types.ts
```diff
@@ -6,3 +6,6 @@ export enum SpanTypeAttribute {
   TASK = "task",
   TOOL = "tool",
 }
+
+// Make SpanType any of the values of SpanTypeAttribute, eg llm, score, function, eval, task, tool
```
manugoyal, line 10: nit: the docstring feels a bit unnecessary, but fine either way.

py/src/braintrust/logger.py
```diff
@@ -88,12 +88,15 @@ def log_feedback(self, **event):
...
-    def start_span(self, name=None, span_attributes={}, start_time=None, set_current=None, parent_id=None, **event):
+    def start_span(
+        self, name=None, type=None, span_attributes={}, start_time=None, set_current=None, parent_id=None, **event
+    ):
         """Create a new span. This is useful if you want to log more detailed trace information beyond the scope of a single log event. Data logged over several calls to `Span.log` will be merged into one logical row.
 
         We recommend running spans within context managers (`with start_span(...) as span`) to automatically mark them as current and ensure they are ended. Only spans run within a context manager will be marked current, so they can be accessed using `braintrust.current_span()`. If you wish to start a span outside a context manager, be sure to end it with `span.end()`.
 
         :param name: Optional name of the span. If not provided, a name will be inferred from the call stack.
+        :param type: Optional type of the span. If not provided, the type will be unset.
```
manugoyal, line 99: The type is a bit unclear to me. Is it supposed to be an enum or a string? Seems like enum, but in typescript we allow strings. In python there is a `Literal` type that could annotate that it must be of a subset of strings, and then we could convert that to the enum / log a warning.

py/src/braintrust/logger.py
```diff
@@ -1852,11 +1877,15 @@ def __init__(
...
 
+        type_arg = dict()
+        if type is not None:
+            type_arg = dict(type=type)
         self.internal_data = dict(
             metrics=dict(
                 start=start_time or time.time(),
             ),
-            span_attributes=dict(**span_attributes, name=name, exec_counter=exec_counter),
+            # Set type first, in case they override it in `span_attributes`.
```
manugoyal, line 1887: Should we do that for name as well? Technically a breaking change, but it feels pretty low-likelihood that someone specifies a name inside `span_attributes` instead of the hoisted name param and expects it to get overriden by the default name.

py/src/braintrust/logger.py
```diff
@@ -1852,11 +1877,15 @@ def __init__(
             _EXEC_COUNTER += 1
             exec_counter = _EXEC_COUNTER
 
+        type_arg = dict()
+        if type is not None:
+            type_arg = dict(type=type)
```
manugoyal, line 1882: Maybe we should clear out `None` fields in `internal_data` as well in python in `SpanImpl.log`. Seems harmless and lets us avoid extra boilerplate like this.

manugoyal (approved):

core/js/src/span_types.ts
ankrgyl: That was a conversation I was having with Copilot lol. Removed

py/src/braintrust/logger.py
ankrgyl: added a clearer message. We should probably add type annotations to all the arguments here.

py/src/braintrust/logger.py
ankrgyl: Yeah good idea

py/src/braintrust/logger.py
ankrgyl: Good idea

ankrgyl:
@manugoyal do you mind sanity checking my changes (specifically the internal data piece)

manugoyal (approved):
lgtm

PR #158, opened 2024-03-25
Support setting BRAINTRUST_APP_PUBLIC_URL

ankrgyl:


manugoyal (approved):
sure. What is the motivation here? Are people running configurations where the SDK can access the webapp with a URL that regular browser consumers cannot?

PR #161, opened 2024-03-29
Only use the data plane if BRAINTRUST_API_URL is set

ankrgyl:


core/js/typespecs/app_types.ts
```diff
@@ -227,11 +227,17 @@ const experimentBaseSchema = generateBaseTableSchema("experiment", {
...
     id: experimentBaseSchema.shape.id,
-    project_id: experimentBaseSchema.shape.project_id,
+    project_id: experimentBaseSchema.shape.project_id.nullish(),
+    project_name: projectSchema.shape.name.nullish(),
+    org_id: projectSchema.shape.org_id.nullish(),
     name: experimentBaseSchema.shape.name,
     description: experimentBaseSchema.shape.description,
     created: experimentBaseSchema.shape.created,
     repo_info: repoInfoSchema.nullish(),
+    ancestor_commits: z
```
manugoyal, line 237: Do we really want to return this for every operation which CRUDs an experiment?

core/js/typespecs/app_types.ts
```diff
@@ -331,6 +337,9 @@ const patchProjectSchema = z
 const createExperimentSchema = z
   .object({
     project_id: experimentSchema.shape.project_id,
+    project_name: experimentSchema.shape.project_name,
+    org_id: experimentSchema.shape.org_id,
+    ancestor_commits: experimentSchema.shape.ancestor_commits,
```
manugoyal, line 342: We have to be careful to remove any fields we don't want to modify from the `patchExperimentSchema`.

ankrgyl:
@manugoyal can you weigh in on this new approach? tl;dr:

* Directly route `/api` routes to the control plane (from the data plane)
* If `BRAINTRUST_API_URL` is set explicitly, then do everything through the dataplane.

manugoyal:
> @manugoyal can you weigh in on this new approach? tl;dr:
> 
> * Directly route `/api` routes to the control plane (from the data plane)
> * If `BRAINTRUST_API_URL` is set explicitly, then do everything through the dataplane.

Sounds good to me. On the second point, I can imagine two reasons why we don't want to just route everything through the dataplane right now:
1. Since it would require a new backend endpoint, doing that would force people to upgrade their backend to use the new SDK.
2. We want to avoid the unnecessary hop through the data plane when possible.

Reason (1) is fine. But I worry about reason (2), because would it imply it may be undesirable to gradually refator the SDKs to go through the REST API managed in the data plane, instead of directly through the control plane?

ankrgyl:
> > @manugoyal can you weigh in on this new approach? tl;dr:
> > 
> > * Directly route `/api` routes to the control plane (from the data plane)
> > * If `BRAINTRUST_API_URL` is set explicitly, then do everything through the dataplane.
> 
> Sounds good to me. On the second point, I can imagine two reasons why we don't want to just route everything through the dataplane right now:
> 
> 1. Since it would require a new backend endpoint, doing that would force people to upgrade their backend to use the new SDK.
> 2. We want to avoid the unnecessary hop through the data plane when possible.
> 
> Reason (1) is fine. But I worry about reason (2), because would it imply it may be undesirable to gradually refator the SDKs to go through the REST API managed in the data plane, instead of directly through the control plane?

I'm actually fine with routing everything through the dataplane. I just dont want to rock the boat too much with this change.

py/src/braintrust/logger.py
```diff
@@ -958,6 +958,9 @@ def check_updated_param(varname, arg, orig):
 
         conn = None
         if api_key is not None:
+            # If the log_url is set, via BRAINTRUST_API_URL, then use that to login.
```
manugoyal, line 961: Would a less tricky way to accomplish this be to set `_state.app_url = os.environ.get("BRAINTRUST_API_URL", app_url)` on line 954? Then we wouldn't need to change line 184 either?

ankrgyl: Yep

PR #163, opened 2024-03-29
Broaden toplevel `log` error to having called `traced` anywhere.

manugoyal:
We are seeing this error more broadly than just calling `experiment.log` within an active span. Generally we think most users don't want to mix traced and non-traced logging in the same object ever, so we broaden the check to fire if it ever happens.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -593,7 +593,7 @@ export class Logger<IsAsyncFlush extends boolean> {
   ): PromiseUnless<IsAsyncFlush, string> {
     if (this.calledStartSpan && !options?.allowConcurrentWithSpans) {
       throw new Error(
-        "Cannot run toplevel `log` method while using spans. To log to the span, call `traced` and then log with `span.log`"
+        "Cannot run toplevel `log` method while using spans. To log to the span, call `logger.traced` and then log with `span.log`"
       );
     }
 
@@ -2187,7 +2187,7 @@ export class Experiment extends ObjectFetcher<ExperimentEvent> {
   ): string {
     if (this.calledStartSpan && !options?.allowConcurrentWithSpans) {
       throw new Error(
-        "Cannot run toplevel `log` method while using spans. To log to the span, call `traced` and then log with `span.log`"
+        "Cannot run toplevel `log` method while using spans. To log to the span, call `experiment.traced` and then log with `span.log`"
       );
     }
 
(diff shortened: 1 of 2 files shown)


js/src/logger.ts
```diff
@@ -582,26 +584,20 @@ export class Logger<IsAsyncFlush extends boolean> {
...
-        checkCurrentSpan instanceof SpanImpl &&
-        checkCurrentSpan.parentObject === this
-      ) {
-        throw new Error(
-          "Cannot run toplevel Logger.log method while there is an active span. To log to the span, use Span.log"
-        );
-      }
+    if (this.calledStartSpan && !options?.allowConcurrentWithSpans) {
+      throw new Error(
+        "Cannot run toplevel Logger.log method while using spans. To log to the span, use Span.log"
```
ankrgyl, line 596: I think we should recommend that they call `startSpan()` and then log with `span.log()`

manugoyal: Updated.

py/src/braintrust/logger.py
```diff
@@ -2446,16 +2462,14 @@ def log(
...
         """
-        if not allow_log_concurrent_with_active_span:
-            check_current_span = current_span()
-            if getattr(check_current_span, "parent_object", None) == self:
-                raise Exception(
-                    "Cannot run toplevel Logger.log method while there is an active span. To log to the span, use Span.log"
-                )
+        if self._called_start_span and not allow_concurrent_with_spans:
+            raise Exception(
+                "Cannot run toplevel `log` method while using spans. To log to the span, call `traced` and then log with `span.log`"
```
ankrgyl, line 2469: isn't `experiment.start_span()` the more likely method?

ankrgyl (approved):

py/src/braintrust/logger.py
manugoyal: Idk don't most people start spans in python using `traced`? This will throw if they used `@traced` anywhere and then try to call `experiment.log`. This mirrors the example we use in the docs.

ankrgyl: Well I think if you're running an experiment, the pattern is something like:

```
for row in rows:
   with experiment.start_span() as span:
       ...
       output = task(input) # this is @traced
       ...
       span.log(input, output, ...)
```

PR #165, opened 2024-03-30
Remove strictness requirements on API typespecs.

manugoyal:
These typespecs are part of the OpenAPI spec that we publish, and while it's nice to have automated checks in the backend API server to ensure the user doesn't specify unnecessary properties, this strictness makes it hard to add properties to various operations without forcing folks to upgrade their backend.

So we remove the strictness requirements now. We can consider adding the checks back later, perhaps in a more ad-hoc way, if they help users.

### DIFF
--- a/core/js/typespecs/api_types.ts
+++ b/core/js/typespecs/api_types.ts
@@ -285,7 +285,6 @@ export const fetchEventsRequestSchema = z
     filters: fetchFiltersSchema.nullish(),
     version: versionSchema.nullish(),
   })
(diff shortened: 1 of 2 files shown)


ankrgyl (approved):

manugoyal:
I'm thinking of going in the opposite direction with https://github.com/braintrustdata/braintrust-sdk/pull/166. I think it makes sense to keep the strictness checks in general, but only remove them for our internal backend response validation.

PR #167, opened 2024-03-31
Constrain score lists to be objects

ankrgyl:
Non-objects (nulls and numbers) are not particularly usable, since we don't know what name to attach to each score. This change forces each returned value to be an object, and returns a runtime error if not (along w/ a type error if you're using typescript).

### DIFF
--- a/js/src/framework.ts
+++ b/js/src/framework.ts
@@ -79,8 +79,7 @@ export type EvalScorerArgs<
   output: Output;
 };
 
-type ScoreValue = Score | number;
-type OneOrMoreScores = ScoreValue | Array<ScoreValue> | null;
+type OneOrMoreScores = Score | number | null | Array<Score>;
 
 export type EvalScorer<
   Input,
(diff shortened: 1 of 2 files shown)


js/src/framework.ts
```diff
@@ -486,21 +485,26 @@ export async function runEvaluator(
                     return null;
                   }
 
-                  const scoreValues = Array.isArray(scoreValue)
+                  if (Array.isArray(scoreValue)) {
+                    for (const s of scoreValue) {
+                      if (!(typeof s === "object" && !isEmpty(s))) {
+                        throw new Error(
+                          `When returning an array of scores, each score must be a non-empty object. Got: ${s}`
```
manugoyal, line 492: Technically passing an empty object will print out `[Object object]`. I wonder if we should print `JSON.stringify(s)`

manugoyal (approved):

js/src/framework.ts
ankrgyl: good idea

PR #176, opened 2024-04-03
Relax prompt data strictness

ankrgyl:
These fields come from external systems

### DIFF
--- a/core/js/typespecs/prompt.ts
+++ b/core/js/typespecs/prompt.ts
@@ -44,31 +44,28 @@ export const promptBlockDataSchema = z.union([
 
 export type PromptBlockData = z.infer<typeof promptBlockDataSchema>;
 
-const braintrustModelParamsSchema = z.strictObject({
+const braintrustModelParamsSchema = z.object({
   use_cache: z.boolean().optional(),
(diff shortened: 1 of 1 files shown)


core/js/typespecs/prompt.ts
```diff
@@ -44,39 +44,36 @@ export const promptBlockDataSchema = z.union([
 
 export type PromptBlockData = z.infer<typeof promptBlockDataSchema>;
 
-const braintrustModelParamsSchema = z.strictObject({
+const braintrustModelParamsSchema = z.object({
```
manugoyal, line 47: Maybe we can use `.passthrough()` explicitly instead of reverting to `.object`. The only reason is that we can then enforce that nobody calls `z.object`. Also a comment calling out that this schema is for data received from an external provider would be nice.

manugoyal (approved):

core/js/typespecs/prompt.ts
manugoyal: Sorry I mean `.strip()`. `.passthrough` will actually expose the field in the returned object which I don't think we need.

PR #177, opened 2024-04-03
Small cleanups to span parent id functionality

manugoyal:
- Use Enum instead of StrEnum due to lack of availability in older pythons

- Add `export` to `NoopSpan` (and thus allow returning None/undefined from the method).

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -144,7 +144,7 @@ export interface Span {
   /**
    * Return a serialized representation of the span that can be used to start subspans in other places. See `Span.traced` for more details.
    */
-  export(): Promise<string | undefined>;
+  export(): Promise<string>;
 
   /**
    * Flush any pending rows to the server.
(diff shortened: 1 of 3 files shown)


ankrgyl (commented):
Should we just return "" instead of `undefined` from the no-op span, so that users don't have to deal with handling undefined everywhere now?

manugoyal:
> Should we just return "" instead of `undefined` from the no-op span, so that users don't have to deal with handling undefined everywhere now?

Good idea. Done.

js/src/logger.ts
```diff
@@ -1911,33 +1920,7 @@ export function traced<IsAsyncFlush extends boolean = false, R = void>(
   callback: (span: Span) => R,
   args?: StartSpanArgs & SetCurrentArg & AsyncFlushArg<IsAsyncFlush>
 ): PromiseUnless<IsAsyncFlush, R> {
-  const { span, isLogger } = ((): { span: Span; isLogger: boolean } => {
```
manugoyal, line 1914: Unrelated cleanup: I noticed that the lower-level `startSpan` function was not looking at the `parent` arg, so I factored out the bit of logic that considers it and call it in both places.

ankrgyl (approved):

PR #178, opened 2024-04-04
Generalize span_info args for oai wrappers.

manugoyal:
In general allow passing any of the params that could be used to create a span. Type-wise, we just include `name` and `span_attributes` in the TS type signatures, but could add anything else in `StartSpanArgs` if needed.

### DIFF
--- a/js/src/oai.ts
+++ b/js/src/oai.ts
@@ -1,5 +1,11 @@
 import { SpanTypeAttribute } from "@braintrust/core";
-import { CompiledPrompt, Span, startSpan, traced } from "./logger";
+import {
+  CompiledPrompt,
+  Span,
+  StartSpanArgs,
+  startSpan,
+  traced,
+} from "./logger";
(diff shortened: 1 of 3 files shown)


js/src/oai.ts
```diff
@@ -144,23 +150,18 @@ function wrapBetaChatCompletion<
...
-      },
-      event: mergeDicts(
+    const { span_info: _, ...params } = allParams;
+    const span = startSpan(
+      mergeDicts(
         {
-          input: messages,
-          metadata: {
-            ...rest,
+          name: "OpenAI Chat Completion",
```
ankrgyl, line 157: should we use this opportunity to rename this to `Chat Completion` (esp since ppl are using it w/ multiple models that are proxied)

ankrgyl (approved):

js/src/oai.ts
manugoyal: Done. I left it as `OpenAI Embedding` but I can change that too if people are using different models there, too.

PR #179, opened 2024-04-04
Use single global BackgroundLogger instance everywhere.

manugoyal:
Originally we would spawn a separate BackgroundLogger for each object we created. This is not particularly useful anymore, because the background logger already batches and parallelizes requests. It also leads to problems with flushing, because you have to call flush on every logger that you have logged to (including the global logger if you are doing distributed tracing).

We change everything to use just the single global logger and add a global flush function for good measure. Fix a few bugs along the way.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -217,10 +217,11 @@ class BraintrustState {
   // (safely) dynamically cast it whenever retrieving the logger.
   public currentLogger: Logger<false> | undefined;
   public currentSpan: IsoAsyncLocalStorage<Span>;
-  // For operations not tied to a particular data object instance, we maintain a
-  // global BackgroundLogger which is refreshed anytime we log in with different
-  // credentials.
-  private _globalBgLogger: BackgroundLogger | undefined = undefined;
+  // Any time we re-log in, we directly update the logConn inside the logger.
+  // This is preferable to replacing the whole logger, which would create the
+  // possibility of multiple loggers floating around, which may not log in a
+  // deterministic order.
+  private _globalBgLogger: BackgroundLogger;
 
   public appUrl: string | null = null;
   public appPublicUrl: string | null = null;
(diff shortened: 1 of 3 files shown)


js/src/logger.ts
```diff
@@ -217,10 +217,11 @@ class BraintrustState {
   // (safely) dynamically cast it whenever retrieving the logger.
   public currentLogger: Logger<false> | undefined;
   public currentSpan: IsoAsyncLocalStorage<Span>;
-  // For operations not tied to a particular data object instance, we maintain a
-  // global BackgroundLogger which is refreshed anytime we log in with different
```
manugoyal, line 221: I found that refreshing the entire logger every time we re-login can lead to multiple loggers floating around, because we'll create one logger before logging in at all, then another one after we log in. Since some records were ending up in the first logger and others in the second, we were no longer logging in deterministic order.

ankrgyl (approved):
nice -- looks surgical

js/src/logger.ts
```diff
@@ -1036,11 +1034,23 @@ class BackgroundLogger {
   private triggerActiveFlush() {
     if (this.activeFlushResolved) {
       this.activeFlushResolved = false;
-      this.activeFlush = this.flushOnce().finally(() => {
```
manugoyal, line 1039: For some reason, changing the logger stuff uncovered this "uncaught promise rejection" issue, where we assign `this.activeFlush` to a promise which has no `catch` clause. Even though we are guaranteed to throw and catch it later, I guess javascript requires you to attach a catch handler upon promise construction.

So to work around this, we always install a `catch` handler which stores the error as an instance variable, and then we throw it from the instance variable.

js/src/logger.ts
```diff
@@ -1960,6 +1973,13 @@ export function startSpan<IsAsyncFlush extends boolean = false>(
   return startSpanAndIsLogger(args).span;
 }
 
+/**
+ * Flush any pending rows to the server.
+ */
+export async function flush(): Promise<void> {
```
manugoyal, line 1979: I added a global flush method in case people want to flush records not attached to any particular toplevel object, such as records produced from distributed tracing. Now all of the flush functions flush the same logger, so technically this is the only flush function anybody should need.

PR #180, opened 2024-04-05
Use REST API endpoints for fetching data.

manugoyal:
Also removed the fallback requests, because after
https://github.com/braintrustdata/braintrust-sdk/pull/159, we require API version 0.0.35, which includes the REST API.

Tested on prod by fetching a very large dataset successfully in the SDK. Verified that things fail without gzip compression by changing the header from `Accept-Encoding: gzip` to `Accept-Encoding: identity`.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -2206,7 +2206,7 @@ class ObjectFetcher<RecordType>
         {
           version: this.pinnedVersion,
         },
-        { headers: { ["Accept-Encoding"]: "gzip" } }
+        { headers: { "Accept-Encoding": "identity" } }
       );
       const data = (await resp.json()).events;
       this._fetchedData = this.mutateRecord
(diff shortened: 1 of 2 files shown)


js/src/logger.ts
```diff
@@ -2195,26 +2201,14 @@ class ObjectFetcher<RecordType>
...
-          fmt: "json2",
+      const resp = await state.logConn().get(
+        `v1/${this.objectType}/${await this.id}/fetch`,
+        {
           version: this.pinnedVersion,
-        });
-        data = await resp.json();
-      }
+        },
+        { headers: { ["Accept-Encoding"]: "gzip" } }
```
ankrgyl, line 2209: are the brackets here necessary?

ankrgyl (approved):
nice!!

js/src/logger.ts
manugoyal: nope. removed.

PR #185, opened 2024-04-12
Add a few more model params

ankrgyl:


core/js/typespecs/prompt.ts
```diff
@@ -80,6 +80,15 @@ const openAIModelParamsSchema = z
           .strip(),
       ])
       .optional(),
+    function_call: z.union([
+      z.literal("auto"),
+      z.literal("none"),
+      z.strictObject({
```
manugoyal, line 86: Should we add `.strip()` here if we are unsure whether there might be more properties than `name`?

manugoyal (approved):
lgtm. Out of curiosity, was the absence of these params causing zod errors? Because I would hope that the `.strip()` annotation will make zod strip out any extraneous fields rather than error.

core/js/typespecs/prompt.ts
ankrgyl: I think it's better to keep it strict, especially since this is a deprecated parameter.

ankrgyl:
> Out of curiosity, was the absence of these params causing zod errors? 

Nope, it was just preventing some param translation stuff from typechecking in the proxy.

PR #190, opened 2024-04-15
Export event schemas

ankrgyl:


core/js/typespecs/api_types.ts
```diff
@@ -304,7 +309,7 @@ function makeFetchEventsResponseSchema<T extends z.AnyZodObject>(
 }
 
 const experimentEventBaseSchema = generateBaseEventOpSchema("experiment");
-const experimentEventSchema = z
+export const experimentEventSchema = z
```
manugoyal, line 312: unnecessary?

core/js/typespecs/api_types.ts
```diff
@@ -1,10 +1,15 @@
 // Type definitions for operating on the api database.
 
 import { extendZodWithOpenApi } from "@asteasolutions/zod-to-openapi";
-import { z } from "zod";
+import { object, z } from "zod";
 extendZodWithOpenApi(z);
 
-import { experimentSchema, datasetSchema, projectSchema } from "./app_types";
+import {
```
manugoyal, line 7: which of these are necessary?

core/js/typespecs/api_types.ts
```diff
@@ -1,10 +1,15 @@
 // Type definitions for operating on the api database.
 
 import { extendZodWithOpenApi } from "@asteasolutions/zod-to-openapi";
-import { z } from "zod";
+import { object, z } from "zod";
```
manugoyal, line 4: unnecessary?

manugoyal (approved):

core/js/typespecs/common_types.ts
```diff
@@ -32,6 +33,7 @@ export const objectTypesWithEvent = z.enum([
   "experiment",
   "dataset",
   "prompt",
+  "prompt_session",
```
manugoyal, line 36: I think this will cause us to generate REST API endpoints for prompt session from https://github.com/braintrustdata/braintrust/blob/b1e0672c1dc1200a0031e997eb908102c61b65be/deployment-internal/openapi/src/api_spec.ts#L727. Not sure if that is intended.

ankrgyl: We do eventually want to have those, but I think I needed to do this to wire things up so I could put `prompt_session` here to begin with. What do you suggest in the short term?

manugoyal: I would just exclude "prompt_session" from that loop in the code I linked.

core/js/typespecs/api_types.ts
ankrgyl: yes

PR #196, opened 2024-04-19
Change `span.export()` to log more information.

manugoyal:
It was pointed out that the current approach to distributed tracing is prone to race conditions between the parent span logger and the subspan logger: if the subspan hits the backend first, it will fail to nest underneath the parent span because the backend tries to resolve the span pointer information by doing a lookup for the parent span id.

To fix this, we avoid the whole backend-pointer-resolution logic by explicitly providing the span id and root span id when logging spans (this was actually what we did originally). We also modify the slug produced by `span.export()` to include the span id and root span id, and encode it as a more-opaque base64 string to reduce string length.


manugoyal:
Note that as part of this change, we remove the `parent_id` argument from the logger interface. It was previously deprecated, and I felt that keeping it in would (1) be cumbersome to manage implementation-wise, because we have to now juggle two different ways of propagating parent information and (2) be prone to the same race conditions as we are solving here. I also figured the number of users utilizing distributed tracing should be relatively low, and the migration path to using `span.export()` rather than `span.id` should be pretty simple.

Happy to revise this if we feel otherwise.

ankrgyl (approved):
Nice job

PR #207, opened 2024-04-28
Update precommit

ankrgyl:


ankrgyl:
I updated prettier-rc to match the main repo while working on a different change, so I figure we might as well get this out of the way now that there will be a few merge conflicts caused by that.

manugoyal (approved):

PR #209, opened 2024-04-30
Add environment variable for writing failed publish payloads to filesystem.

manugoyal:
For debugging. Also fix a few other miscellaneous issues.

Fixes BRA-1264.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -845,7 +845,8 @@ class BackgroundLogger {
   // 6 MB for the AWS lambda gateway (from our own testing).
   public maxRequestSize: number = 6 * 1024 * 1024;
   public defaultBatchSize: number = 100;
-  public numRetries: number = 3;
+  public numTries: number = 3;
+  public failedPublishPayloadsDir: string | undefined = undefined;
 
   constructor(logConn: LazyValue<HTTPConnection>) {
     this.logConn = logConn;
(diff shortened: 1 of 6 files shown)


js/src/logger.ts
```diff
@@ -867,9 +868,16 @@ class BackgroundLogger {
...
-    const numRetriesEnv = Number(iso.getEnv("BRAINTRUST_NUM_RETRIES"));
-    if (!isNaN(numRetriesEnv)) {
-      this.numRetries = numRetriesEnv;
+    const numTriesEnv = Number(iso.getEnv("BRAINTRUST_NUM_RETRIES"));
+    if (!isNaN(numTriesEnv)) {
+      this.numTries = numTriesEnv + 1;
+    }
+
+    const failedPublishPayloadsDirEnv = iso.getEnv(
+      "BRAINTRUST_FAILED_PUBLISH_PAYLOADS_DIR",
```
ankrgyl, line 877: should we set this by default to something, like in their home directory?

ankrgyl (approved):

js/src/logger.ts
manugoyal: Hmm I'm not sure if we want to have the SDK write stuff to a specific location on their filesystem by default. Isn't that part of the reason we had to remove `CACHE_PATH`? But if we're okay with asking people to turn it off explicitly then I guess we could.

ankrgyl: Let's keep it as is (opt-in) for now.

PR #214, opened 2024-05-07
Incorporate .env files in CLI

ankrgyl:
Follows the `.env.*` conventions laid out (and implemented) by Next.js.

For both TS and Py, add a `--env-file` flag to specify an env file

### DIFF
--- a/js/src/cli.ts
+++ b/js/src/cli.ts
@@ -37,6 +37,7 @@ import {
 } from "./framework";
 import { configureNode } from "./node";
 import { isEmpty } from "./util";
+import { loadEnvConfig } from "@next/env";
 
 // This requires require
 // https://stackoverflow.com/questions/50822310/how-to-import-package-json-in-typescript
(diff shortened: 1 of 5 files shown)


js/src/cli.ts
```diff
@@ -800,6 +801,9 @@ async function main() {
 
   const parsed = parser.parse_args();
 
+  // Load the environment variables from the .env files using the same rules as Next.js
+  loadEnvConfig(process.cwd(), true);
```
manugoyal, line 805: We actually run the braintrust CLI in the internal expect tests, so we could add an env file to the expect tests and check that it gets picked up.

manugoyal (approved):

js/src/cli.ts
ankrgyl: Unfortunately the current working directory is the root :(. I poked around with this but don't think it's worth it.

ankrgyl: Actually did this via the `--env-file` arg.

PR #215, opened 2024-05-09
Add some more tuning params to the loggers.

manugoyal:
`BRAINTRUST_ALL_PUBLISH_PAYLOADS_DIR`: Set this to log all payloads we publish (regardless of success or failure) to the specified directory.

`BRAINTRUST_QUEUE_DROP_WHEN_FULL` (in python): configures the logger to drop additional elements instead of doing a blocking insert.

`BRAINTRUST_QUEUE_DROP_EXCEEDING_MAXSIZE` (in javascript): configures the logger to drop elements after exceeding the given size. This is basically a combination of `BRAINTRUST_QUEUE_SIZE` and `BRAINTRUST_QUEUE_DROP_WHEN_FULL`. The reason we can't use the same params in JS is that JS doesn't have a blocking queue functionality.

Fixes BRA-1332, BRA-1215

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -846,7 +846,9 @@ class BackgroundLogger {
   public maxRequestSize: number = 6 * 1024 * 1024;
   public defaultBatchSize: number = 100;
   public numTries: number = 3;
+  public queueDropExceedingMaxsize: number | undefined = undefined;
   public failedPublishPayloadsDir: string | undefined = undefined;
+  public allPublishPayloadsDir: string | undefined = undefined;
 
   constructor(logConn: LazyValue<HTTPConnection>) {
     this.logConn = logConn;
@@ -873,11 +875,25 @@ class BackgroundLogger {
       this.numTries = numTriesEnv + 1;
     }
 
+    const queueDropExceedingMaxsizeEnv = Number(
+      iso.getEnv("BRAINTRUST_QUEUE_DROP_EXCEEDING_MAXSIZE"),
+    );
+    if (!isNaN(queueDropExceedingMaxsizeEnv)) {
+      this.queueDropExceedingMaxsize = queueDropExceedingMaxsizeEnv;
+    }
+
     const failedPublishPayloadsDirEnv = iso.getEnv(
       "BRAINTRUST_FAILED_PUBLISH_PAYLOADS_DIR",
     );
     if (failedPublishPayloadsDirEnv) {
       this.failedPublishPayloadsDir = failedPublishPayloadsDirEnv;
+    }
+
+    const allPublishPayloadsDirEnv = iso.getEnv(
+      "BRAINTRUST_ALL_PUBLISH_PAYLOADS_DIR",
+    );
+    if (allPublishPayloadsDirEnv) {
+      this.allPublishPayloadsDir = allPublishPayloadsDirEnv;
     }
 
     // Note that this will not run for explicit termination events, such as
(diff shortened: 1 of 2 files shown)


js/src/logger.ts
```diff
@@ -889,7 +905,22 @@ class BackgroundLogger {
...
+      const numElementsToAdd = Math.min(
+        Math.max(this.queueDropExceedingMaxsize - this.items.length, 0),
+        items.length,
+      );
+      return [items.slice(0, numElementsToAdd), items.slice(numElementsToAdd)];
+    })();
+    this.items.push(...addedItems);
+    if (droppedItems.length) {
+      console.warn(
+        `Queue is currently full at size ${this.queueDropExceedingMaxsize}. Dropping additional elements`,
```
ankrgyl, line 921: maybe log how many elements were dropped? And if they're dropped, should we log them to the filesystem if `BRAINTRUST_ALL_PUBLISH_PAYLOADS_DIR` or `BRAINTRUST_FAILED_PUBLISH_PAYLOADS_DIR` is true?

py/src/braintrust/logger.py
```diff
@@ -459,13 +472,20 @@ def __init__(self, log_conn: LazyValue[HTTPConnection]):
...
                 self.queue.put_nowait(event)
             except queue.Full:
                 # Notify consumers to start draining the queue.
                 self.queue_filled_semaphore.release()
-                self.queue.put(event)
+                if not self.queue_drop_when_full:
+                    self.queue.put(event)
+                elif not did_notify_drop:
+                    _logger.warning(
+                        f"Queue is currently full at size {self.queue_maxsize}. Dropping additional elements"
```
ankrgyl, line 486: can we print ho w many were dropped, and same comment, if dropped can we save them to the filesystem?

py/src/braintrust/logger.py
```diff
@@ -459,13 +472,20 @@ def __init__(self, log_conn: LazyValue[HTTPConnection]):
...
+                    )
+                    did_notify_drop = True
```
rmann-nflx, line 488: Small nit: Instead of having this as a boolean I'd have it as a "cooldown" so it's printed at most once every X seconds.

But circling back to Ankur's comments, having some signal about the number of elements dropped is useful as well and I am not sure how you would reconcile only logging once/on cooldown and then logging the dropped count.

js/src/logger.ts
manugoyal: Done.

py/src/braintrust/logger.py
manugoyal: Yup makes sense! I implemented a batched logging method that logs every N seconds (where N is configurable). Each time we drop records, we add to the counter, and print a log message if it has been more than N seconds since the last one.

js/src/logger.ts
```diff
@@ -1065,6 +1102,78 @@ class BackgroundLogger {
...
+    if (!(wrappedItems.length && publishPayloadsDir.length)) {
+      return;
+    }
+    try {
+      const allItems = await this.unwrapLazyValues(wrappedItems);
+      const dataStr = constructLogs3Data(
+        allItems.map((x) => JSON.stringify(x)),
+      );
+      for (const payloadDir of publishPayloadsDir) {
+        await BackgroundLogger.writePayloadToDir({
```
ankrgyl, line 1139: should we log where these are? So if someone sees the message in the logs, they can correlate it to the dropped values in the directory?

ankrgyl (approved):
lgtm pending one logging comment

js/src/logger.ts
manugoyal: We could. I guess they are the ones setting the environment variable that tells us to log there, but it doesn't hurt to be more explicit. I guess the problem is to avoid being spammy. So we could add another log-every-n-seconds based message.

PR #216, opened 2024-05-10
Adjustments to typespecs to make stainless happy.

manugoyal:


core/js/typespecs/app_types.ts
```diff
@@ -146,29 +146,36 @@ export const datasetSchema = z
...
+    prompt_data: promptDataSchema
+      .nullish()
+      .describe("The prompt, model, and its parameters"),
+    tags: z
+      .array(z.string())
+      .nullish()
+      .describe("A list of tags for the prompt"),
+    metadata: promptBaseSchema.shape.metadata,
+  })
+  .openapi("Prompt");
```
manugoyal, line 178: Just adding a name to this type, so stainless can generate a `Prompt` object in its SDKs. This also cuts down on some duplication in the openAPI spec.

core/js/typespecs/app_types.ts
```diff
@@ -380,56 +387,51 @@ export const aclObjectTypeEnum = z
...
-    aclObjectSchema
-      .merge(aclGroupObjectSchema)
-      .merge(aclPermissionObjectSchema)
-      .openapi("GroupPermissionAcl"),
-    aclObjectSchema
-      .merge(aclGroupObjectSchema)
-      .merge(aclRoleObjectSchema)
-      .openapi("GroupRoleAcl"),
-  ])
+  .strictObject({
```
manugoyal, line 391: The whole "disjoint union" thing was too complicated for the stainless SDK generator, because it wasn't able to find the "id" property needed for pagination by looking at each of the union members. I felt it might be too confusing to enumerate each possible combination of user/group/permission/role to be worth the type safety, so I just flattened the object.

ankrgyl (approved):

PR #236, opened 2024-05-30
Accept json content type explicitly

ankrgyl:
We noticed that Vercel edge functions appear to do the equivalent of `Accept: ""` which breaks our API handler code. This change explicitly sets the accept header, in the hopes that it will fix that issue.


ankrgyl:
I published an RC version which includes this, and verified that it fixes it!

manugoyal (approved):

PR #237, opened 2024-06-01
Align with internal addition of `restrict_object_type` to each `role_permissions` entry.

manugoyal:


core/js/typespecs/app_types.ts
```diff
@@ -328,9 +347,16 @@ export const roleSchema = z
     description: roleBaseSchema.shape.description,
     deleted_at: roleBaseSchema.shape.deleted_at,
     member_permissions: z
-      .array(permissionEnum)
+      .array(
```
manugoyal, line 350: This is the only change. The rest is just re-ordering.

ankrgyl: is there some back compat issues we need to be careful about?

ankrgyl (approved):

core/js/typespecs/app_types.ts
manugoyal: Nothing in the SDK or backend directly uses the roles API (just forwarding things along to the webapp).

It is true that if someone was using the roles REST API on an old backend it will break once it gets to the webapp, but I think that's okay since roles are still an in-development feature.

PR #243, opened 2024-06-08
Basic bundling

ankrgyl:


core/js/typespecs/app_types.ts
```diff
@@ -173,39 +173,84 @@ export const datasetSchema = z
   .openapi("Dataset");
 export type Dataset = z.infer<typeof datasetSchema>;
 
+export const validRuntimes = ["node"] as const;
```
manugoyal, line 176: Would prefer to define the enums as zod enums, similar to how we do elsewhere in this file. Just

```
export const runtimesSchema = z.enum(["node"]);
export type Runtime = z.infer<typeof runtimeSchema>
```

You can iterate through the list of members with `runtimesSchema.options`.

core/js/typespecs/app_types.ts
```diff
@@ -173,39 +173,84 @@ export const datasetSchema = z
...
+      z.strictObject({ score: z.number() }),
+    ]),
+  }),
+  bundle_id: z.string(),
+});
+export type CodeBundle = z.infer<typeof codeBundleSchema>;
+
+export const functionDataSchema = z.union([
+  z.strictObject({
+    type: z.literal("prompt"),
```
manugoyal, line 235: Maybe a comment stating that for backwards-compatibility reasons, the prompt data is hoisted outside this object.

js/src/logger.ts
```diff
@@ -2714,6 +2714,10 @@ export class ReadonlyExperiment extends ObjectFetcher<ExperimentEvent> {
 
 let executionCounter = 0;
 
+export function newId() {
```
manugoyal, line 2717: What's the point of this?

js/src/framework.ts
```diff
@@ -233,8 +234,14 @@ function initExperiment<IsOpen extends boolean = false>(
...
 }
 
+export type SpanContext = {
+  currentSpan: typeof currentSpan;
+  NOOP_SPAN: typeof NOOP_SPAN;
+};
+
 declare global {
   var _evals: EvaluatorFile;
+  var _spanContext: SpanContext | undefined;
```
manugoyal, line 244: Where is this used?

manugoyal (approved):

js/src/logger.ts
ankrgyl: Just factor this logic out so I don't need to remember how we create a uuid each place we need to create an id.

core/js/typespecs/app_types.ts
ankrgyl: can do

PR #244, opened 2024-06-09
Update project config to include weighted scores

ankrgyl:


core/js/typespecs/app_types.ts
```diff
@@ -786,7 +799,7 @@ const patchProjectScoreSchema = z
     name: projectScoreSchema.shape.name.nullish(),
     description: projectScoreSchema.shape.description,
     score_type: projectScoreSchema.shape.score_type.nullish(),
-    categories: projectScoreSchema.shape.categories,
+    config: projectScoreSchema.shape.categories,
```
manugoyal, line 802: why?

manugoyal (approved):

core/js/typespecs/app_types.ts
ankrgyl: This shouldn't be changed. I briefly flirted iwth renaming the field before realizing how big of an undertaking that would be.

PR #246, opened 2024-06-11
fix(sdk-py): add timeout parameter to Eval

sachinpad:
Currently, must be specified by the Eval function.
We could set a global timeout from the cmd line in the future.

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -207,6 +207,12 @@ class Evaluator:
     Whether to update an existing experiment with `experiment_name` if one exists. Defaults to false.
     """
 
+    timeout: Optional[float] = None
+    """
+    The duration, in seconds, after which to time out the evaluation.
+    Defaults to None, in which case there is no timeout.
+    """
+
 
 @dataclasses.dataclass
 class EvalResultWithSummary(SerializableDataClass):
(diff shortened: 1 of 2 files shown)


py/src/braintrust/framework.py
```diff
@@ -207,14 +207,22 @@ class Evaluator:
...
+    Defaults to None, in which case there is no timeout.
+    """
+
 
 @dataclasses.dataclass
 class EvalResultWithSummary(SerializableDataClass):
     summary: ExperimentSummary
     results: List[EvalResult]
+    error: Optional[Exception] = None
+    exc_info: Optional[str] = None
```
ankrgyl, line 222: we're actually planning to do this per span (rather than at the top-level for an experiment). Is this used just to capture the timeout?

ankrgyl (commented):
If you run in synchronous mode, i.e. use the sync OpenAI client, what happens? I assume it won't cancel the async task? Should we kill the thread pool that we spin up too?

py/src/braintrust/framework.py
sachinpad: Yup, just used to capture the timeout for now. Should we move this up to be at the span level right now?

sachinpad: Propagated TimeoutError, per offline discussion

sachinpad:
> If you run in synchronous mode, i.e. use the sync OpenAI client, what happens? I assume it won't cancel the async task? Should we kill the thread pool that we spin up too?

From the [ThreadPoolExecutor.shutdown docs](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown):

> the entire Python program will not exit until all pending futures are done executing.

I did forcibly kill the process once we've done what we need to in main.

manugoyal (approved):
lgtm. Based on the internal-repo tests, I assume this works.

PR #248, opened 2024-06-12
feat(sdk-py): include experiment id in summary

sachinpad:


ankrgyl (approved):

py/src/braintrust/logger.py
```diff
@@ -1942,6 +1942,7 @@ def summarize(self, summarize_scores=True, comparison_experiment_id=None):
 
         return ExperimentSummary(
             project_name=self.project.name,
+            experiment_id=self.id,
```
manugoyal, line 1945: fyi @ankrgyl: note an interesting consequence of this is that experiment summary with `summarize_scores=False` will now require a network roundtrip in order to resolve the id.

ankrgyl: Oh interesting. Didn't it require one before to get the experiment's name?

manugoyal: Ah yes good point. Never mind.

py/src/braintrust/logger.py
```diff
@@ -2838,6 +2839,8 @@ class ExperimentSummary(SerializableDataClass):
 
     project_name: str
     """Name of the project that the experiment belongs to."""
+    experiment_id: Optional[str]
+    """ID of the experiment. May be None if the eval was run locally."""
```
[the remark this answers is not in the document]
sachinpad: Done

PR #252, opened 2024-06-12
fix(sdk-js): timeout evaluation with user-provided timeout

sachinpad:
Also sets up jest for the SDK dir.

### DIFF
--- a/js/src/framework.ts
+++ b/js/src/framework.ts
@@ -150,6 +150,12 @@ export interface Evaluator<
    * Whether to update an existing experiment with `experiment_name` if one exists. Defaults to false.
    */
   update?: boolean;
+
+  /**
+   * The duration, in milliseconds, after which to time out the evaluation.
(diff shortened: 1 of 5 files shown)


js/src/framework.ts
```diff
@@ -394,6 +400,33 @@ export async function runEvaluator(
...
+): Promise<EvalResultWithSummary<any, any, any, any>> {
+  let result = runEvaluatorInternal(
+    experiment,
+    evaluator,
+    progressReporter,
+    filters,
+  );
+  let timer = async () => {
+    await new Promise((_, r) =>
+      setTimeout(() => {
```
manugoyal, line 412: Would this just immediately run and throw an error if `evaluator.timeout` is undefined?

sachinpad: Oops, fixed

manugoyal (approved):

aphinx:
Fixes BRA-1453

PR #254, opened 2024-06-13
Make span identifier backwards compatible with old version.

manugoyal:
We accomplish this by just reviving the source file for the old format, and in the new format, we'll parse it as old and upgrade it to new if the encoding version number is an older one. We also explicitly name everything as "*V1" and "*V2" for clarity.


core/js/src/span_identifier_v1.ts
```diff
@@ -0,0 +1,172 @@
+// Mirror of core/py/src/braintrust_core/span_identifier_v1.py.
```
manugoyal, line 1: Just a copy of the old `span_identifier.ts` file with everything renamed to suffix with `V1`

core/py/src/braintrust_core/span_identifier_v1.py
```diff
@@ -0,0 +1,139 @@
+# Serialization format for capturing all relevant information about a span
```
manugoyal, line 1: Just a copy of the old `span_identifier.py` file with everything renamed to suffix with `V1`

ankrgyl (approved):

PR #255, opened 2024-06-13
feat(sdk-js): provide option to limit evaluation concurrency

sachinpad:


ankrgyl (commented):
I think it'd be useful to actually limit concurrency _within_ the `Eval()` more than across them (because you may run like 3 evals but each with 100 concurrent examples). If we do that, then we can also enable people who run `Eval()` directly to limit concurrency.

Perhaps to simplify things, within the CLI, if `maxConcurrency` is specified, we just run one `Eval()` block at a time (since we don't know how much data each one has?)

sachinpad:
> I think it'd be useful to actually limit concurrency _within_ the `Eval()` more than across them (because you may run like 3 evals but each with 100 concurrent examples). If we do that, then we can also enable people who run `Eval()` directly to limit concurrency.
> 
> Perhaps to simplify things, within the CLI, if `maxConcurrency` is specified, we just run one `Eval()` block at a time (since we don't know how much data each one has?)

Ok, sounds good!

sachinpad:
Started with the simple case, in which we just limit the concurrency inside an eval by a user-passed-in number.

sachinpad:
Just changed to divide the user-provided concurrency by the number of scorers, which should have the desired effect we discussed

sachinpad:
Changed it back as per discussion with Ameya

js/src/framework.ts
```diff
@@ -673,8 +688,9 @@ async function runEvaluatorInternal(
         },
       });
     }
-  });
-  const results = await Promise.all(evals);
+  }, evaluator.maxConcurrency);
```
AmeyaKhare-at, line 691: looks like based on the async.queue impl, if concurrency is omitted (functionally the same as setting it explicitly to undefined), the concurrency gets set to 1: https://github.com/caolan/async/blob/master/lib/queue.js#L104

might need to change this to the following to preserve current behavior:

```suggestion
  }, evaluator.maxConcurrency ?? data.length);
```

sachinpad: Fixed! Thanks!

AmeyaKhare-at (approved):

PR #263, opened 2024-06-20
Refactor global state into something that can be propagated as a local

ankrgyl:


js/src/logger.ts
```diff
@@ -274,6 +274,27 @@ class BraintrustState {
     this._logConn = null;
   }
 
+  public copyLoginInfo(other: BraintrustState) {
+    this.appUrl = other.appUrl;
```
manugoyal, line 278: How about `appPublicUrl`?

js/src/logger.ts
```diff
@@ -304,17 +325,21 @@ class BraintrustState {
   }
 }
 
-let _state: BraintrustState;
+let _dangerousGlobalState: BraintrustState;
```
manugoyal, line 328: personally I feel like calling it `_globalState` would be enough.

js/src/logger.ts
```diff
@@ -562,7 +588,7 @@ function logFeedbackImpl(
         [IS_MERGE_FIELD]: true,
       };
     });
-    _state.globalBgLogger().log([record]);
+    state.globalBgLogger().log([record]);
```
manugoyal, line 591: maybe we should change the method `globalBgLogger` to just `bgLogger`, since it's not necessarily a global itself anymore? And maybe adjust the toplevel comment [here](https://github.com/braintrustdata/braintrust-sdk/blob/91d79548574f3b5a721a10d5e753e1d003da0a66/js/src/logger.ts#L928).

js/src/logger.ts
```diff
@@ -610,7 +637,7 @@ function spanComponentsToObjectIdLambda(
         "Impossible: computeObjectMetadataArgs not supported for experiments",
       );
     case SpanObjectTypeV2.PROJECT_LOGS:
-      const args = components.computeObjectMetadataArgs;
+      const args = { state, ...components.computeObjectMetadataArgs };
```
manugoyal, line 640: nit: personally feels slightly clearer to have two arguments to `computeLoggerMetadata`, one being the exact struct that is serialized in span components, and another being the state.

js/src/logger.ts
```diff
@@ -1376,19 +1413,21 @@ export function init<IsOpen extends boolean = false>(
...
     const lazyMetadata: LazyValue<ProjectExperimentMetadata> = new LazyValue(
       async () => {
-        await login({
-          orgName: orgName,
-          apiKey,
-          appUrl,
-        });
+        const state =
+          stateArg ??
+          (await login({
```
manugoyal, line 1418: I kind-of imagined this would look like `(stateArg ?? _dangerousGlobalState).login(loginParams)`. And then the public `login` function is just `_dangerousGlobalState.login(loginParams)`. That way the only global thing any of these codepaths need to access is `_dangerousGlobalState`. The other nice thing is that you could just have one `const state = stateArg ?? _dangerousGlobalState` at the beginning of this function and use that everywhere else.

Is there a reason something like that wouldn't work?

manugoyal (approved):
looks great!

js/src/logger.ts
ankrgyl: good catch. also cleared in the other case.

js/src/logger.ts
ankrgyl: changed

js/src/logger.ts
ankrgyl: done

js/src/logger.ts
ankrgyl: Sure

js/src/logger.ts
ankrgyl: oh that's a clever idea. Let me rework.

PR #265, opened 2024-06-20
Fix distributed tracing object id checks in JS

ankrgyl:
If the span components are serialized in, e.g. Python, they may have `null` instead of `undefined` for the object ids.


manugoyal (approved):
thanks

PR #267, opened 2024-06-21
A bunch of small SDK changes

ankrgyl:
* Enable `projectId` param in `Eval()`
* Allowing setting a span's name after initialization
* Change how the `id` in state is initialized. `uuidv4()` breaks certain platforms, like Cloudflare's workerd, if run on initialization.
* Allow non-dict args on prompts (mirroring the same semantics as the UI/playground).
  * Did not implement this in python
* Support streaming usage metrics for OpenAI


js/src/logger.ts
```diff
@@ -155,6 +156,11 @@ export interface Span {
    */
   close(args?: EndSpanArgs): number;
 
+  /**
+   * Set the span's name after it's created
```
manugoyal, line 160: maybe "Set span attributes (such as span name) after it's created"?

js/src/logger.ts
```diff
@@ -2878,6 +2886,13 @@ export class SpanImpl implements Span {
     return this._id;
   }
 
+  public setAttributes(args: Omit<StartSpanArgs, "event">): void {
+    this.internalData.span_attributes = {
```
manugoyal, line 2890: I realize this is confusing, but `internalData` is cleared immediately after it is set because we call `log` immediately after every time. Similarly, this method should probably trigger a `log` otherwise the span attributes may not get logged.

But better yet would be to make a minor refactor to something like this
```
public log(event: ExperimentLogPartialArgs): void {
  this.logInternal({ event });
}

private logInternal({ event, internalData }: { event: ExperimentLogPartialArgs: internalData?: Partial<ExperimentEvent> }) {
  // Basically the current implementation of `log`.
}
```

And then we can get rid of the `internalData` member variable and just call `logInternal` instead.

js/src/logger.ts
```diff
@@ -3332,6 +3348,12 @@ export class Prompt {
       throw new Error("Empty prompt");
     }
 
+    const dictArgParsed = z.record(z.unknown()).safeParse(buildArgs);
+    const variables: Record<string, unknown> = {
+      input: buildArgs,
+      ...(dictArgParsed.success ? dictArgParsed.data : {}),
```
manugoyal, line 3354: Do we really want to silently ignore invalid `buildArgs`? Should we at least log a warning or something?

py/src/braintrust/logger.py
```diff
@@ -135,6 +135,21 @@ def flush(self):
...
+    @abstractmethod
+    def set_attributes(self, name=None, type=None, span_attributes=None):
+        """Set attributes on the span. These attributes will be attached to all log events within the span.
+        The attributes are equivalent to the arguments to start_span
+
+        :param name: Optional name of the span. If not provided, a name will be inferred from the call stack.
+        :param type: Optional type of the span. Use the `SpanTypeAttribute` enum or just provide a string directly.
+        If not provided, the type will be unset.
+        :param span_attributes: Optional additional attributes to attach to the span, such as a type name.
+        :param start_time: Optional start time of the span, as a timestamp in seconds.
```
manugoyal, line 147: these are not actual arguments

manugoyal (approved):
lgtm. Though it might be nice to make a small cleanup to the span logging implementation in the process.

js/src/logger.ts
ankrgyl: We don't ignore it -- we just use it as `input`. Same semantics as the playground UI. A warning seems misplaced, since this is valid behavior.

py/src/braintrust/logger.py
ankrgyl: Good catch, they were since deleted!

js/src/logger.ts
ankrgyl: That seems to work? Can you take a look at the code change and confirm I accomplished your suggestion?

js/src/logger.ts
```diff
@@ -2792,7 +2800,6 @@ export function newId() {
 export class SpanImpl implements Span {
   // `internalData` contains fields that are not part of the "user-sanitized"
   // set of fields which we want to log in just one of the span rows.
-  private internalData: Partial<ExperimentEvent>;
```
manugoyal, line 2795: Maybe move this comment to `logInternal`?

manugoyal (approved):
logInternal lgtm!

js/src/logger.ts
ankrgyl: can do

PR #268, opened 2024-06-21
Fixes to enable usage from Swift via PythonKit.

jeffdav:
To test our swift code, we've been building a prototype that uses the Braintrust Python SDK via [PythonKit](https://github.com/pvieito/PythonKit).

When we past Swift lambda's to `Eval()` (as `PythonFunction`s), places where the SDK uses `inspect.*` need to be tweaked, as the our lambdas show up as builtin `PyCapsule` objects.

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -244,23 +244,39 @@ async def await_or_run(event_loop, f, *args, **kwargs):
                 thread_pool.thread_pool(), run_f, args, kwargs, contextvars.copy_context()
             )
 
+def is_inspectable(obj):
+    """
+    True when obj is inspectable via `inspect`.
+    False when it is, e.g., a native function or builtin.
+    """
+    if obj.__name__ == 'pythonkit_swift_function' or obj.__name__ == "pythonkit_swift_function_with_keywords":
+        return False
+
+    t = type(obj)
+    if t.__module__ == 'builtins' and t.__name__ == 'PyCapsule':
+        return False
+
+    return True
+
 
 async def call_user_fn(event_loop, fn, **kwargs):
-    signature = inspect.signature(fn)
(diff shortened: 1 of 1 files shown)


py/src/braintrust/framework.py
```diff
@@ -811,7 +829,7 @@ async def run_evaluator_task(datum):
             project=evaluator.project_name, experiment=base_experiment.name, open=True, set_current=False
         ).as_dataset()
 
-    if inspect.isfunction(data_iterator):
+    if inspect.isfunction(data_iterator) or not is_inspectable(data_iterator):
```
ankrgyl, line 832: maybe check `not is_inspectable(...)` first, in case the `isfunction` check throws up?

py/src/braintrust/framework.py
```diff
@@ -733,7 +749,9 @@ async def run_evaluator_task(datum):
 
                 # Check if the task takes a hooks argument
                 task_args = [datum.input]
-                if len(inspect.signature(evaluator.task).parameters) == 2:
+                if not is_inspectable(evaluator.task):
+                    task_args.append(hooks)
+                elif len(inspect.signature(evaluator.task).parameters) == 2:
```
ankrgyl, line 754: maybe `or` these together since they do the same thing?

ankrgyl (commented):
Looks pretty good to me! I will kick off our internal tests on it

ankrgyl:
@manugoyal mind taking a quick look too?

py/src/braintrust/framework.py
jeffdav: In my testing it just returns false because the object isn't a python function, which is correct behavior.  More often than not the first clause will be true, which is why I left it first.  Happy to flip it though.

py/src/braintrust/framework.py
jeffdav: Sure.

jeffdav:
Closing as this functionality was added in a different PR.  Thanks all!

PR #270, opened 2024-06-22
Use x-bt-cached instead of x-cached

ankrgyl:


js/src/wrappers/oai.ts
```diff
@@ -203,7 +203,7 @@ interface APIPromise<T> extends Promise<T> {
   withResponse(): Promise<EnhancedResponse>;
 }
 
-export const X_CACHED_HEADER = "x-cached";
+export const X_CACHED_HEADER = "x-bt-cached";
```
manugoyal, line 206: Would it be useful to define this in [here](https://github.com/braintrustdata/braintrust-sdk/blob/main/core/js/src/http_headers.ts) and import the constant in other places? I think the proxy already depends on core, so we'd just need to bump the dependency version [here](https://github.com/braintrustdata/braintrust-proxy/blob/cf094754f9e8d1acecef8e840a3c027da2910439/packages/proxy/package.json#L72).

manugoyal (approved):

PR #272, opened 2024-06-23
More little tweaks

ankrgyl:
* Allow passing an `onStart` callback to `Eval()`.
* Propagate the project id out of an experiment's summary.
* Using "playground prompts", not just "saved prompts" in the `Prompt` object.
* Serializing/deserializing login state.


js/src/logger.ts
```diff
@@ -217,6 +218,24 @@ declare global {
...
+const LOGIN_ATTRIBUTES = [
+  ...REQUIRED_LOGIN_ATTRIBUTES,
+  "loginToken",
+  "orgId",
+  "gitMetadataSettings",
+] as const;
+
+export type SerializedBraintrustState = Pick<
+  BraintrustState,
+  NonNullable<(typeof LOGIN_ATTRIBUTES)[number]>
```
manugoyal, line 236: According to my IDE, this seems to make no difference to the type signature. I still see
```typescript
type SerializedBraintrustState = {
    appUrl: string | null;
    appPublicUrl: string | null;
    orgName: string | null;
    logUrl: string | null;
    loginToken: string | null;
    orgId: string | null;
    gitMetadataSettings?: GitMetadataSettings | undefined;
}
```

Which makes sense because typescript would have rejected your code for passing in possibly-null values like `this.appUrl`, which it doesn't know cannot be null if you are logged in.

If it's important to get this right, something like the following seems to work for me (and gives us the compiler errors we expect):

```
const REQUIRED_LOGIN_ATTRIBUTES = [
  "appUrl",
  "appPublicUrl",
  "orgName",
  "logUrl",
] as const;
const OPTIONAL_LOGIN_ATTRIBUTES = [
  "loginToken",
  "orgId",
  "gitMetadataSettings",
] as const;

type SerializedBraintrustState =
    & { [K in (typeof REQUIRED_LOGIN_ATTRIBUTES)[number]]: NonNullable<BraintrustState[K]> }
    & { [K in (typeof OPTIONAL_LOGIN_ATTRIBUTES)[number]]: BraintrustState[K] }
;
```

js/src/logger.ts
```diff
@@ -287,6 +306,40 @@ export class BraintrustState {
...
+          `Cannot deserialize BraintrustState with empty ${attr} attribute`,
+        );
+      }
+    }
+    const state = new BraintrustState({});
+    for (const attr of LOGIN_ATTRIBUTES) {
+      // Typescript does not have a way of making this dependent on the iteration of the loop
+      state[attr] = serialized[attr] as any;
+    }
+    return state;
```
manugoyal, line 340: Seems a bit weird because `state` will have `loggedIn === false`, but we are initializing it from a serialized state that must have been logged in? Furthermore `state.login` will do nothing because `state.logUrl` is set?

js/src/logger.ts
```diff
@@ -287,6 +306,40 @@ export class BraintrustState {
...
+    for (const attr of REQUIRED_LOGIN_ATTRIBUTES) {
+      if (isEmpty(serialized[attr])) {
+        throw new Error(
+          `Cannot deserialize BraintrustState with empty ${attr} attribute`,
+        );
+      }
+    }
+    const state = new BraintrustState({});
+    for (const attr of LOGIN_ATTRIBUTES) {
+      // Typescript does not have a way of making this dependent on the iteration of the loop
```
manugoyal, line 337: If we want extreme type safety, maybe it would be worthwhile to make the class state something like this?
```
// Instead of separate appUrl, appPublicUrl, etc, member variables.
public loginState: { status: "logged_in"; data: SerializedBraintrustState } |.{ status: "logged_in": data: undefined };
```

This would also make deserialize simpler. In an ideal world, we could just zod-ify this. Note that we have already zod-ified the `GitMetadataSettings` [here](https://github.com/braintrustdata/braintrust-sdk/blob/25708b2d11167437e6de01fa0406dfd3c26bd8d4/core/js/typespecs/app_types.ts#L270).

manugoyal (commented):
looks good. Mostly some thoughts on doing the type signature stuff for `BraintrustState` serialization/deserialization. Maybe we should use zod?

js/src/logger.ts
ankrgyl: this is fixed

js/src/logger.ts
ankrgyl: done

js/src/logger.ts
```diff
@@ -287,6 +302,64 @@ export class BraintrustState {
...
   }
 
+  public serialize(): SerializedBraintrustState {
+    if (!this.loggedIn) {
+      throw new Error(
+        "Cannot serialize BraintrustState without being logged in",
+      );
+    }
+
+    if (
```
manugoyal, line 312: I think you could this copy-pasta and weird object construction stuff in `deserialize` by just storing a SerializedBraintrustState or empty in the class. But this is okay too.

core/js/src/git_fields.ts
```diff
@@ -10,12 +13,27 @@ export interface RepoInfo {
   git_diff?: string;
 }
 
-export type GitFields = Array<keyof RepoInfo>;
-export type CollectMetadata = "all" | "none" | "some";
-export type GitMetadataSettings = {
-  collect: CollectMetadata;
-  fields?: GitFields;
-};
+export const gitFieldsSchema = z.enum([
```
manugoyal, line 16: These are exactly the keys of `repoInfoSchema`. Is it possible to get them from there?

manugoyal (approved):

PR #273, opened 2024-06-23
Use new cache headers

ankrgyl:
Tracks https://github.com/braintrustdata/braintrust-proxy/pull/71

### DIFF
--- a/js/src/wrappers/oai.ts
+++ b/js/src/wrappers/oai.ts
@@ -8,7 +8,6 @@ import {
 } from "../logger";
 import { getCurrentUnixTimestamp, isEmpty } from "../util";
 import { mergeDicts } from "@braintrust/core";
-import { parse } from "path";
 
 interface BetaLike {
   chat: {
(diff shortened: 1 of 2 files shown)


js/src/wrappers/oai.ts
```diff
@@ -203,23 +202,36 @@ interface APIPromise<T> extends Promise<T> {
   withResponse(): Promise<EnhancedResponse>;
 }
 
+const LEGACY_CACHED_HEADER = "x-cached";
```
manugoyal, line 205: Re-iterating from https://github.com/braintrustdata/braintrust-sdk/pull/270#discussion_r1649757262, but would it be useful to move these headers to the core library, so that other libraries can use the constant?

manugoyal (approved):

js/src/wrappers/oai.ts
ankrgyl: Yeah can do.

ankrgyl: I just don't want to directly rely on it via the proxy, because it requires publishing an update to core and explicitly depending on that version. I don't think it's worth the headache to propagate a constant like this.

ankrgyl: Actually they are literally only referenced in these two places. I think I'll just leave it as-is for now.

PR #274, opened 2024-06-24
Add py.typed and fix type signature for @traced

ankrgyl:
* `py.typed` instructs mypy that it can use the types in `braintrust` to do type checking
* The overloaded case helps mypy realize that calls to `@traced` without additional arguments can simply propagate the original type of `F`.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -1303,7 +1303,7 @@ def traced(*span_args, **span_kwargs) -> Callable[[F], F]:
     # We determine if the decorator is invoked bare or with arguments by
     # checking if the first positional argument to the decorator is a callable.
     if len(span_args) == 1 and len(span_kwargs) == 0 and callable(span_args[0]):
-        return decorator(span_args[1:], span_kwargs, cast(span_args[0], F))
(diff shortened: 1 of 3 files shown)


py/src/braintrust/logger.py
```diff
@@ -1288,16 +1296,16 @@ async def wrapper_async(*f_args, **f_kwargs):
...
+            return cast(F, wrapper_async)
         else:
-            return wrapper_sync
+            return cast(F, wrapper_sync)
 
     # We determine if the decorator is invoked bare or with arguments by
     # checking if the first positional argument to the decorator is a callable.
     if len(span_args) == 1 and len(span_kwargs) == 0 and callable(span_args[0]):
-        return decorator(span_args[1:], span_kwargs, span_args[0])
+        return decorator(span_args[1:], span_kwargs, cast(span_args[0], F))
```
manugoyal, line 1306: isn't the argument order "type" then "value"?

manugoyal (approved):
cool. Separately from this PR, would it be helpful for us to run mypy over the sdk/framework?

py/src/braintrust/logger.py
ankrgyl: Yes! It failed a test internally as a result :)

PR #276, opened 2024-06-25
Python eval framework: handle user-defined builtin functions more gracefully.

manugoyal:
Some users want to invoke the eval framework through FFIs which will create function objects that are not inspectable with `inspect.signature`. Instead of crashing, we default to more reasonable behavior.


jeffdav:
Works for me!

ankrgyl (approved):

PR #283, opened 2024-07-01
Org api

ankrgyl:
Add org API endpoints and install the api/proxy URL as part of the setup/update process if they exist in the stack and have changed.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -1096,7 +1096,7 @@ def login(app_url=None, api_key=None, org_name=None, force_login=False):
             return
 
         if app_url is None:
-            app_url = os.environ.get("BRAINTRUST_APP_URL", GLOBAL_BRAINTRUST_APP_URL)
+            app_url = BRAINTRUST_APP_URL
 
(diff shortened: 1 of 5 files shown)


py/src/braintrust/logger.py
```diff
@@ -60,6 +60,8 @@
 Metadata = Dict[str, Any]
 DATA_API_VERSION = 2
 
+BRAINTRUST_APP_URL = os.environ.get("BRAINTRUST_APP_URL", "https://www.braintrust.dev")
```
manugoyal, line 63: should this be `GLOBAL_BRAINTRUST_APP_URL`? if so, don't forget to adjust the import in `py/src/braintrust/cli/install/api.py`

manugoyal (approved):

py/src/braintrust/logger.py
ankrgyl: Whoops. No, I think what we want to do here is to compute the env var in global scope, and use that in both places. So should use `BRAINTRUST_APP_URL` everywhere.

PR #288, opened 2024-07-02
Native support for streams

ankrgyl:
This turned into a fairly large change, which adds support for a variety of different streaming operations. Here's a summary of stuff to look at:

Main:

* Typespecs for manipulating functions/invoking them (these were moved into the repo from the main repo)
* A new `@braintrust/vercel-ai-sdk` package which implements our adapter.
* functions/invoke.ts calls the invoke endpoint
* Propagate proxyUrl as state we derive from logging in
* Use `waitUntil` in Vercel to make turning `asyncFlush: true` safe. We may want to switch the default now...
* Add `BraintrustStream` and some basic tools to parse/transform it. This is the streaming type returned via `invoke`
* Support logging streams

Misc:

* Upgraded bundling in sdk to use `tsup` so we don't need to deal with inline esbuild flags
* Made the `zod` dependency external, so that you can BYO zod and use type inference with the SDK (first used in the new `invoke` function)
* Hand-coded the `braintrust` package's version as 0.0.141, so that the `vercel-ai-sdk` package, when published, uses it. We have some follow up work to do to fix these versions.


core/js/typespecs/functions.ts
```diff
@@ -0,0 +1,103 @@
+import { z } from "zod";
```
ankrgyl, line 1: this code is just moved

integrations/vercel-ai-sdk/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "devDependencies": {
+    "@types/node": "^20.10.5",
+    "tsup": "^8.0.1",
+    "typescript": "^5.3.3",
+    "vitest": "^1.6.0"
+  },
+  "dependencies": {
+    "@ai-sdk/provider": "^0.0.11",
+    "ai": "^3.2.16",
+    "braintrust": "^0.0.141"
```
ankrgyl, line 34: this will use the workspace package when availabile

js/src/logger.ts
```diff
@@ -908,6 +951,7 @@ export class Logger<IsAsyncFlush extends boolean> {
     type Ret = PromiseUnless<IsAsyncFlush, R>;
 
     if (this.asyncFlush) {
+      this.triggerWaitUntilFlush();
```
ankrgyl, line 954: @manugoyal

js/src/logger.ts
```diff
@@ -3215,6 +3279,36 @@ export class SpanImpl implements Span {
     let sanitizedAndInternalData = { ...internalData };
     mergeDicts(sanitizedAndInternalData, sanitized);
 
+    const serializableInternalData: typeof sanitizedAndInternalData = {};
```
ankrgyl, line 3282: @manugoyal

py/src/braintrust/logger.py
```diff
@@ -2207,17 +2226,32 @@ def log_internal(self, event=None, internal_data=None):
             **sanitized_and_internal_data,
             **{IS_MERGE_FIELD: self._is_merge},
         )
-        serialized_partial_record = _check_json_serializable(partial_record)
-        partial_record = json.loads(serialized_partial_record)
-        if "metrics" in partial_record and "end" in partial_record["metrics"]:
-            self._logged_end_time = partial_record["metrics"]["end"]
 
-        if len(partial_record.get("tags", [])) > 0 and self.span_parents:
+        serializable_partial_record = {}
```
ankrgyl, line 2230: @manugoyal

js/src/logger.ts
manugoyal: Also since this function is getting quite complicated, it'd be nice to separate into smaller sub-functions, or use scoping to indicate the lifetime of variables. Like maybe we can write it as
```
const [serializableInternalData, lazyInternalData] = (() => {
})();
```

because `sanitizedAndInternalData` is not used elsewhere.

js/src/functions/stream.ts
```diff
@@ -0,0 +1,204 @@
...
+   * Get the underlying ReadableStream.
+   *
+   * @returns The underlying ReadableStream<BraintrustStreamChunk>.
+   */
+  public toReadableStream(): ReadableStream<BraintrustStreamChunk> {
+    return this.stream;
+  }
+
+  /**
+   * Get the final value of the stream. This will return a promise that resolves
```
manugoyal, line 68: The term "final value" is ambiguous to me. I read a bunch of the PR thinking that this meant the last chunk returned by the stream, but it's actually "the concatenation of all stream chunks". Maybe clarify this somewhere, and/or call this something like `fullStreamContents`?

js/src/functions/stream.ts
```diff
@@ -0,0 +1,204 @@
...
+   * when the stream is closed, and contains the final value of the stream. Multiple
+   * calls to `finalValue()` will return the same promise, so it is safe to call
+   * this multiple times.
+   *
+   * This function consumes the stream, so if you need to use the stream multiple
+   * times, you should call `copy()` first.
+   *
+   * @returns A promise that resolves with the final value of the stream.
```
manugoyal, line 76: or `undefined` if the stream is empty.

js/src/functions/stream.ts
```diff
@@ -0,0 +1,204 @@
...
+    return this.stream;
+  }
+
+  /**
+   * Get the final value of the stream. This will return a promise that resolves
+   * when the stream is closed, and contains the final value of the stream. Multiple
+   * calls to `finalValue()` will return the same promise, so it is safe to call
+   * this multiple times.
+   *
+   * This function consumes the stream, so if you need to use the stream multiple
```
manugoyal, line 73: Should we add some sort of basic check-and-error protection for this? Seems like an easy footgun.

js/src/functions/stream.ts
```diff
@@ -0,0 +1,204 @@
+import { callEventSchema } from "@braintrust/core/typespecs";
```
manugoyal, line 1: This class seems worth unit-testing.

js/src/logger.ts
```diff
@@ -3215,6 +3279,36 @@ export class SpanImpl implements Span {
...
+              .toReadableStream()
+              .pipeThrough(createFinalValuePassThroughStream(resolve))
+              .pipeTo(devNullWritableStream());
+          });
+        });
+      } else if (value instanceof ReadableStream) {
+        lazyInternalData[key] = new LazyValue(async () => {
+          return await new Promise((resolve) => {
+            value
+              .pipeThrough(createFinalValuePassThroughStream(resolve))
```
manugoyal, line 3303: Is this safe? What if someone logs a `ReadableStream` of their own somewhere which doesn't contain one of the types admissible by `createFinalValuePassThroughStream`? Is it possible to check for this sort-of compatibility up front (or force people to use `BraintrustStream` everywhere)? Otherwise it seems like the `createFinalValuePassThroughStream` code will just assume it's a `BraintrustStreamChunk` if not string or `Uint8Array`.

js/src/logger.ts
```diff
@@ -3215,6 +3279,36 @@ export class SpanImpl implements Span {
     let sanitizedAndInternalData = { ...internalData };
     mergeDicts(sanitizedAndInternalData, sanitized);
 
+    const serializableInternalData: typeof sanitizedAndInternalData = {};
+    const lazyInternalData: Record<string, LazyValue<unknown>> = {};
+
+    for (const [key, value] of Object.entries(sanitizedAndInternalData) as [
```
manugoyal, line 3285: Do we care about pulling out streams in nested components of the object, such as within the `metadata`? I wonder if it would be ideal to stick this logic into the `serializedPartialRecord` json `replacer` function, where if we encounter a stream, we replace it with `undefined` and store the stream in `lazyInternalData`. And then deep-merge it back in?

py/src/braintrust/logger.py
manugoyal: As long as it mirrors the JS implementation, this seems fine / same questions/comments apply.

js/src/logger.ts
ankrgyl: Changed the type signatures, skipped the refactoring (open to doing that at the end, perhaps).

core/js/typespecs/functions.ts
```diff
@@ -0,0 +1,111 @@
...
+    prompt_session_id: z.string().describe("The ID of the prompt session."),
+    prompt_session_function_id: z
+      .string()
+      .describe("The ID of the function in the prompt session."),
+    version: z.string().optional().describe("The version of the function."),
+  }),
+]);
+export type FunctionId = z.infer<typeof functionIdSchema>;
+
+// NOTE: After a few attempts to make this work with strictObject, I gave up, which means
```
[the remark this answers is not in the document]
ankrgyl: fixed

js/src/logger.ts
```diff
@@ -845,6 +875,18 @@ export class Logger<IsAsyncFlush extends boolean> {
     return SpanObjectTypeV2.PROJECT_LOGS;
   }
 
+  private triggerWaitUntilFlush() {
+    // If we're running in Vercel, waitUntil() will ensure that the logs flush
+    // before the function dies (but perhaps after it returns). This makes it safe
+    // to use `asyncFlush: true` on Vercel.
+    //
+    // If the syncFlush flag is true, however, then the user has asked us _never_ to
```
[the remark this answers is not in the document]
ankrgyl: I don't trust process.exit nor do I trust it being called at the end of the lifecycle of a Vercel serverless function.

js/src/logger.ts
```diff
@@ -2499,13 +2548,20 @@ function startSpanAndIsLogger<IsAsyncFlush extends boolean = false>(
     return {
       span,
       isLogger: components.objectType === SpanObjectTypeV2.PROJECT_LOGS,
+      // We don't actually know if it's async flush in this case, so just propagate
```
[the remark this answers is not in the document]
ankrgyl: updated

js/src/logger.ts
```diff
@@ -2476,7 +2525,7 @@ export function setFetch(fetch: typeof globalThis.fetch): void {
 
 function startSpanAndIsLogger<IsAsyncFlush extends boolean = false>(
   args?: StartSpanArgs & AsyncFlushArg<IsAsyncFlush> & OptionalStateArg,
-): { span: Span; isLogger: boolean } {
+): { span: Span; isLogger: boolean; isAsyncFlush: boolean | undefined } {
```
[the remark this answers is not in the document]
ankrgyl: sure

js/src/logger.ts
```diff
@@ -2548,7 +2605,11 @@ function _check_org_info(
 }
 
 function _urljoin(...parts: string[]): string {
-  return parts.map((x) => x.replace(/^\//, "")).join("/");
+  return parts
```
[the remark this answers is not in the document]
ankrgyl: done

js/src/functions/stream.ts
ankrgyl: I don't think so. Final is what Vercel uses

js/src/functions/stream.ts
ankrgyl: updated

js/src/functions/stream.ts
ankrgyl: Maybe, but I'd rather not over engineer it yet (easy to add that later).

js/src/functions/stream.ts
ankrgyl: I think it's pretty well covered by the tests and is just a zod schema. Is there something specific you had in mind?

js/src/logger.ts
ankrgyl: Yep. If it's a braintrust stream, it parses the value. Otherwise it just forwards along the text.

js/src/logger.ts
ankrgyl: Maybe but also would prefer not to over engineer it until we know more about the pattern.

js/src/functions/stream.ts
manugoyal: Ok but i think it should be clarified in the docs.

js/src/functions/stream.ts
manugoyal: sorry I meant the `BraintrustStream` class. I was commenting on the file as a whole. If you think it's well-covered by existing tests then it's fine.

js/src/logger.ts
manugoyal: I was just thinking that the `ReadableStream` type signature seems to allow any type of data (it actually becomes `any` in this code), but we are assuming in the stream transformer that it is of type `string | Uint8Array | BraintrustStreamChunk`. Is that assumption necessarily safe, or maybe it's uncommon-enough that a weird crash is okay?

I'm kind-of just wondering why we need to handle the case of a bare `ReadableStream` in the first place?

js/src/logger.ts
manugoyal: I think refactoring would be nice. The function is getting quite large and complicated. Assuming it's covered by tests, there's minimal risk.

manugoyal (approved):

js/src/logger.ts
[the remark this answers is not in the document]
ankrgyl: Just added it to `triggerActiveFlush()`

PR #292, opened 2024-07-03
Add trace wrapper

ankrgyl:
This wrapper resembles the `@braintrust.traced` decorator in Python.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -3227,7 +3227,7 @@ export class SpanImpl implements Span {
     };
     const serializedPartialRecord = JSON.stringify(partialRecord, (k, v) => {
       if (v instanceof SpanImpl) {
-        return `<span id=${v.id}>`;
+        return `<span>`;
       } else if (v instanceof Experiment) {
         return `<experiment>`;
       } else if (v instanceof Dataset) {
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -2348,7 +2348,29 @@ export function traced<IsAsyncFlush extends boolean = false, R = void>(
   }
 }
 
-export function trace<
+/**
+ * Wrap a function with `traced`, using the arguments as `input` and return value as `output`.
+ * Any functions wrapped this way will automatically be traced, similar to the `@traced` decorator
+ * in Python. If you want to correctly propagate the function's name and define it in one go, then
+ * you can do so like this:
+ *
+ * ```ts
+ * const myFunc = wrapTraced(async function myFunc(input) {
+ *  const result = await client.chat.completions.create({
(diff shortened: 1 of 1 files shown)


js/src/logger.ts
```diff
@@ -2348,6 +2348,47 @@ export function traced<IsAsyncFlush extends boolean = false, R = void>(
...
+    return ((...fnArgs: Parameters<F>) =>
+      traced((span) => {
+        span.log({ input: fnArgs });
+        const output = fn(...fnArgs);
+        if (output instanceof Promise) {
+          output.then((result) => span.log({ output: result }));
+        } else {
+          span.log({ output: output });
+        }
+        return output;
```
manugoyal, line 2375: It feels weird to chain something to a promise (on line 2371) and then return the un-chained object. Because now someone who chains something to the output promise will be forking the promise chain rather than waiting for `span.log` to finish, which feels weird.

Maybe it should read
```
if (output instanceof Promise) {
  return (async () => {
    const result = await output;
    span.log({ output: result });
    return result;
  })();
} else {
  span.log({ output });
  return output;
}
```

js/src/logger.ts
```diff
@@ -2348,6 +2348,47 @@ export function traced<IsAsyncFlush extends boolean = false, R = void>(
   }
 }
 
+export function trace<
```
manugoyal, line 2351: We should probably have documentation for this function, similar to the `@traced` decorator.

manugoyal (approved):
seems nice. I do worry that the name `trace` sounds too similar to `traced`, and since they have fairly different signatures, people might get confused.

Maybe something like `wrapTraced` would fit better alongside `traced`?

js/src/logger.ts
```diff
@@ -2348,6 +2348,47 @@ export function traced<IsAsyncFlush extends boolean = false, R = void>(
...
+        }
+        return output;
+      }, spanArgs)) as IsAsyncFlush extends false ? never : F;
+  } else {
+    return ((...fnArgs: Parameters<F>) =>
+      traced(async (span) => {
+        span.log({ input: fnArgs });
+        const outputResult = fn(...fnArgs);
+        const output =
+          outputResult instanceof Promise ? await outputResult : outputResult;
```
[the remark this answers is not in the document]
ankrgyl: Ah cool

js/src/logger.ts
ankrgyl: That works. I was kind of thinking you _might_ want that, but now that I think about it, promises (unlike streams) are nothing until they resolve. so we might as well combine

js/src/logger.ts
ankrgyl: Added.

PR #295, opened 2024-07-04
Setup js tests

manugoyal:
Switch from jest to vitest. Also add a github workflow to run the tests.


ankrgyl:
I was just thinking about this! Great

PR #296, opened 2024-07-05
Adding position to project scores type

danielericlee:
We are adding a `position` column to the `project_scores` table to support re-ordering of human review score in the UI. I'm not sure if I need to update types in more than one spot, but this change got it working in VSCode.


core/js/typespecs/app_types.ts
```diff
@@ -554,6 +554,7 @@ export const projectScoreSchema = z
         destination: z.literal("expected").nullish(),
       })
       .nullish(),
+    position: z.string().nullish(),
```
manugoyal, line 557: Since these are public-facing types (it'll end up in the REST API), we should add some docs about it with `.describe`.

If we really want to hide this from the user, we could also augment the `projectScore` type signature with this column internally wherever we need it, but I don't think it's super necessary to hide.

manugoyal (approved):

core/js/typespecs/app_types.ts
danielericlee: Great, will add a description via `.describe` 👍

PR #297, opened 2024-07-05
Use .strip() for views fields

aphinx:
This will make it easier to add new fields to the view schema without breaking compatibility.


ankrgyl (approved):
LGTM but @manugoyal is deep into this topic so tagging him

manugoyal (approved):
lgtm. Concurrently I'm taking a pass to make all typespecs `strip` because I think we run into this problem everywhere. So this PR is in line with that direction.

PR #298, opened 2024-07-05
Default base experiment to hill climbing experiment

aphinx:
- If `Eval` is called with `BaseExperiment(name=...)` as data, register the named experiment as the base experiment.
- Separately, I noticed that in Python, the named base experiment isn't actually used as data -- the default base experiment is used instead -- so I fixed that.

Fixes BRA-1455.


py/src/braintrust/framework.py
```diff
@@ -500,6 +500,14 @@ def Eval(
         except RuntimeError:  # 'RuntimeError: There is no current event loop...'
             loop = None
 
+        base_experiment_name = None
+
+        data_iterator = evaluator.data
+        if inspect.isclass(data_iterator):
+            data_iterator = data_iterator()
```
aphinx, line 507: is it fine to create an instance here? it appears that this code is here to handle the `data=BaseExperiment` case in particular, in which case it seems fine.

ankrgyl: Hmm I don't think so. I don't think we can add the property that this is ok to be side-effect free all of a sudden.

aphinx: OK, makes sense. In that case, we probably need to create the instance here and pass it along in an optional arg to `run_evaluator` so it only occurs once

ankrgyl: Yeah although annoying, that feels like the right solution. _or_ maybe you could memoize it. Like place it into a wrapper of some sort, so that you can call it multiple times?

aphinx: what about simply calling it _before_ creating the `Evaluator` and assigning the result to the `.data` field?

ankrgyl: Does that affect when it's run eg. in an eval file vs today?

js/src/framework.ts
```diff
@@ -260,6 +260,29 @@ function initExperiment<IsOpen extends boolean = false>(
   });
 }
 
+export async function callEvaluatorData<
```
aphinx, line 263: The idea here is that we have to call `data` (if it's a function) in order to determine whether it's a `BaseExperiment`. We do this by calling the function (and `await`ing it, if it's a Promise), prior to registering the experiment and invoking `runEvaluator`. Is this fine? With the way things are laid out, we can't really get away with _not_ calling the function here if we want to check it, but we _could_ avoid awaiting the promise if we tweak the type definitions.

Or we could just avoid changing the "order of operations" altogether by assigning the base experiment in `runEvaluator` using a separate API call (maybe add a new API endpoint for updating the base experiment?). I was trying to avoid moving stuff around too much but maybe that's cleaner than trying to manipulate `data` before running the evaluator.

py/src/braintrust/framework.py
aphinx: ^ Yeah that would run the data function during the lazy load instead of during the eval, which is no good. I changed it to just check whether it's an instance of `BaseExperiment`, but in the TS SDK things were a bit trickier.

js/src/framework.ts
aphinx: Expanded the `EvalData` type to allow us to call the `data` fn before initializing the experiment.

ankrgyl: Does this function still need to be async?

ankrgyl (approved):

PR #299, opened 2024-07-06
Small tweaks to support partial scores

ankrgyl:


core/py/src/braintrust_core/score.py
```diff
@@ -51,8 +51,12 @@ async def _run_eval_async(self, output, expected=None, **kwargs) -> Score:
         # By default we just run the sync version in a thread
         return self._run_eval_sync(output, expected, **kwargs)
 
+    @classmethod
```
manugoyal, line 54: Out of curiosity, why do we need our own method for this? Is it an expected method for some caller (if so maybe worth documenting).

manugoyal (approved):

core/js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@braintrust/core",
-  "version": "0.0.6",
+  "version": "0.0.45",
```
manugoyal, line 3: Is this intended to be kept in sync with `version.py` now? I think that file is still on version `0.0.44`.

core/py/src/braintrust_core/score.py
ankrgyl: I think I can revert this since I reworked how class names are defined

core/js/package.json
ankrgyl: Nope. This just allows autoevals to utilize some of the new features in core. But I might be able to revert this too.

ankrgyl:
We ended up not needing these changes!

PR #300, opened 2024-07-06
Replace all usages of `strictObject` with `object`.

manugoyal:
We've decided that by default, our zod object schemas should be in `strip` mode rather than `strict` mode, because there are many usage sites which only want to use the schema for validation, which may not be always up-to-date with the latest schema definition (e.g. the API server, sdk libraries). So `strip` is safer for validation.

For users who do want to check no extra fields, we introduce a `parseNoStrip` utility function which parses the object and checks that no fields were removed from the final object.

### DIFF
--- a/core/js/typespecs/common_types.ts
+++ b/core/js/typespecs/common_types.ts
@@ -15,6 +15,15 @@ export const jsonSchema: z.ZodType<Json> = z.lazy(() =>
   z.union([literalSchema, z.array(jsonSchema), z.record(jsonSchema)]),
 );
 
+// It is often hard for us to control every piece of code that serializes
+// datetimes to strings to ensure they are always strictly ISO8601-compliant.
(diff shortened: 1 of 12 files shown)


ankrgyl (approved):

core/js/typespecs/common_types.ts
```diff
@@ -13,7 +15,21 @@ export const jsonSchema: z.ZodType<Json> = z.lazy(() =>
   z.union([literalSchema, z.array(jsonSchema), z.record(jsonSchema)]),
 );
 
-export const datetimeStringSchema = z.string().datetime({ offset: true });
+export const datetimeStringSchema = z
```
manugoyal, line 18: This solves https://linear.app/braintrustdata/issue/BRA-1399/different-timestamps-returned-by-create-delete-patch-and-list-group by having zod transform the string to a datetime.

PR #304, opened 2024-07-09
Update patch dataset schema

tara-nagar:
Adding metadata to patch dataset schema


ankrgyl (approved):
LGTM but @manugoyal should sanity check and let's make sure to land alognside the corresponding openapi changes in the main repo

PR #307, opened 2024-07-10
Support propagating "state" to traced

ankrgyl:


js/src/logger.ts
```diff
@@ -2135,6 +2151,10 @@ export interface LoginOptions {
    * A custom fetch implementation to use.
    */
   fetch?: typeof globalThis.fetch;
+  /**
+   * If true, will not flush the logger when the process exits.
```
manugoyal, line 2155: Maybe "flush pending writes" since "logger" is already a term that means something else.

manugoyal (approved):

js/src/logger.ts
ankrgyl: improved

PR #308, opened 2024-07-12
Robustify signature inspection in `traced` decorator.

manugoyal:
Similar to https://github.com/braintrustdata/braintrust-sdk/pull/276, `inspect.signature` does not work for all function types.


ankrgyl (approved):
lgtm

PR #310, opened 2024-07-12
[Python] Treat tools empty string as null

aphinx:
This actually aligns with the typsecript code, which checks the truthiness of the tools string instead of whether it's null(ish).

Fixes BRA-1497.


ankrgyl (approved):

py/src/braintrust/logger.py
```diff
@@ -2635,9 +2635,7 @@ def build(self, **build_args):
                 for m in self.prompt.messages
             ]
             ret["tools"] = (
-                json.loads(chevron.render(self.prompt.tools, data=build_args))
-                if self.prompt.tools is not None
-                else None
+                json.loads(chevron.render(self.prompt.tools, data=build_args)) if self.prompt.tools else None
```
ankrgyl, line 2638: may want to `.strip()` it so that `" "` doesn't pass through the filter

py/src/braintrust/logger.py
```diff
@@ -2635,7 +2635,9 @@ def build(self, **build_args):
                 for m in self.prompt.messages
             ]
             ret["tools"] = (
-                json.loads(chevron.render(self.prompt.tools, data=build_args)) if self.prompt.tools else None
+                json.loads(chevron.render(self.prompt.tools, data=build_args))
+                if (self.prompt.tools or "").strip()
```
ankrgyl, line 2639: do we need to do something similar in JS?

ankrgyl (approved):

PR #323, opened 2024-07-27
Add basic support for errors

ankrgyl:
Add support for capturing errors and writing them to the `error` field in each span if they occur.


js/src/logger.ts
```diff
@@ -997,18 +997,32 @@ export class Logger<IsAsyncFlush extends boolean> implements Exportable {
...
+    try {
+      ret = runFinally(
+        () => {
+          if (setCurrent ?? true) {
+            return withCurrent(span, callback);
+          } else {
+            return callback(span);
+          }
+        },
+        () => span.end(),
```
manugoyal, line 1011: This seems slightly wrong because we're "end"-ing the span before we catch and log the error. Since `end` just logs an end time this should in theory be benign, but idiomatically we should be running the finally clause after the catch.

js/src/logger.ts
```diff
@@ -997,18 +997,32 @@ export class Logger<IsAsyncFlush extends boolean> implements Exportable {
...
-          return withCurrent(span, callback);
-        } else {
-          return callback(span);
-        }
-      },
-      () => span.end(),
-    );
+
+    let ret;
+    try {
```
manugoyal, line 1002: I think it would be cleaner to add a `runCatch` utility analogous to our `runFinally` util and use that. Something like
```
// In util.ts

// Given a function to execute and a catch handler, runs the function and
// executes the catch handler either synchronously if `f` is a synchronous
// function or chained if `f` returns a Promise.
export function runCatch<R>(f: () => R, catchF: (e: unknown) => R): R {
  try {
    const ret = f();
    if (ret instanceof Promise) {
      return (ret as any).catch(catchF) as R;
    } else {
      return ret;
    }
  } catch (e) {
    return catchF(e);
  }
}
```

Then here we could write:
```
    const ret = runFinally(
        () => runCatch(
          () => {
            if (setCurrent ?? true) {
              return withCurrent(span, callback);
            } else {
              return callback(span);
            }
          },
          (e) => {
            logError(span, e);
            throw e;
          }),
        () => span.end());
```

And leave the rest the same.

py/src/braintrust/logger.py
```diff
@@ -2394,13 +2395,34 @@ def __enter__(self):
...
+            if self.set_current:
+                _state.current_span.reset(self._context_token)
 
-        if self.set_current:
-            _state.current_span.reset(self._context_token)
+            self.end()
+
+
+def stringify_exception(exc_type, exc_value, tb):
+    lines = traceback.format_exception(exc_type, exc_value, tb)
```
manugoyal, line 2410: Python exposes more granular methods to format individual parts of the stack trace. You could write something like this:
```
return ''.join(traceback.format_exception_only(exc_type, exc_value) + ['\nTraceback (most recent call last):\n'] + traceback.format_tb(tb))
```

manugoyal (commented):
generally lgtm. Some implementation suggestions.

js/src/logger.ts
ankrgyl: updated (but made `runFinally` accept a catch clause, because everywhere we use it, we want to do this)

js/src/logger.ts
ankrgyl: makes sense

py/src/braintrust/logger.py
ankrgyl: nice thanks

js/src/util.ts
```diff
@@ -1,18 +1,24 @@
 export const GLOBAL_PROJECT = "Global";
 
-// Given a callback, runs the callback as a "finally" component. If
-// `callback` returns a Promise, `finallyF` is chained to the promise. Otherwise,
-// it is run afterwards synchronously.
-export function runFinally<R>(f: () => R, finallyF: () => void): R {
+// Given a function to execute, a catch handler, and a finally handler, runs the function and
+// executes the catch handler if an error occurs. If the function returns a Promise, the finally handler
+// is chained to the promise. Otherwise, it is run afterwards synchronously.
+export function runFinally<R>(
```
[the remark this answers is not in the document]
ankrgyl: renamed

PR #329, opened 2024-07-31
Support updating spans from the SDK

ankrgyl:


js/src/logger.ts
```diff
@@ -748,6 +748,66 @@ function logFeedbackImpl(
...
+ * @param event The event data to update the span with. See `Experiment.log` for a full list of valid fields.
+ * @param state (optional) Login state to use. If not provided, the global state will be used.
+ */
+export function updateSpan({
+  exported,
+  event,
+  state,
+}: {
+  exported: string;
+  event: Partial<ExperimentEvent>;
```
manugoyal, line 793: Should we forbid them from specifying an ID in this struct too?

ankrgyl: updated

js/src/logger.ts
```diff
@@ -748,6 +748,64 @@ function logFeedbackImpl(
...
+ *
+ * @param exported The output of `span.export()`.
+ * @param event The event data to update the span with. See `Experiment.log` for a full list of valid fields.
+ * @param state (optional) Login state to use. If not provided, the global state will be used.
+ */
+export function updateSpan({
+  exported,
+  state,
+  ...event
+}: { exported: string } & Omit<Partial<ExperimentEvent>, "id"> &
```
manugoyal, line 791: Should we also have `Omit<Partial<ExperimentEvent>, "id">` in the `updateSpanImpl` type signature for extra safety?

js/src/logger.ts
```diff
@@ -1005,6 +1063,26 @@ export class Logger<IsAsyncFlush extends boolean> implements Exportable {
     logFeedbackImpl(this.state, this.parentObjectType(), this.lazyId, event);
   }
 
+  /**
+   * Update a span in the experiment using its id. It is important that you only update a span once the original span has been fully written and flushed,
+   * since otherwise updates to the span may conflict with the original span.
+   *
+   * @param event The event data to update the span with. Must include `id`. See `Experiment.log` for a full list of valid fields.
```
manugoyal, line 1070: We can enforce this in the type signature too? `Omit<Partial<ExperimentEvent>, "id"> & Required<Pick<ExperimentEvent>, "id">` or something?

manugoyal (approved):
generally looks fine to me. Internally we should make sure unit tests cover passing ID in the wrong place where possible.

js/src/logger.ts
ankrgyl: yep.

PR #332, opened 2024-08-06
Log errors in `Eval()` at the top level

ankrgyl:
Mark the entire eval row as failed if the task fails. This makes it easier to filter down to failing rows.

### DIFF
--- a/core/js/typespecs/openai/messages.ts
+++ b/core/js/typespecs/openai/messages.ts
@@ -78,7 +78,7 @@ const chatCompletionAssistantMessageParamSchema = z.object({
   role: z.literal("assistant"),
   content: z.string().nullish(),
   function_call: functionCallSchema.nullish(),
-  name: z.string().optional(),
+  name: z.string().nullish(),
   tool_calls: z.array(chatCompletionMessageToolCallSchema).nullish(),
 });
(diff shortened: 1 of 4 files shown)


core/js/typespecs/openai/messages.ts
```diff
@@ -77,9 +77,9 @@ export const chatCompletionMessageToolCallSchema = z.object({
 const chatCompletionAssistantMessageParamSchema = z.object({
   role: z.literal("assistant"),
   content: z.string().nullish(),
-  function_call: functionCallSchema.optional(),
+  function_call: functionCallSchema.nullish(),
```
manugoyal, line 80: should we make the `name` nullish too?

manugoyal (approved):

core/js/typespecs/openai/messages.ts
ankrgyl: Yeah

ankrgyl: For future reference - we can't do this willy nilly since it breaks the openai wrappers

manugoyal: What's the breakage?

PR #341, opened 2024-08-15
Wrapper support for structured outputs beta

aphinx:


js/src/wrappers/oai.ts
```diff
@@ -145,7 +147,43 @@ interface NonStreamingChatResponse {
...
+    try {
+      span.log({
+        output: ret.choices,
+        metrics: {
+          time_to_first_token: getCurrentUnixTimestamp() - startTime,
+          tokens: ret.usage?.total_tokens,
+          prompt_tokens: ret.usage?.prompt_tokens,
+          completion_tokens: ret.usage?.completion_tokens,
+        },
+      });
```
ankrgyl, line 178: should we refactor this into a common function since it's shared with the normal chat completion stuff?

ankrgyl (approved):

PR #345, opened 2024-08-26
Minor tweaks to enable async scoring.

manugoyal:


core/js/typespecs/functions.ts
```diff
@@ -88,6 +97,21 @@ const invokeFunctionNonIdArgsSchema = z.object({
...
+        .string()
+        .describe("The id of the container object you are logging to"),
+      row_id: z.string().describe("Id of the row to update"),
+      [ASYNC_SCORING_TOKEN_FIELD]: z
+        .string()
+        .describe("Token that launched the update"),
+    })
+    .nullish()
+    .describe(
+      "If specified, log the result of the function invocation as a score to the specified row. In this case, the function output must be a number between 0 and 1, inclusive",
```
ankrgyl, line 113: what happens if it returns a `Score` object?

core/js/typespecs/functions.ts
```diff
@@ -88,6 +97,21 @@ const invokeFunctionNonIdArgsSchema = z.object({
     .describe(
       "Whether to stream the response. If true, results will be returned in the Braintrust SSE format.",
     ),
+  update_score: z
```
ankrgyl, line 100: this seems a little strange to include in the public API for invoke. Should we make it a hidden argument or a separate endpoint or something?

manugoyal: Agreed that this makes sense to maintain internally for now. Will move.

core/js/typespecs/functions.ts
manugoyal: Good question. See comment in internal PR about what the correct output format should be for the function output.

ankrgyl (approved):

ankrgyl (approved):

PR #352, opened 2024-09-02
Expose span.withCurrent()

ankrgyl:


manugoyal (commented):
Out of curiosity, why not just export the global version of the function? It has a wider interface too (accepts an arbitrary state)

ankrgyl:
We should probably do that. I originally dindt' because I thought we'd need to use the `state` parameter in the span, but that turned out to be wrong. I'll change it.

PR #368, opened 2024-09-16
Added `update` method to datasets

arvindanand1123:
Added `update` to merge events into database. Also refactored `insert` to share common code with `update`.


js/src/logger.ts
```diff
@@ -3765,6 +3765,73 @@ export class Dataset<
...
+    }
+
+    const rowId = id || uuidv4();
+    const args = new LazyValue(async () => ({
+      id: rowId,
+      input,
+      expected: expected === undefined ? output : expected,
+      tags,
+      dataset_id: await this.id,
+      created: new Date().toISOString(),
```
ankrgyl, line 3826: @manugoyal should we omit `created` for updates?

ankrgyl (commented):
looks pretty good to me! @manugoyal can conduct the official review :)

js/src/logger.ts
arvindanand1123: Would `updatedAt` or something like that make sense?

arvindanand1123: So I tried removing it and typescript yells at me --> hence why the build is failing. Should we modify the type to make `created` optional?

js/src/logger.ts
```diff
@@ -3765,6 +3765,69 @@ export class Dataset<
...
+   * Update fields of a single record in the dataset. The updated fields will be batched and uploaded behind the scenes.
+   * You must pass in an `id` of the record to update. Only the fields provided will be updated; other fields will remain unchanged.
+   *
+   * @param event The fields to update in the record.
+   * @param event.input (Optional) The new input value for the record (an arbitrary, JSON serializable object).
+   * @param event.expected (Optional) The new expected output value for the record (an arbitrary, JSON serializable object).
+   * @param event.tags (Optional) A list of strings to update the tags of the record.
+   * @param event.metadata (Optional) A dictionary to update the metadata of the record. The values in `metadata` can be any
+   * JSON-serializable type, but its keys must be strings.
+   * @param event.id The unique identifier of the record to update.
```
manugoyal, line 3778: Should we put this first in the docs, since it's a required argument?

js/src/logger.ts
```diff
@@ -3765,6 +3765,69 @@ export class Dataset<
...
+   * @param event.output (Deprecated) The new expected output value for the record. Use `expected` instead.
```
manugoyal, line 3779: I think we can avoid adding deprecated fields to new methods.

js/src/logger.ts
```diff
@@ -3765,6 +3765,69 @@ export class Dataset<
...
+    output,
+  }: {
+    readonly id: string;
+    readonly input?: unknown;
+    readonly expected?: unknown;
+    readonly tags?: string[];
+    readonly metadata?: Record<string, unknown>;
+    readonly output?: unknown;
+  }): string {
+    if (metadata !== undefined) {
```
manugoyal, line 3797: Let's share some of this code between insert and update. Probably both of them could call some helper function that accepts an `extraRowProps` object that is added into the `args` object.

js/src/logger.ts
```diff
@@ -3765,6 +3765,69 @@ export class Dataset<
...
+
+    const rowId = id;
+    const args = new LazyValue(async () => ({
+      id: rowId,
+      input,
+      expected: expected === undefined ? output : expected,
+      tags,
+      dataset_id: await this.id,
+      metadata,
+      created: new Date().toISOString(),
```
manugoyal, line 3823: I think we should leave out created. Even though it will be ignored by the backend, it looks weird to be passing it into update. We can make it optional in the typedef here: https://github.com/braintrustdata/braintrust-sdk/blob/708eb61ae54de0fcccffafc686218a959fa0e12e/core/js/src/object.ts#L95.

py/src/braintrust/logger.py
```diff
@@ -2569,6 +2569,57 @@ def compute_args():
...
+        :param id: The unique identifier of the record to update.
+        :param input: (Optional) The new input value for the record (an arbitrary, JSON serializable object).
+        :param expected: (Optional) The new expected output value for the record (an arbitrary, JSON serializable object).
+        :param tags: (Optional) A list of strings to update the tags of the record.
+        :param metadata: (Optional) A dictionary to update the metadata of the record. The values in `metadata` can be any
+            JSON-serializable type, but its keys must be strings.
+        :param output: (Deprecated) The new expected output value for the record. Use `expected` instead.
+        :returns: The `id` of the updated record.
+        """
+        if metadata:
```
manugoyal, line 2586: Similar comments to JS.

manugoyal (commented):
logically looks good! Just some cleanup-related comments.

js/src/logger.ts
arvindanand1123: Will remove the output param in that case

js/src/logger.ts
```diff
@@ -3732,39 +3804,64 @@ export class Dataset<
...
+   * @param event The fields to update in the record.
+   * @param event.id The unique identifier of the record to update.
+   * @param event.input (Optional) The new input value for the record (an arbitrary, JSON serializable object).
+   * @param event.expected (Optional) The new expected output value for the record (an arbitrary, JSON serializable object).
+   * @param event.tags (Optional) A list of strings to update the tags of the record.
+   * @param event.metadata (Optional) A dictionary to update the metadata of the record. The values in `metadata` can be any
+   * JSON-serializable type, but its keys must be strings.
+   * @returns The `id` of the updated record.
+   */
+  public update({
```
manugoyal, line 3837: similarly here

js/src/logger.ts
```diff
@@ -3701,6 +3701,78 @@ export class Dataset<
...
+    metadata?: Record<string, unknown>;
+    tags?: string[];
+    output?: unknown;
+    isMerge?: boolean;
+  }): LazyValue<BackgroundLogEvent> {
+    return new LazyValue(async () => {
+      const dataset_id = await this.id;
+      const expectedValue = expected === undefined ? output : expected;
+
+      let args: BackgroundLogEvent = {
```
manugoyal, line 3755: Could write as
```
const args: BackgroundLogEvent = {
    // Your existing stuff
    ...(isMerge ? { [IS_MERGE_FIELD]: true } : {})
};
```

manugoyal (approved):
lgtm!

PR #370, opened 2024-09-17
Export some stuff from JS SDK

ankrgyl:


ankrgyl:
Maybe we should just export all of it?

manugoyal (approved):
lgtm. Regarding exporting everything, I think it makes sense to export the set of argument/result types that users interact with in public functions, but I prefer not to export types that aren't part of public interfaces.

PR #371, opened 2024-09-17
`Prompt.build` should only render `tools` when non-empty.

kevin1:
The existing behavior outputs `"tools": None` when there are no tools, which is not supported by the OpenAI API. Now it will not be rendered.


ankrgyl (approved):
lgtm

PR #376, opened 2024-09-18
MetricSummary print token counts without decimals.

kevin1:


js/src/framework.ts
```diff
@@ -981,11 +981,9 @@ function formatMetricSummary(
   summary: MetricSummary,
   longestMetricName: number,
 ) {
-  const diffString = isEmpty(summary.diff)
-    ? ""
-    : ` (${summary.diff > 0 ? "+" : ""}${(summary.diff * 100).toFixed(2)}%)`;
+  const fractionDigits = summary.unit == "tok" ? 0 : 2;
```
ankrgyl, line 984: a slightly more dynamic way to do this would be to test if the number is an integer right?

kevin1: Done. I was trying to keep the logic aligned with Python SDK, but then I realized that Python can also have an int value for `MetricSummary.metric`. So that type annotation has been updated too.

ankrgyl (approved):

PR #377, opened 2024-09-19
Python `--verbose` flag maps to INFO log level.

kevin1:
Previously, the verbose flag was used for both DEBUG log level and printing full stack traces. Now it will only control printing full stack traces.

To see DEBUG logs, pass the `--verbose` flag twice.


ankrgyl (approved):
Amazing

PR #382, opened 2024-09-24
Env var API

ankrgyl:


core/js/typespecs/common_types.ts
```diff
@@ -61,6 +61,7 @@ export const objectTypes = z.enum([
   "organization",
   "api_key",
   "ai_secret",
+  "env_var",
```
manugoyal, line 64: Let's also add it to `getObjectArticle` so the docs look correct. I wonder if there's a javascript package for this...

core/js/typespecs/app_types.ts
```diff
@@ -1200,6 +1222,31 @@ export const patchAISecretSchema = z.object({
   secret: z.string().nullish(),
 });
 
+export const createEnvVarSchema = z.object({
+  object_type: envVarObjectTypeEnum,
+  object_id: z
```
manugoyal, line 1227: Rather than copying, you could write `envVarSchema.pick({ object_type: true, object_id: true, name: true }).extend({ value: ... })`

core/js/typespecs/app_types.ts
```diff
@@ -1200,6 +1222,31 @@ export const patchAISecretSchema = z.object({
...
+  value: z
+    .string()
+    .nullish()
+    .describe(
+      "The value of the environment variable. Will be encrypted at rest.",
+    ),
+});
+
+export const patchEnvVarSchema = z.object({
+  name: z.string().nullish().describe("The name of the environment variable"),
```
manugoyal, line 1241: Same here. Would prefer to pick from existing schemas instead of copy-pasting.

manugoyal (approved):

core/js/typespecs/app_types.ts
ankrgyl: fixed

core/js/typespecs/app_types.ts
ankrgyl: done

core/js/typespecs/common_types.ts
ankrgyl: done

PR #389, opened 2024-09-27
Decouple root span id from span id

aphinx:
We still set root_span_id to the root span's span_id for now, but we are gearing up for being able to decouple them without breaking stuff. The new `is_root` field for events is basically equivalent to `len(span_parents) == 0` but is added for quicker access in some cases.


core/js/typespecs/api_types.ts
```diff
@@ -372,6 +373,7 @@ export const experimentEventSchema = z
     span_parents: experimentEventBaseSchema.shape.span_parents,
     root_span_id: experimentEventBaseSchema.shape.root_span_id,
     span_attributes: experimentEventBaseSchema.shape.span_attributes,
+    is_root: experimentEventBaseSchema.shape.is_root,
```
manugoyal, line 376: We should be able to check for the presence of this in our internal REST api unit tests.

js/src/logger.ts
```diff
@@ -827,7 +827,7 @@ export function updateSpan({
 }
 
 interface ParentSpanIds {
-  spanId: string;
+  spanId?: string;
```
manugoyal, line 830: Discussed offline that we can undo these changes since they are not (yet) necessary.

manugoyal (approved):

PR #391, opened 2024-09-30
restrict creation of names to be nonempty

sachinpad:


manugoyal (approved):
lgtm. Before landing this, we should make sure it won't cause any issues in prod if there are currently objects with empty names.

core/js/typespecs/app_types.ts
```diff
@@ -865,12 +869,13 @@ export const createExperimentSchema = z
 
 export const patchExperimentSchema = createExperimentSchema
   .omit({ project_id: true, ensure_new: true })
+  .merge(z.object({ name: experimentSchema.shape.name.nullish() }))
```
manugoyal, line 872: nit: a slightly more concise way to write this (here and elsewhere):
```
.extend({ name: ... })
```

manugoyal (approved):

core/js/typespecs/app_types.ts
sachinpad: Done

PR #396, opened 2024-10-01
Add schema to track origin object.

kevin1:


core/js/typespecs/api_types.ts
```diff
@@ -170,6 +170,13 @@ function generateBaseEventOpSchema(objectType: ObjectTypeWithEvent) {
       .describe(
         "Human-identifying attributes of the span, such as name, type, etc.",
       ),
+    origin: z
+      .object({
+        object_type: z.string(),
+        object_id: z.string().uuid(),
```
ankrgyl, line 176: should we include descriptions?

ankrgyl (approved):

core/js/typespecs/api_types.ts
ankrgyl: feel free to steal these: https://github.com/braintrustdata/braintrust-sdk/blob/custom-viewers/core/js/typespecs/app_types.ts#L342

kevin1: Done

core/js/typespecs/api_types.ts
```diff
@@ -170,6 +171,19 @@ function generateBaseEventOpSchema(objectType: ObjectTypeWithEvent) {
...
+    origin: z
+      .object({
+        object_type: eventObjectType.describe(
+          "Type of the object the event is originating from.",
+        ),
+        object_id: z
+          .string()
+          .uuid()
+          .describe("ID of the object the event is originating from."),
+        id: z.string().describe("ID of the original event."),
```
manugoyal, line 183: I wonder if for posterity we should include an `_xact_id`, in case the row changes or is deleted. We don't have to use it in the UI right now, but wondering if it's worth including.

manugoyal (approved):

core/js/typespecs/api_types.ts
kevin1: Done

ankrgyl (approved):

PR #404, opened 2024-10-09
Add support for uploading file & blob attachments in JS.

kevin1:
We introduce a new class, `Attachment`, which represents the upload parameters for a file on disk or in-memory Blob/Buffer and can be constructed and JSON serialized synchronously. The caller can place `Attachment` objects anywhere in the event to be logged.

```ts
const myFile: Uint8Array = ...;
currentSpan().log({
  input: {
    question: "What is this?",
    context: {
      attachment: new Attachment({
        data: myFile,
        filename: "bridge.jpg",
        contentType: "image/jpeg",
      }),
    },
  },
});
```

Prior to event upload, the SDK serializes `Attachment` objects which replaces them with a reference to the object store. The above call is equivalent to:

```ts
currentSpan().log({
  input: {
    question: "What is this?",
    context: {
      attachment: {
        type: "braintrust_attachment",
        filename: "bridge.jpg",
        content_type: "image/jpeg",
        key: "1935ec5b-51fd-4267-a2af-db38ee37f46e",
      },
    },
  },
});
```

This serialization is synchronous, so the presence of attachments doesn't affect upload performance.

Users who desire more control can upload the file ahead of time by retaining a reference to the `Attachment` object.


js/src/logger.ts
```diff
@@ -675,6 +676,114 @@ export interface LogOptions<IsAsyncFlush> {
 
 export type PromiseUnless<B, R> = B extends true ? R : Promise<Awaited<R>>;
 
+/**
+ * A wrapper object that allows lazily uploading attachments and their metadata. This object can be logged in place of an attachment that is being uploaded asynchronously.
+ */
+export class LazyAttachmentUpload {
```
ankrgyl, line 682: I think for the user facing class, we should just call it `Attachment`. All of our classes are lazy, so no need to specify that explicitly.

ankrgyl:
```
      attachment: {
        filename: "bridge.jpg",
        content_type: "image/jpeg",
        path: "attachments/19/1935ec5b-51fd-4267-a2af-db38ee37f46e",
      },
```

We might want to put something in here, like `type: "braintrust_attachment"` (or something clean/clear) so we can check this in our UI in an unambiguous way.

js/src/logger.ts
```diff
@@ -675,6 +676,114 @@ export interface LogOptions<IsAsyncFlush> {
...
+    await this.uploadFile();
+    return metadata;
+  }
+
+  /**
+   * Request a signed url from data plane without uploading the file.
+   */
+  async uploadMetadata(): Promise<AttachmentReference> {
+    if (!this.metadataPromise) {
+      this.metadataPromise = (async () => {
```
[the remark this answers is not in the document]
kevin1: Done:

- Use `LazyValue` to wrap the upload promise
- Generate the metadata (AttachmentReference) on client side, avoiding the need for an additional `LazyValue<AttachmentReference>`

js/src/logger.ts
```diff
@@ -675,6 +677,108 @@ export interface LogOptions<IsAsyncFlush> {
...
+        ret.objectStoreResponse = await fetch(ret.signedUrl, {
+          method: "PUT",
+          headers: {
+            "Content-Type": this.reference.content_type,
+          },
+          body: data,
+        });
+
+        if (ret.objectStoreResponse.ok) {
+          this.reference.upload_status = "done";
```
manugoyal, line 757: Should we catch any errors and set the upload status to `error`? I suppose ideally we could log the error message too.

js/src/logger.ts
```diff
@@ -1143,6 +1252,15 @@ export class Logger<IsAsyncFlush extends boolean> implements Exportable {
     });
   }
 
+  // TODO: do we want convenience methods?
```
manugoyal, line 1255: Personally feels unnecessary but not a strong opinion.

js/src/logger.ts
```diff
@@ -1367,6 +1491,19 @@ class BackgroundLogger {
       }
     }
 
+    const attachments = this.attachments;
+    this.attachments = [];
+    // TODO: Error handling.
+    const uploadResponses = await Promise.all(
```
manugoyal, line 1497: Would it make sense to do this concurrently with logging, rather than after?

manugoyal: Also note that `Promise.all` can abort the entire set of promises if any of them fails. Maybe in this case we want to use `Promise.allSettled` to try resolving all of them?

js/src/logger.ts
```diff
@@ -1367,6 +1491,19 @@ class BackgroundLogger {
...
+      attachments.map((attachment) => attachment.result.get()),
+    );
+    uploadResponses.forEach((result) => {
+      console.log(
```
manugoyal, line 1501: Probably want to remove this before merging?

js/src/logger.ts
```diff
@@ -784,13 +890,16 @@ function updateSpanImpl({
...
+  const attachments = extractAttachments(updateEvent);
+
   const record = new LazyValue(async () => ({
     id,
     ...updateEvent,
     ...(await parentIds()),
     [IS_MERGE_FIELD]: true,
   }));
   state.bgLogger().log([record]);
+  state.bgLogger().logAttachments(attachments);
```
manugoyal, line 902: Should we add the attachments parameter to `log` as a way to enforce that people don't forget about it? The logger is an internal-only API so it shouldn't be a problem to change the signature.

js/src/logger.ts
```diff
@@ -2863,6 +3000,39 @@ function validateAndSanitizeExperimentLogPartialArgs(
   }
 }
 
+/**
+ * Helper function for uploading attachments. Recursively extracts `Attachment` values.
+ * @param event The event to be extract. Will be modified.
+ * @returns The function to get the modified event and the extracted `Attachment` objects.
+ */
+function extractAttachments<T extends Partial<BackgroundLogEvent>>(
```
manugoyal, line 3008: I wonder if the extra type safety is actually important here. Since we're mutating the object in place, could we just make it `event: Record<string, unknown>`?

js/src/logger.ts
```diff
@@ -2863,6 +3000,39 @@ function validateAndSanitizeExperimentLogPartialArgs(
...
+  event: T,
+): Attachment[] {
+  const attachments: Attachment[] = [];
+
+  function helper(value: any) {
+    if (typeof value !== "object") {
```
manugoyal, line 3014: Probably want to check `value instanceof Object` here, since `typeof null === "object"`

manugoyal (commented):
generally looks good! Just some implementation comments.

js/src/logger.ts
kevin1: Removed

js/src/logger.ts
kevin1: Done

js/src/logger.ts
kevin1: I am closing this comment because we no longer mutate the object in place.

js/src/logger.ts
```diff
@@ -675,6 +677,108 @@ export interface LogOptions<IsAsyncFlush> {
...
+  ) {
+    if (!iso.randomUUID) {
+      throw new Error("This platform does not support randomUUID");
+    }
+    const uuid = iso.randomUUID();
+    this.reference = {
+      type: BRAINTRUST_ATTACHMENT,
+      filename,
+      content_type: contentType,
+      key: `attachments/${uuid.slice(0, 2)}/${uuid}`,
```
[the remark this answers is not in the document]
kevin1: Yes, the shard is computed by hashing the key up to the final `/`. I think at our current scale, two characters (up to 256 shards) should be enough to absorb the writes.

js/src/logger.ts
```diff
@@ -675,6 +677,108 @@ export interface LogOptions<IsAsyncFlush> {
...
+   * @param data A string representing the path of the file on disk, or a `Blob`/`ArrayBuffer` with the file's contents. The caller is responsible for ensuring the file/blob/buffer is not modified until upload is complete.
+   * @param filename The desired name of the file in Braintrust after uploading. This parameter is for visualization purposes only and has no effect on attachment storage.
+   * @param contentType The MIME type of the file.
+   */
+  constructor(
+    data: string | Blob | ArrayBuffer,
+    filename: string,
+    contentType: string,
+  ) {
+    if (!iso.randomUUID) {
```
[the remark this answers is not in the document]
kevin1: Oops! Changed to uuidv4.

js/src/logger.ts
kevin1: Removed

js/src/logger.ts
kevin1: Done

js/src/logger.ts
```diff
@@ -675,6 +678,107 @@ export interface LogOptions<IsAsyncFlush> {
...
+            "Content-Type": this.contentType,
+          },
+          body: data,
+        });
+
+        if (ret.objectStoreResponse.ok) {
+          this.reference.upload_status = "done";
+        }
+      } finally {
+        return ret;
```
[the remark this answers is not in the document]
kevin1: Removed `finally`

core/js/typespecs/attachment.ts
```diff
@@ -0,0 +1,13 @@
+import { z } from "zod";
+
+export const BRAINTRUST_ATTACHMENT = "braintrust_attachment";
+
+export const attachmentReferenceSchema = z.object({
+  type: z.literal(BRAINTRUST_ATTACHMENT),
+  filename: z.string().min(1),
+  content_type: z.string().min(1),
```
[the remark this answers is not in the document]
kevin1: Added a description and moved out of typespecs

js/src/logger.ts
```diff
@@ -736,7 +914,16 @@ function logFeedbackImpl(
         [IS_MERGE_FIELD]: true,
       };
     });
+    const recordAttachmentUpdate = new LazyValue(async () => {
+      return {
+        id,
+        ...JSON.parse(JSON.stringify(filteredUpdateEvent)),
```
manugoyal, line 920: I wonder if we should roundtrip the `updateEvent` while we're here, so we don't forget about it.

kevin1: Done, deep copying before all log callsites

js/src/logger.ts
kevin1: Done

js/src/logger.ts
kevin1: Done

js/src/logger.ts
```diff
@@ -675,6 +677,108 @@ export interface LogOptions<IsAsyncFlush> {
...
+  toJSON(): AttachmentReference {
+    return this.reference;
+  }
+
+  private initUploader(): LazyValue<AttachmentUploadResult> {
+    return new LazyValue(async () => {
+      const ret: AttachmentUploadResult = {};
+
+      try {
+        await _globalState.login({});
```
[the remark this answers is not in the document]
kevin1: Done

core/js/src/attachment-reference.ts
```diff
@@ -0,0 +1,31 @@
+import { z } from "zod";
```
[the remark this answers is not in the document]
kevin1: Done, moved to `core/js/typespecs`

js/src/logger.ts
```diff
@@ -680,6 +682,185 @@ export interface LogOptions<IsAsyncFlush> {
...
+
+      return { signedUrl, metadataResponse, objectStoreResponse };
+    };
+
+    // Catches error messages and updates the attachment status.
+    const errorWrapper = async () => {
+      const status: AttachmentStatus = { upload_status: "done" };
+
+      if (!this.state) {
+        await _globalState.login({});
```
manugoyal, line 814: hmm should we write it like
```
const state = this.state ?? _globalState;
await state.login({});
```

So that we log in even when they are using a separate state? It may not matter in practice if non-global states happen to always be logged in, but seems best to write it this way anyways.

js/src/logger.ts
```diff
@@ -680,6 +682,185 @@ export interface LogOptions<IsAsyncFlush> {
...
+      const conn = this.state.apiConn();
+      const orgId = this.state.orgId ?? "";
+
+      try {
+        await doUpload(conn, orgId);
+      } catch (error) {
+        status.upload_status = "error";
+        status.error_message =
+          error instanceof Error ? error.message : JSON.stringify(error);
+      } finally {
```
manugoyal, line 826: nit: I don't think we need to put this in a `finally` block if `catch` is capturing the exception and not throwing. Since control flow is guaranteed to get past the `try/catch` block.

js/src/logger.ts
```diff
@@ -2924,6 +3131,78 @@ function validateAndSanitizeExperimentLogPartialArgs(
   }
 }
 
+/**
+ * Creates a deep copy of the given event. Replaces references to user objects
+ * with placeholder strings to ensure serializability, except for
+ * {@link Attachment} objects, which are preserved.
+ */
+function deepCopyEvent<T extends Partial<BackgroundLogEvent>>(event: T): T {
```
manugoyal, line 3139: Could we add a simple unit test for this? Should just need to be in `logger.test.ts` or something.

js/src/logger.ts
```diff
@@ -2924,6 +3131,78 @@ function validateAndSanitizeExperimentLogPartialArgs(
...
+    } else if (v instanceof Logger) {
+      return `<logger>`;
+    } else if (v instanceof Attachment) {
+      const idx = attachments.push(v);
+      return { [IDENTIFIER]: idx - 1 };
+    }
+    return v;
+  });
+  const x = JSON.parse(serialized, (_k, v) => {
+    if (
```
manugoyal, line 3166: If you are feeling adventurous, could also write this with zod
```
const serializedAttachmentSchema = z.strictObject({ [IDENTIFIER]: z.number() })
...
const parsedAttachment = serializedAttachmentSchema.safeParse(v);
if (parsedAttachment.success) { return attachments[parsedAttachment.data[IDENTIFIER]]; }
else { return v; }
```

js/src/logger.ts
```diff
@@ -2924,6 +3131,78 @@ function validateAndSanitizeExperimentLogPartialArgs(
...
+}
+
+/**
+ * Helper function for uploading attachments. Recursively extracts `Attachment`
+ * values and replaces them with their `AttachmentReferences`.
+ *
+ * @param value The event to filter. Will be modified in-place.
+ * @param attachments Flat array of extracted attachments.
+ */
+function extractAttachments(value: any, attachments: Attachment[]): void {
```
manugoyal, line 3185: The type of `value` could be `Record<string, unknown>`?

js/src/logger.ts
```diff
@@ -2924,6 +3131,78 @@ function validateAndSanitizeExperimentLogPartialArgs(
...
+  for (const key of Object.keys(value)) {
```
manugoyal, line 3186: Could also iterate over `Object.entries` and avoid a separate lookup for the value.

js/src/logger.ts
```diff
@@ -2924,6 +3131,78 @@ function validateAndSanitizeExperimentLogPartialArgs(
   }
 }
 
+/**
+ * Creates a deep copy of the given event. Replaces references to user objects
+ * with placeholder strings to ensure serializability, except for
+ * {@link Attachment} objects, which are preserved.
+ */
+function deepCopyEvent<T extends Partial<BackgroundLogEvent>>(event: T): T {
+  // We both check for serializability and round-trip `partialRecord` through
```
manugoyal, line 3140: nit: replace `partialRecord` with `event`.

js/src/logger.ts
```diff
@@ -1423,6 +1606,25 @@ class BackgroundLogger {
       }
     }
 
+    const attachmentErrors: unknown[] = [];
+    // For now, upload attachments serially.
```
manugoyal, line 1610: Would it be about as easy to do this with `Promise.allSettled`?

js/src/logger.ts
```diff
@@ -1431,11 +1633,15 @@ class BackgroundLogger {
...
     for (let i = 0; i < this.numTries; ++i) {
       try {
-        const itemPromises = wrappedItems.map((x) => x.get());
-        return mergeRowBatch(await Promise.all(itemPromises));
+        const items = await Promise.all(wrappedItems.map((x) => x.get()));
+
+        const attachments: Attachment[] = [];
+        items.forEach((item) => extractAttachments(item, attachments));
+
+        return [mergeRowBatch(items), attachments];
```
manugoyal, line 1644: I wonder if we should extract *after* doing the merging. Since in theory someone could merge-overwrite one attachment with another and then we'd be unnecessarily uploading something.

js/src/logger.ts
```diff
@@ -1561,7 +1767,8 @@ class BackgroundLogger {
       return;
     }
     try {
-      const allItems = await this.unwrapLazyValues(wrappedItems);
+      // TODO: come up with a representation of attachments.
```
manugoyal, line 1770: Could we dump the attachment references?

manugoyal (approved):
looks great! Mostly some minor nits.

js/src/logger.ts
kevin1: Done

js/src/logger.ts
manugoyal: Good point. Fine with the comment you added.

js/src/logger.ts
```diff
@@ -680,6 +682,205 @@ export interface LogOptions<IsAsyncFlush> {
...
+        throw error;
+      }
+
+      // TODO multipart upload.
+      let objectStoreResponse: Response | undefined;
+      try {
+        objectStoreResponse = await checkResponse(
+          await fetch(signedUrl, {
+            method: "PUT",
+            headers: {
```
manugoyal, line 810: Out of curiosity, do we actually need these headers if they are included when creating the signed url?

js/src/logger.ts
```diff
@@ -1561,16 +1792,24 @@ class BackgroundLogger {
...
+      const promises: Promise<void>[] = [];
       for (const payloadDir of publishPayloadsDir) {
-        await BackgroundLogger.writePayloadToDir({
-          payloadDir,
-          payload: dataStr,
-        });
+        const write = (payload: string) =>
+          BackgroundLogger.writePayloadToDir({ payloadDir, payload });
+        promises.push(write(dataStr));
+        promises.push(write(attachmentStr));
```
manugoyal, line 1809: I suspect this will end up with a lot of empty `[]` entries in the debug directory. Could we either bundle the `dataStr` and `attachmentStr` into one object with keys `data` and `attachments`, or omit the attachments when the list is empty?

js/src/logger.ts
```diff
@@ -1561,16 +1792,24 @@ class BackgroundLogger {
...
       }
+
+      await Promise.allSettled(promises);
```
manugoyal, line 1812: I think `Promise.all` is fine here since this way won't propagate any errors from running the promises. Writing the debug info is a best-effort thing anyhow. We could also just serially await the promises since this whole thing happens in the background.

manugoyal (approved):

js/src/logger.ts
kevin1: Yeah it's not the most ergonomic. What you are referring to in s3 terminology is "hoistable headers." It seems to work for GET only. 

For POST, we provide the headers once to compute the signature. Then the client needs to PUT the same headers to pass the signature check. I made this more ergonomic by having our endpoint return the exact headers that the client needs to pass to S3 during upload. This will help decouple the API and SDK.

js/src/logger.ts
kevin1: Done, restored serial code

js/src/logger.ts
kevin1: > one object with keys `data` and `attachments`

Done

js/src/logger.ts
kevin1: I agree that this will allow us to make the signature use `Record<string, unknown>`. I'm not sure what we gain though other than making the type checker happy. The input is already round-trip serialized anyway. We can guard that `Object.entries` will work by checking `value instanceof Object`. In case of `Date` or `Function`, `Object.entries` returns empty array so the recursive call is a no-op. In case of `Array`, Object.entries iterates the array so we are also covered.

On the other hand, we need to pay the cost of zod's deep copy at each level of recursion for an event which is potentially large.

kevin1: > We can guard that `Object.entries` will work by checking `value instanceof Object`. In case of `Date` or `Function`, `Object.entries` returns empty array so the recursive call is a no-op. In case of `Array`, Object.entries iterates the array so we are also covered.

Unit test will cover these cases now: 41280add25df748c1c0c1be4e7db5dec9d0cdb87

PR #411, opened 2024-10-13
Parent header

aphinx:


core/js/src/http_headers.ts
```diff
@@ -7,3 +7,6 @@ export const BT_CURSOR_HEADER = "x-bt-cursor";
 
 // User impersonation header.
 export const BT_IMPERSONATE_USER = "x-bt-impersonate-user";
+
+// Project ID header for OTEL exporters to specify a destination Braintrust project.
+export const BT_PROJECT_ID = "x-bt-project-id";
```
ankrgyl, line 12: Should we make it a parent so it can be more flexible (e.g. an experiment id, or another span)?

aphinx: That seems good -- maybe two separate headers "x-bt-parent-type" and "x-bt-parent-id", defaulting to "project" if the type header is omitted?

(And this is just to future-proof the header itself so we don't have to add more of these later, right)

aphinx: discussed offline. changed to `x-bt-parent` string (from span components `to_str`)

manugoyal (approved):

PR #416, opened 2024-10-18
Propagate git metadata params into Eval framework.

manugoyal:


ankrgyl (approved):

js/src/framework.ts
```diff
@@ -194,6 +200,16 @@ export interface Evaluator<
    * and compared to this experiment. This takes precedence over `baseExperimentName` if specified.
    */
   baseExperimentId?: string;
+
+  /**
+   * Optional settings for collecting git metadata. By default, will collect all git metadata fields allowed in org-level settings.
+   */
+  gitMetadataSettings?: GitMetadataSettings;
```
ankrgyl, line 207: should we propagate them to the eval api as well?

manugoyal: Good idea. Done.

PR #418, opened 2024-10-18
Add types for custom columns - BRA-1483

tara-nagar:


core/js/typespecs/app_types.ts
```diff
@@ -169,6 +169,27 @@ export const envVarSchema = z
...
+export const customColumnSchema = z
+  .object({
+    id: customColumnBaseSchema.shape.id,
+    object_type: customColumnObjectTypeEnum,
+    object_id: z
+      .string()
+      .uuid()
+      .describe("The id of the object the custom column is scoped for"),
+    name: z.string().describe("The name of the custom column"),
+    value: z.string().describe("The name of the custom column"),
```
j13huang, line 186: > value of the custom column

Should this be path or something?

tara-nagar: yeah good point, I didn't want to necessarily restrict it to paths but can rename to expr

ornellaaltunyan (approved):

PR #421, opened 2024-10-25
Mark the `inputs` field as deprecated

ankrgyl:


manugoyal (approved):

js/src/logger.ts
```diff
@@ -3430,7 +3430,7 @@ export class Experiment
    * @param event.metrics: (Optional) a dictionary of metrics to log. The following keys are populated automatically: "start", "end".
    * @param event.id: (Optional) a unique identifier for the event. If you don't provide one, BrainTrust will generate one for you.
    * @param event.dataset_record_id: (Optional) the id of the dataset record that this event is associated with. This field is required if and only if the experiment is associated with a dataset.
-   * @param event.inputs: (Deprecated) the same as `input` (will be removed in a future version).
+   * @deprecated @param event.inputs: (Deprecated) the same as `input` (will be removed in a future version).
```
kevin1, line 3433: `@deprecated` is a block tag: https://jsdoc.app/#block-tags I think because it's meant to apply to the entire jsdoc comment.
This might look weird when rendered into Markdown

ankrgyl: Why don't we just remove it... I think it's been long enough marked as deprecated

PR #422, opened 2024-10-29
Remove rather than deprecate the `inputs` argument

ankrgyl:


kevin1 (approved):
Should we also remove from the schema https://github.com/braintrustdata/braintrust-sdk/blob/96f99cdd47de4d5e4180c99bb7015fa77d329687/core/js/src/object.ts#L15

manugoyal (approved):
+1 to kevin's suggestion

ankrgyl:
Done!

PR #426, opened 2024-10-31
Associate init_dataset to experiments correctly

ankrgyl:


js/src/framework.ts
```diff
@@ -699,7 +703,12 @@ async function runEvaluatorInternal(
           );
           rootSpan.log({ output, metadata });
 
-          const scoringArgs = { ...datum, metadata, output };
+          const scoringArgs = {
```
ankrgyl, line 706: this is a somewhat unrelated change, but it prevents polluting the scorer args with things like `_xact_id`, `id`, etc.

core/js/src/object.ts
```diff
@@ -21,6 +25,7 @@ export type OtherExperimentLogFields = {
   metadata: Record<string, unknown>;
   metrics: Record<string, unknown>;
   datasetRecordId: string;
+  origin?: z.infer<typeof objectReferenceSchema>;
```
manugoyal, line 28: Pretty sure it doesn't need to be marked optional, similar to all the other fields

manugoyal (approved):

core/js/src/object.ts
ankrgyl: good catch

PR #434, opened 2024-11-07
Improve the type safety of Python `Eval`.

kevin1:
Make the following pattern pass type checking:

```py
dataset = init_dataset(…)
Eval(data=dataset, ...)
```


py/src/braintrust/logger.py
```diff
@@ -2523,7 +2531,31 @@ def make_final_value_callback(v):
     return serializable_partial_record, lazy_partial_record
 
 
-class Dataset(ObjectFetcher):
+class Origin(TypedDict):
```
manugoyal, line 2534: Should we add comments linking these to the typespecs they are mirroring?

manugoyal (approved):

py/src/braintrust/logger.py
kevin1: I'm going to generate the typespecs automatically. Moved that to a separate PR. This PR will be focused on fixing the type annotation of Eval which will be easier to test.

manugoyal (approved):
lgtm. Since we're changing the return type of `Eval`, maybe include the "I tested this out in a notebook" thing in the PR description and run it by ankur in case we're missing anything.

manugoyal (commented):
It's a bit of a tricky change, which I feel is likely to break existing code one way or another. If we don't think the breakage / remediation is too burdensome, this is probably the way to go.

Another option could be to have separate `EvalSync/EvalAsync` functions with clearer types and say "if you care about types, use these functions"

py/src/braintrust/framework.py
```diff
@@ -546,44 +551,133 @@ def Eval(
...
+            git_metadata_settings=git_metadata_settings,
+            repo_info=repo_info,
+        )
+
+    # https://stackoverflow.com/questions/55409641/asyncio-run-cannot-be-called-from-a-running-event-loop-when-using-jupyter-no
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:  # 'RuntimeError: There is no current event loop...'
+        loop = None
+    if loop:
```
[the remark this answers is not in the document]
kevin1: How about I leave the `Eval` behavior unchanged with a warning if you call it in an async context? And the warning will tell you to use `EvalAsync`.

py/src/braintrust/framework.py
```diff
@@ -546,44 +551,133 @@ def Eval(
...
+        # Notebook or existing async context.
```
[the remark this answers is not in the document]
kevin1: Thread is removed in the latest version

py/src/braintrust/framework.py
```diff
@@ -445,31 +445,134 @@ def _make_eval_name(name: str, experiment_name: Optional[str]):
...
+    data: EvalData,
     task: Callable[[Input, EvalHooks], Union[Output, Awaitable[Output]]],
     scores: List[EvalScorer],
     experiment_name: Optional[str] = None,
     trial_count: int = 1,
     metadata: Optional[Metadata] = None,
     is_public: bool = False,
     update: bool = False,
-    reporter: Optional[Union[ReporterDef, str]] = None,
+    reporter: Optional[ReporterDef] = None,
```
manugoyal, line 458: Can we avoid the argument defaults if everything is being delegated to this method? Just thinking of ways to reduce the amount of argument copy-paste. Seems like there's a `TypedDict` and an `Unpack` thing.

manugoyal (approved):
lgtm!

py/src/braintrust/framework.py
kevin1: Unfortunately `TypedDict` + `Unpack` breaks the doc generator, so we can't use it on public API:

Screenshot 2024-11-12 at 11 27 49 AM

This leaves the internal `_EvalCommon`. We can move the argument to `TypedDict`, but this is not reducing repetition, while making the usage less convenient (need to type `kwargs["arg_name"]`).

I will settle for removing the argument defaults.

PR #435, opened 2024-11-08
Add type annotations to Python SDK public interfaces.

kevin1:


core/py/src/braintrust_core/_types.py
```diff
@@ -0,0 +1,225 @@
+"""Do not import this file directly. See __init__.py for the classes that are considered stable.
```
manugoyal, line 1: It's kind-of messy right now, but can we actually keep this stuff outside of the python core library. We need to overhaul our dependency structure, but basically we'd like `core` (or some better name) to be only the shared dependencies between the `braintrust` package and the `autoevals` package. These shared deps should be very seldom updated. We will probably want to do the same thing for javascript, but haven't gotten around to it yet.

So basically anything that isn't required by both `autoevals` and `braintrust` should go in the `braintrust` package.

manugoyal (approved):
lgtm. Let's just move the typedefs into the main `braintrust` package.

core/py/src/braintrust_core/_types.py
kevin1: Done

PR #438, opened 2024-11-10
ensure choice_scores nonempty in prompt data schema

sachinpad:


core/js/typespecs/prompt.ts
```diff
@@ -186,16 +186,23 @@ export const promptOptionsSchema = z.object({
...
   .object({
     prompt: promptBlockDataSchema.nullish(),
     options: promptOptionsSchema.nullish(),
     // This should be a union once we support multiple parser types
-    parser: z
-      .object({
-        type: z.literal("llm_classifier"),
-        use_cot: z.boolean(),
-        choice_scores: z.record(z.number().min(0).max(1)),
+    parser: promptParserSchema
```
ankrgyl, line 200: why not put this above in `promptParserSchema`?

ankrgyl: Sorry to clarify, why not put the validation logic in `promptParserSchema`?

sachinpad: It's so that promptParserSchema can be exported and used elsewhere, like in the UI, to admit parsers without the validation that we do on the server. We could put the validation in promptParserSchema and `extend` it in the 2 places we use it in the UI--I don't have a preference between the two approaches

sachinpad: Resolved offline

ankrgyl (approved):

PR #439, opened 2024-11-11
Add Span.permalink method for conveniently generating links.

kevin1:


js/src/logger.ts
```diff
@@ -161,6 +161,22 @@ export interface Span extends Exportable {
    */
   end(args?: EndSpanArgs): number;
 
+  /**
+   * Export an identifier for this span which can be used to start a subspan in
+   * another place, such as another process or service.
```
ankrgyl, line 166: Maybe add a sentence about how this is different than an `id` (like allude to it having extra info?). Ppl often get (rightfully) confused about this

ankrgyl (approved):

js/src/logger.ts
kevin1: done

PR #446, opened 2024-11-14
Fix Zod nullish in Python typespec generation.

kevin1:
Preivously, we generated `NotRequired[Foo]` which does not allow `None` values. Now we will generate `NotRequired[Optional[Foo]]`.


core/js/typespecs/api_types.ts
```diff
@@ -37,14 +37,17 @@ import { objectNullish } from "../src/zod_util";
 
 const auditSourcesSchema = z.enum(VALID_SOURCES);
 
+export const spanTypeSchema = z
+  .enum(spanTypeAttributeValues)
```
manugoyal, line 41: the zod docs say to use `nativeEnum` for typescript enums (https://zod.dev/?id=native-enums). Haven't looked into how it's different from plain old `enum`

manugoyal (approved):

core/js/typespecs/api_types.ts
kevin1: `spanTypeAttributeValues` is a `string[]`. I think they also recommend zod enums over native enums for schema definitions.

PR #448, opened 2024-11-15
Upload attachments with the Python SDK.

kevin1:
See description of #404.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -2,6 +2,7 @@ import atexit
 import concurrent.futures
 import contextlib
 import contextvars
+import copy
 import dataclasses
 import datetime
 import inspect
@@ -29,6 +30,7 @@ from typing import (
     Mapping,
     Optional,
     Sequence,
+    Set,
     Tuple,
     TypedDict,
     TypeVar,
@@ -48,7 +50,14 @@ from urllib3.util.retry import Retry
 
 from braintrust.functions.stream import BraintrustStream
 
-from ._types import DatasetEvent, ExperimentEvent, PromptOptions, SpanAttributes
+from ._types import (
+    AttachmentReference,
+    AttachmentStatus,
+    DatasetEvent,
+    ExperimentEvent,
+    PromptOptions,
+    SpanAttributes,
+)
 from .bt_json import bt_dumps
 from .db_fields import (
     ASYNC_SCORING_CONTROL_FIELD,
@@ -343,7 +352,7 @@ class BraintrustState:
 _state: BraintrustState = None  # type: ignore
 
 
-_http_adapter = None
+_http_adapter: Optional[HTTPAdapter] = None
 
 
 def set_http_adapter(adapter: HTTPAdapter) -> None:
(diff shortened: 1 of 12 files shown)


py/src/braintrust/logger.py
```diff
@@ -1649,6 +1722,46 @@ def _validate_and_sanitize_experiment_log_full_args(event: Mapping[str, Any], ha
...
+            return v
+        else:
+            # No need to handle primitives explicitly because deepcopy will do
+            # it for us.
+            return copy.deepcopy(v)
+
+    ret: Dict[str, Any] = {}
+
+    for k, v in event.items():
+        k = str(k)
```
manugoyal, line 1751: nit: should we put stuff like this into this function (or just leave it out altogether)? https://github.com/braintrustdata/braintrust-sdk/blob/546c402c6c12aea4a218b63bc4420d8b4c1e37d8/py/src/braintrust/bt_json.py#L6

py/src/braintrust/logger.py
```diff
@@ -1649,6 +1722,46 @@ def _validate_and_sanitize_experiment_log_full_args(event: Mapping[str, Any], ha
...
+
+        if isinstance(v, Mapping):
+            v = _deep_copy_event(v)
+        elif isinstance(v, (List, Tuple, Set)):
```
manugoyal, line 1755: Maybe a tiny bit more pythonic to use `collections.abc.Sequence` instead of `List/Tuple`.

py/src/braintrust/logger.py
```diff
@@ -1761,6 +1875,161 @@ def version(self) -> str:
...
+    `AttachmentReference`.
+    """
+
+    def __init__(
+        self,
+        *,
+        data: Union[str, bytes, bytearray],
+        filename: str,
+        content_type: str,
+        state: Optional[BraintrustState] = None,
```
manugoyal, line 1894: I don't think we need to support user-provided state in python because we generally haven't exposed multi-state handling in python like we have in javascript. It is fine to just use the global state. In other places, we'll call `login()` beforehand to ensure `_state` is initialized: https://github.com/braintrustdata/braintrust-sdk/blob/546c402c6c12aea4a218b63bc4420d8b4c1e37d8/py/src/braintrust/logger.py#L1058.

manugoyal (commented):
cool, lgtm! Mainly just one thing about the handling of `BraintrustState` in the Attachment class.

py/src/braintrust/logger.py
```diff
@@ -652,11 +664,37 @@ def flush(self, batch_size=None):
...
+            attachment_errors: List[Exception] = []
+            for attachment in attachments:
+                try:
+                    result = attachment.upload()
+                    if result["upload_status"] == "error":
+                        raise RuntimeError(result.get("error_message"))
+                except Exception as e:
+                    attachment_errors.append(e)
+
+            if len(attachment_errors) == 1:
```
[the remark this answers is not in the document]
kevin1: Done

py/src/braintrust/logger.py
kevin1: The purpose of this line is to prevent dict keys from holding references to user objects. `json.dumps` will already coerce all dict key values to str, so we don't need to add it to `bt_json`. However, `bt_json` runs at log upload time so we cannot rely on it to remove the references.

I'll leave a comment for this.

py/src/braintrust/logger.py
kevin1: `Sequence` also includes `str`, `bytes`, `bytearray`, `array`, etc. The condition would have to be something like:

```py
isinstance(v, (Sequence, Set)) and not isinstance(v, (str, Buffer))
```

another option:
```py
isinstance(v, Iterable) and not isinstance(v, (str, Buffer, Mapping))
```

At that point, I thought it would be less error-prone to have a best-effort list of allowed types rather than figure out the comprehensive list of disallowed types.

py/src/braintrust/logger.py
kevin1: Done

manugoyal (approved):

PR #449, opened 2024-11-16
return org id from /api/organization/members

sachinpad:
so that it can be used from PATCH /v1/organization/members
to retrieve the corresponding object cache entry
and populate the audit headers.


sachinpad:
> Should we return org_name too?

Resolved offline

manugoyal (approved):

PR #459, opened 2024-11-22
Downloadable attachments when iterating datasets & experiments.

kevin1:


js/src/logger.ts
```diff
@@ -927,6 +995,29 @@ with a Blob/ArrayBuffer, or run the program on Node.js.`,
       return new LazyValue(async () => new Blob([data]));
     }
   }
+
+  private async getMetadata() {
```
manugoyal, line 999: If we are re-using this between the downloader and the status-getter, should we make it a lazy value so that it's cached? Otherwise maybe it would be slightly better to call `/attachment/status` in the status-getter?

js/src/logger.ts
```diff
@@ -745,42 +748,80 @@ export class Attachment {
...
+      key: reference.key,
+      state,
+    });
   }
 
   /**
    * On first access, (1) reads the attachment from disk if needed, (2)
    * authenticates with the data plane to request a signed URL, (3) uploads to
-   * object store, and (4) updates the attachment.
+   * object store, and (4) updates the attachment. If this object represents an
```
manugoyal, line 822: I wonder if it'd be a better DX if we more clearly separate the methods for upload-mode and download-mode. Or maybe even separate the classes since there are so few public methods to begin with.

I just feel like it'd look a little weird to see code like

```
const attachment = Attachment.fromReference(ref);
const status = attachment.upload();
```

and not think it's uploading something. Like what if we have a separate `ReadonlyAttachment` class that handles all of this new stuff (analogous to `ReadonlyExperiment`)

js/src/logger.ts
kevin1: Not caching with LazyValue is intentional here because the status could change during the lifetime of the Attachment object. For example if a dataset is being iterated concurrently with some attachment uploads. In that case, I am imagining a caller may want to retry reading the attachment later. 

I agree with calling `/attachment/status` if the `downloadUrl` is not needed, so that the backend request handler can potentially return faster.

manugoyal: Hmm interesting. I see how you are throwing an exception inside `initDownloader` to cancel the LazyValue caching in case the file is not finished uploading, so it does make sense to fetch the status independently.

PR #460, opened 2024-11-22
add tool choice to /function/invoke schema

sachinpad:


manugoyal (approved):

ankrgyl:
@manugoyal @sachinpad I pinged @sachinpad offline -- this is not the spot that we wanted to add tool choice

PR #467, opened 2024-12-02
support pushing js scorers

sachinpad:


js/src/framework2.ts
```diff
@@ -82,6 +84,90 @@ export class ToolBuilder {
...
+          model: opts.model,
+          params: opts.params,
+        },
+        parser: {
+          type: "llm_classifier",
+          use_cot: opts.useCot,
+          choice_scores: opts.choiceScores,
+        },
+      };
+      // TODO(sachin): Support tools for LLM scorers.
```
ankrgyl, line 152: LLM scorers don't support tools since the choice scores thing turns into tools. So we shouldn't let ppl write them.

ankrgyl (approved):

js/src/framework2.ts
sachinpad: Removed it. Also restricted the input type so that the compiler checks that tools are not provided.

PR #475, opened 2024-12-06
[BRA-1834] feat: Add hooks.metadata and deprecate hooks.meta

ibolmo:
## Changes
- Introduces new `hooks.metadata` (a reference to the metadata for the eval)
- Deprecates existing `hooks.meta` in favor of metadata
- Improves existing framework tests

## Why
This change provides a way for developers to read metadata and allows them a chance to mutate it without calling a function. 

## Testing
- Updated framework tests to verify new metadata functionality
- Verified backward compatibility with existing `hooks.meta` usage

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "braintrust",
-  "version": "0.0.161",
+  "version": "0.0.160",
   "description": "SDK for integrating Braintrust",
   "repository": {
     "type": "git",
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "braintrust",
-  "version": "0.0.160",
+  "version": "0.0.161",
   "description": "SDK for integrating Braintrust",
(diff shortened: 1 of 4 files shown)


js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "braintrust",
-  "version": "0.0.160",
+  "version": "0.0.161",
```
ankrgyl, line 3: we update this via `version.py` when we bump the versions themselves. no need to here

ankrgyl (commented):
Code lgtm. Can you add the equivalent in python too?

js/package.json
ibolmo: gotcha. reverted.

ankrgyl (approved):

ankrgyl:
Oh btw the failing test is b/c the type annotation is probably not supported in python3.8, which we still need to support for a wee bit longer

PR #482, opened 2024-12-12
compress audit resources header with gzip

sachinpad:
and provide a utility function in the python SDK to parse it

### DIFF
--- a/py/src/braintrust/audit.py
+++ b/py/src/braintrust/audit.py
@@ -1,0 +1,17 @@
+"""
+Utilities for working with audit headers.
+"""
+import base64
+import gzip
+import json
+from typing import List, TypedDict
+
+
+class AuditResource(TypedDict):
+    type: str
+    id: str
+    name: str
+
+
+def parse_audit_resources(marshaled: str) -> List[AuditResource]:
+    return json.loads(gzip.decompress(base64.b64decode(marshaled)))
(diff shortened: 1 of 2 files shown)


py/src/braintrust/audit.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class AuditResource(TypedDict):
+    type: str
+    id: str
+    name: str
+
+
+def parse_audit_resources(marshaled: str) -> List[AuditResource]:
+    return json.loads(gzip.decompress(base64.b64decode(marshaled)))
```
manugoyal, line 17: Does it make sense to have the "marshalled" payload be something slightly more elaborate so we can cleanly evolve it in the future? Like maybe it's a JSON array consisting of a version counter and a payload. Or if we wanted to be more fancy at the cost of a few more bytes, a metadata object and a payload.

sachinpad: Done!

py/src/braintrust/audit.py
```diff
@@ -0,0 +1,21 @@
...
+
+class AuditResource(TypedDict):
+    type: str
+    id: str
+    name: str
+
+
+def parse_audit_resources(hdr: str) -> List[AuditResource]:
+    j = json.loads(hdr)
+    if j["metadata"]["version"] == 1:
```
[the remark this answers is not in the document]
sachinpad: Done

PR #483, opened 2024-12-12
improve type safety of framework2.py

sachinpad:


ibolmo (approved):
didn't give it a test run, but looks solid

py/src/braintrust/_types.py
```diff
@@ -352,4 +352,147 @@ class AttachmentStatus(TypedDict):
...
+    role: Literal["user"]
+    name: NotRequired[Optional[str]]
+
+
+class FunctionCall1(TypedDict):
+    arguments: str
+    name: str
+
+
+class ChatCompletionMessageParam3(TypedDict):
```
ibolmo, line 432: I take it this is to avoid coupling the sdk with a particular openai version? I noticed something similar in autoevals

sachinpad: Yeah, and to provide the strong typing from the zod types in JS: you can see the generated type list https://github.com/braintrustdata/braintrust/pull/3776/files#diff-14c2b65c214476c421b4e1ecaaacd5ee481b5918ce3f9352db120f02be03ac3bR59

PR #497, opened 2024-12-22
use local cache for loadPrompt

sachinpad:
return cached prompt if server is unavailable


js/src/logger.ts
```diff
@@ -2697,23 +2699,33 @@ export async function loadPrompt({
...
+      forceLogin,
+    });
+    response = await state.apiConn().get_json("v1/prompt", {
+      project_name: projectName,
+      project_id: projectId,
+      slug,
+      version,
+    });
+  } catch (e) {
+    console.error("Failed to load prompt:", e);
```
ankrgyl, line 2718: maybe make this a warning? And explain that you'll try to fall back to the locally cached version?

Also, should we cache them on disk? We used to do this for some stuff, but a fairly common pattern is to create something like `~/.cache/braintrust` (overridable with an env variable, preferrably named like `BRAINTRUST_QUEUE_DROP_EXCEEDING_MAXSIZE`, etc) and save stuff there.

ankrgyl (commented):
Also can you implement the same in python?

js/src/logger.ts
sachinpad: Done!

sachinpad:
> Also can you implement the same in python?

Resolved offline

js/src/logger.ts
```diff
@@ -305,6 +306,15 @@ export class BraintrustState {
   private _apiConn: HTTPConnection | null = null;
   private _proxyConn: HTTPConnection | null = null;
 
+  public promptCache = new PromptCache({
+    cacheDir:
+      iso.getEnv("BRAINTRUST_PROMPT_CACHE_DIR") ??
+      "/tmp/braintrust/prompt_cache",
```
ankrgyl, line 312: better to use `$HOME/.braintrust/cache`, because the OS is allowed to wipe temporary directory files at any point. Also, it's not platform-safe to use `/tmp` (there's a portable tmpdir library in JS that you can use, but you should put that logic in the node implementation).

js/src/logger.ts
```diff
@@ -2697,23 +2707,36 @@ export async function loadPrompt({
...
+    const prompt = state.promptCache.get({
+      slug,
+      projectId,
+      projectName,
+      version: version ?? "latest",
+    });
+    if (!prompt) {
+      throw new Error(
+        `Prompt ${slug} (version ${version ?? "latest"}) not found in ${[projectName ?? projectId]} (not found on server or in local cache)`,
+      );
```
ankrgyl, line 2736: maybe throw or include the original exception here, so that the network error is surfaced to the user.

js/src/logger.ts
```diff
@@ -2728,8 +2751,12 @@ export async function loadPrompt({
   }
 
   const metadata = promptSchema.parse(response["objects"][0]);
-
-  return new Prompt(metadata, defaults || {}, noTrace);
+  const prompt = new Prompt(metadata, defaults || {}, noTrace);
+  state.promptCache.set(
```
ankrgyl, line 2755: assuming this is async, this is a "leaked" promise, so if it fails will make the program hang and lead to undefined behavior. You may want to just add a `.catch(console.warn)` here.

js/src/logger.ts
sachinpad: Done

js/src/logger.ts
sachinpad: Done

js/src/logger.ts
sachinpad: Done

ankrgyl (approved):
nice

PR #498, opened 2024-12-23
Add updated_at to AI Secret schema

tara-nagar:


core/js/typespecs/app_types.ts
```diff
@@ -138,6 +138,7 @@ export const aiSecretSchema = z
   .object({
     id: aiSecretBaseSchema.shape.id,
     created: aiSecretBaseSchema.shape.created,
+    updated_at: datetimeStringSchema.nullish(),
```
manugoyal, line 141: Maybe we should add an `updated_at` field to `generateBaseTableSchema`, so that we can use it elsewhere and include docs for the field.

manugoyal (approved):

core/js/typespecs/app_types.ts
```diff
@@ -44,6 +44,9 @@ function generateBaseTableSchema(
       .describe(
         `Date of ${objectName} deletion, or null if the ${objectName} is still active`,
       ),
+    updated_at: datetimeStringSchema
```
ibolmo, line 47: I think we need to update the openapi schema? I'm seeing a new error with the latest main. 
image

PR #500, opened 2024-12-23
add prompt cache to JS SDK

sachinpad:
to be stored as state


ankrgyl (commented):
Would suggest putting both of these files in a subdirectory within `sdk`

PR #505, opened 2024-12-25
add shim to invoke function

sachinpad:


js/src/functions/invoke.ts
```diff
@@ -187,3 +187,57 @@ export async function invoke<Input, Output, Stream extends boolean = false>(
...
+ * });
+ * ```
+ *
+ * @param options Options for the function.
+ * @param options.projectName The project name containing the function.
+ * @param options.slug The slug of the function to invoke.
+ * @param options.version Optional version of the function to use. Defaults to latest.
+ * @returns A function that can be used as a task or scorer in Eval().
+ */
+export function useFunction({
```
ankrgyl, line 221: `use` is a very special prefix in JS because it implies a hook in React. Let's think about other naming options offline. Also may be good to be consistent with the framework2 style (e.g. `project.scorer.create`, `project.scorer.load`, etc)?

sachinpad: Done

ankrgyl (approved):

PR #506, opened 2024-12-25
refactor prompt cache constructor

sachinpad:
to take the memory cache directly,
as we do the disk cache.


ankrgyl (approved):
seems fine? any particular driver behind this change?

sachinpad:
> seems fine? any particular driver behind this change?

Nope! Just thought the current signature was a bit weird

PR #514, opened 2025-01-08
parse Score in init_function

sachinpad:


py/src/braintrust/framework.py
```diff
@@ -1001,6 +1001,11 @@ async def await_or_run_scorer(root_span, scorer, name, **kwargs):
             scorer_args = kwargs
 
             result = await call_user_fn(event_loop, score, **scorer_args)
+            if isinstance(result, dict):
+                try:
+                    result = Score.from_dict(result)
+                except Exception as e:
+                    raise ValueError(f"When returning a dict, it must be a valid Score object. Got: {result}") from e
             if isinstance(result, Iterable):
```
ankrgyl, line 1009: should we change this to be elif?

sachinpad: I think we need L1016 to run right?

ankrgyl: ahh ok. mind adding a newline?

ankrgyl (approved):

PR #522, opened 2025-01-14
wrap openai moderations

sachinpad:


js/src/wrappers/oai.ts
```diff
@@ -348,83 +348,131 @@ function parseChatCompletionParams<P extends ChatParams>(
...
+  return async (allParams: T & SpanInfo, options?: unknown) => {
     const { span_info: _, ...params } = allParams;
     return traced(
       async (span) => {
-        // We could get rid of this type coercion if we could somehow enforce
-        // that `P extends EmbeddingCreateParams` BUT does not have the property
-        // `span_info`.
         const { data: result, response } = await create(
-          params as P,
+          params as T,
```
manugoyal, line 368: Can we get rid of this type coercion if we change the signature of `create` to this?

```
create: (params: Omit<T & SpanInfo, "span_info">, options?: unknown) => APIPromise<R>,
```

manugoyal (approved):

js/src/wrappers/oai.ts
sachinpad: done

Issue #523, opened 2025-01-14
Langchain traces in evals


HalfdanJ:
I'm unsure if this is intended to "just work" out of the box, or not, so forgive me if i'm just misunderstanding braintrust. 

I'm using the `setGlobalHandler` with langchain, and I can see logs come through in the global project when running my app. If i use the same function calls in an eval, I dont see any logs come through, neither in `logs` or traces in the eval. Should that just work? If so, what can I do to debug?

ankrgyl:
CC @ibolmo

ibolmo:
Thanks for asking. Things should work, but perhaps it may be helpful if you could share with me a repro example with evals not working.

ibolmo:
fyi, I _think_ I'm able to repro with:


```ts
import { BraintrustCallbackHandler } from "@braintrust/langchain-js";
import { ChatOpenAI } from "@langchain/openai";
import { Eval, initLogger } from "braintrust";

initLogger({
  projectName: "langchain-est",
  apiKey: process.env.BRAINTRUST_API_KEY,
});

const handler = new BraintrustCallbackHandler();

// setGlobalHandler(handler);

const data = [
  {
    input: "What is the capital of France?",
    expected: "Paris",
  },
];

const task = async (input: string) => {
  const model = new ChatOpenAI({
    modelName: "gpt-4o-mini",
  });

  const result = await model.invoke(input, {
    callbacks: [handler],
  });

  return String(result.content);
};

const test = () =>
  Eval("langchain-test", {
    data,
    task,
    scores: [
      ({ output }) => {
        return output.includes("Paris") ? 1 : 0;
      },
    ],
  });

async function main() {
  console.log(await task(data[0].input));
  // await test();
}

main();
```

I'm looking into this

ibolmo:
Ok I think there's possibly two issues:

1. `currentLogger` (if logger is not provided) seems to return a noop, and that may be causing an issue were reports are not sent. 
2. after providing the logger (see the updated example above), I'm not seeing the spans created by the logger to showup in the eval. I think that's the post here.

I'm going to look into both of those. They may be related.

ibolmo:
@HalfdanJ let me know if you run into any problems!

AnselmC:
@ibolmo that partially solved things but unfortunately it looks like all traces are attributed to a single datapoint:

[...]

PR #525, opened 2025-01-14
Fix langchain-js logging when run as part of an eval

ibolmo:
We used to try to autodetect or inititialize a logger. This was good DX, but in the event the user is running their code in an Eval it did not emit the spans to the eval session.

Now, we'll continue to accept an optional logger, but we'll use `startSpan` which will check for an experiment session or use a project logger or fallback to a noop span. In the case, that nothing is configure, we will warn the user that they'll want to configure logging. To keep backward compatibility, we'll set a `initLogger` (a default logger).

**Note**: regressed the attempt to add tags to the root span. If running in an experiment, we may not have the root span available to add tags. Instead, the tags will go to the span's metadata.

Fixes: https://github.com/braintrustdata/braintrust-sdk/issues/523


integrations/langchain-js/src/BraintrustCallbackHandler.ts
```diff
@@ -102,12 +104,24 @@ export class BraintrustCallbackHandler
       parentSpan = this.spans.get(parentRunId)!;
     } else if (!Object.is(currentParent, NOOP_SPAN)) {
       parentSpan = currentParent;
-    } else {
+    } else if (this.options.logger) {
+      // if provided, use the logger as the parent span
```
sachinpad, line 108: nit: start comment with a capital and end with a period

sachinpad (approved):

integrations/langchain-js/src/BraintrustCallbackHandler.ts
ibolmo: updated!

PR #527, opened 2025-01-16
Rework DictEvalHooks to actually be a dict

ankrgyl:
This makes it json serializable

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -937,6 +937,12 @@ class DictEvalHooks(EvalHooks):
 
         self.metadata.update(info)
 
+    def __json__(self):
+        return {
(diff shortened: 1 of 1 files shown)


py/src/braintrust/framework.py
```diff
@@ -12,6 +12,7 @@
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
 from importlib import metadata
+from json import JSONEncoder
```
sachinpad, line 15: unused?

sachinpad (approved):

py/src/braintrust/framework.py
ankrgyl: fixed

PR #530, opened 2025-01-20
Fix support for binary packages

ankrgyl:
We need to do some hula hoops to identify which packages have binary dependencies, and mark them as external.


js/src/cli.ts
```diff
@@ -696,16 +696,80 @@ async function collectFiles(
...
+        const path = require.resolve(args.path, { paths: [args.resolveDir] });
+        const match = path.match(
+          /node_modules[/\\]((?:@[^/\\]+[/\\])?[^/\\]+)/,
+        );
+        if (match) {
+          addNativePackage(match[1]);
+        }
+        return { path: args.path, external: true };
+      } catch {
+        return { path: args.path, external: true };
```
ibolmo, line 744: should we just `return { path: args.path, external: true };` outside of the try/catch?

js/src/cli.ts
```diff
@@ -696,16 +696,80 @@ async function collectFiles(
...
+      }
+    });
+
+    // Handle direct imports of native packages
+    build.onResolve(
+      { filter: /@[^/]+\/[^/]+-(?:darwin|linux|win32)/ },
```
ibolmo, line 750: is the @ always preceding?

js/src/cli.ts
```diff
@@ -696,16 +696,80 @@ async function collectFiles(
 // our packages (braintrust and autoevals) external, in case they're
 // installed in a relative path.
 const markOurPackagesExternalPlugin = {
-  name: "make-all-packages-external",
+  name: "make-our-packages-external",
   setup(build: esbuild.PluginBuild) {
-    const filter = /^(braintrust|autoevals|@braintrust\/)/;
-    build.onResolve({ filter }, (args) => ({
+    // Mark our packages as external
+    const ourPackagesFilter = /^(braintrust|autoevals|@braintrust\/)/;
```
ibolmo, line 702: nit (feel free to ignore)

move to a constant above?

js/src/cli.ts
```diff
@@ -696,16 +696,80 @@ async function collectFiles(
...
+    // Keep track of packages that contain .node files
+    const nativePackages = new Set<string>();
+
+    // Helper to add a package and its platform-specific variants
+    const addNativePackage = (pkgName: string) => {
+      nativePackages.add(pkgName);
+      if (pkgName.includes("@")) {
+        const [scope, name] = pkgName.split("/");
+        const platformPkgs = [
+          `${scope}/${name}-darwin-arm64`,
```
ibolmo, line 723: are these standard suffixes?  quick grep looks like it. looks like (by searching our repo as an example and finding in our lock files) there's some variants like -ia32- missing. what determines the targets?

ibolmo (approved):
is the test just making sure the CI builds correctly? 

do we have an example in monorepo that could recreate the build error and with this update shows that it is now gone?

sachinpad (approved):

js/src/cli.ts
ankrgyl: I just had an AI generate them. Happy to add to this list if you have specific recommendations or we can keep adding to it as we go

js/src/cli.ts
ankrgyl: yep

js/src/cli.ts
ankrgyl: good question, another thing I'd rather stay conservative with at the start and iterate on.

js/src/cli.ts
ibolmo: just notice this notification 😅 <br><br>no strong feelings here. we can always come back to add more as needed 👍

Issue #521, opened 2025-01-13
JSON error when passing image_url


luiz-alma:
When I'm passing a formatted context to `braintrust.build` like so: `prompt.build(context=context, query=query, image_url=image_url)`, I'm getting `json.decoder.JSONDecodeError: Invalid control character at: line 1 column 23 (char 22)`. That happens whenever `context` has some line breaks.

I tried removing the `image_url` and it works. Also tried removing the `context` and leaving `image_url`, which also works. Seems like there's some issue with parsing the prompt variables.

ankrgyl:
Hi there, can you share a repro (ideally the prompt, or a similar prompt that still reproduces the issue)?

luiz-alma:
Here's the simplest version that's failing:
```python
query_variables = {'context': '\n        ###### USER CONTEXT ######', 'query': 'is avocado ok?', 'image_url': ''}
```

I basically call this with:
`prompt.built(context=query_variables["context"], query=query_variables["query"], image_url=query_variables["image_url"])`

ankrgyl:
We need a prompt definition that reproduces it to be able to assist you further.

luiz-alma:
Does it work if I send a prompt_id and revision?

ankrgyl:
Only if you add one of us, e.g. ankur@braintrustdata.com, as a user to your account with "Engineer" privileges.

luiz-alma:
It's really simple implementation:
<img width="717" alt="Screenshot 2025-01-13 at 1 24 24 PM" src="https://github.com/user-attachments/assets/51ff424f-5263-40ec-9468-b45dbbf24746" />

luiz-alma:
@ankrgyl any updates?

ankrgyl:
Hi @luiz-alma we confirmed the repro and will take a look at it sometime this week.

ibolmo:
hi @luiz-alma I've got a possible fix here. will follow-up when we get this ready for you to try

luiz-alma:
awesome! thanks for the update

PR #539, opened 2025-01-24
[BRA-1943] adopt rendering of messages like in ts SDK library

ibolmo:
Adopt typescript solution to rendering prompt messages (looks like we had the method, but didn't utilize it). 

Fixes: https://github.com/braintrustdata/braintrust-sdk/issues/521


py/src/braintrust/logger.py
```diff
@@ -3353,20 +3346,55 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
...
-            if c.type == "text"
-            else {**c.as_dict(), "image_url": {**c.image_url, "url": render(c.image_url.url)}}
-            if c.type == "image_url"
-            else c
-            for c in message.content
-        ],
-    }
+def render_message(render: Callable[[str], str], message: PromptMessage):
+    base = {k: v for (k, v) in message.as_dict().items() if v is not None}
+    # TODO: shouldn't load_prompt guarantee content is a PromptMessage?
```
ibolmo, line 3351: @ankrgyl would you expect this as well? I can track it down. I imagine load_prompt should be trying something akin to `PromptMessage(**serialized)`.

py/src/braintrust/logger.py
```diff
@@ -3353,20 +3346,55 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
...
+            base["content"] = render(content)
+        else:
+            rendered_content = []
+            for c in content:
+                if isinstance(c, str):
+                    rendered_content.append(c)
+                    continue
+
+                if not isinstance(c, dict):
+                    c = c.as_dict()
```
ibolmo, line 3364: in my test this `c` was not instance of TextPart or ImagePart

aphinx (approved):

PR #542, opened 2025-01-26
[BRA-1957] Remove noisy warnings and allow NOOP_SPAN for langchain-js integration

ibolmo:
DX improvements from a community request.

### DIFF
--- a/integrations/langchain-js/src/BraintrustCallbackHandler.ts
+++ b/integrations/langchain-js/src/BraintrustCallbackHandler.ts
@@ -2,8 +2,6 @@ import { isObject } from "@braintrust/core";
 import {
   BaseCallbackHandler,
   BaseCallbackHandlerInput,
-  HandleLLMNewTokenCallbackFields,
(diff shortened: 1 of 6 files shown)


integrations/langchain-js/src/BraintrustCallbackHandler.ts
```diff
@@ -196,25 +198,6 @@ export class BraintrustCallbackHandler<IsAsyncFlush extends boolean>
     );
   }
 
-  async handleLLMNewToken(
```
aphinx, line 199: are these deleted because they were spewing unhelpful warnings? or just cleanup?

aphinx (approved):
seems fine to me

integrations/langchain-js/src/BraintrustCallbackHandler.ts
ibolmo: indeed. a customer noticed that these would fill their logs and it wasn't helpful

PR #545, opened 2025-01-28
support templated structured outputs in SDK

sachinpad:


js/src/logger.ts
```diff
@@ -4703,6 +4707,80 @@ export type PromptRowWithId<
     ? Pick<PromptRow, "_xact_id">
     : Partial<Pick<PromptRow, "_xact_id">>);
 
+export function deserializePlainStringAsJSON(s: string) {
```
manugoyal, line 4710: If this function isn't used anywhere else we don't need to export it.

js/src/logger.ts
```diff
@@ -4703,6 +4707,80 @@ export type PromptRowWithId<
...
+  }
+
+  try {
+    return { value: JSON.parse(s), error: undefined };
+  } catch (e) {
+    return { value: s, error: e };
+  }
+}
+
+function renderTemplatedObject(obj: unknown, args: unknown): unknown {
```
manugoyal, line 4722: Maybe put these in a separate file for cleanliness?

py/src/braintrust/logger.py
```diff
@@ -3369,6 +3369,55 @@ def render_message(render, message):
     }
 
 
+_base_get_key = chevron.renderer._get_key
```
manugoyal, line 3372: Might be nice to wrap this up into a separate module rather than globally monkey-patching the library (note that this monkey-patching will affect anybody who imports our library). ChatGPT says it's possible to re-wrap functions in python with a custom global environment, so maybe we can try that?
```
import types
import functools

def copy_function_with_new_globals(original_func, new_globals):
    """
    Create a copy of the original function with a new global namespace.
    
    Parameters:
    - original_func: The function to be copied.
    - new_globals: A dictionary representing the new global variables and functions.
    
    Returns:
    - A new function identical to original_func but using new_globals as its global namespace.
    """
    new_func = types.FunctionType(
        original_func.__code__,
        new_globals,
        name=original_func.__name__,
        argdefs=original_func.__defaults__,
        closure=original_func.__closure__
    )
    new_func = functools.update_wrapper(new_func, original_func)
    new_func.__kwdefaults__ = original_func.__kwdefaults__
    return new_func
```

js/src/logger.ts
sachinpad: Kept the export but actually used it

js/src/logger.ts
sachinpad: Is it worth it to have a separate file just for the render-related stuff?

py/src/braintrust/logger.py
sachinpad: Done

js/src/logger.ts
manugoyal: I figured if these functions are being used in other places then a separate file makes it somewhat clear that these are self-contained utils. Fine either way.

py/src/braintrust/logger.py
```diff
@@ -3416,6 +3417,71 @@ def render_message(render: Callable[[str], str], message: PromptMessage):
     return base
 
 
+def _create_custom_render():
```
manugoyal, line 3420: nit: a good instance where moving to a separate file makes it clear that we don't need/use the `chevron` import anywhere else. Or we could move the `import chevron` into this function.

manugoyal (approved):

py/src/braintrust/logger.py
sachinpad: Hmm looks like we also use it elsewhere in prompt.build, e.g. L3571

PR #546, opened 2025-01-29
fix pyright for logger.py

sachinpad:


py/src/braintrust/logger.py
```diff
@@ -2392,7 +2402,9 @@ def compute_parent_object_id():
             return parent_object_id.get()
 
         arg_parent_object_id = LazyValue(compute_parent_object_id, use_mutex=False)
-        if parent_components.row_id:
+        if parent_components.row_id is not None:
```
manugoyal, line 2405: Hmm kind-of annoying. Maybe we should change the `span_identifier_v3.py` representation to store these three fields in their own TypedDict, which is Optional, so that the type system can make this inference naturally.

manugoyal (commented):
nice! Just one thought about the span components v3 typing.

PR #552, opened 2025-02-07
Origin created

j13huang:
Add created field for datasets into origin

### DIFF
--- a/core/js/typespecs/common_types.ts
+++ b/core/js/typespecs/common_types.ts
@@ -116,6 +116,6 @@ export const objectReferenceSchema = z.object({
   _xact_id: z.string().describe("Transaction ID of the original event."),
   created: z
     .string()
-    .optional()
+    .nullish()
     .describe("Created timestamp of the original event."),
 });
--- a/core/js/typespecs/common_types.ts
+++ b/core/js/typespecs/common_types.ts
@@ -114,5 +114,8 @@ export const objectReferenceSchema = z.object({
     .describe("ID of the object the event is originating from."),
   id: z.string().describe("ID of the original event."),
   _xact_id: z.string().describe("Transaction ID of the original event."),
-  created: z.string().describe("Created timestamp of the original event."),
+  created: z
+    .string()
+    .optional()
+    .describe("Created timestamp of the original event."),
 });
--- a/core/js/typespecs/common_types.ts
+++ b/core/js/typespecs/common_types.ts
@@ -114,4 +114,5 @@ export const objectReferenceSchema = z.object({
     .describe("ID of the object the event is originating from."),
   id: z.string().describe("ID of the original event."),
   _xact_id: z.string().describe("Transaction ID of the original event."),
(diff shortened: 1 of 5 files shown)


j13huang:
Hmm this needs to be a number I think

core/js/typespecs/common_types.ts
```diff
@@ -114,4 +114,8 @@ export const objectReferenceSchema = z.object({
     .describe("ID of the object the event is originating from."),
   id: z.string().describe("ID of the original event."),
   _xact_id: z.string().describe("Transaction ID of the original event."),
+  created: z
+    .string()
+    .optional()
```
ankrgyl, line 119: would make it `nullish()` rather than optional, because sometimes query engines will fill it in as `null` to match the schema.

js/src/logger.ts
```diff
@@ -3640,6 +3644,7 @@ export type EvalCase<Input, Expected, Metadata> = {
   // These fields are only set if the EvalCase is part of a Dataset.
   id?: string;
   _xact_id?: TransactionId;
+  created?: string;
```
ankrgyl, line 3647: along the lines of nullish, make this `created?: string | null`.

ankrgyl (commented):
Make sure to make the equivalent logging changes for python too

Also think you need to update [here](https://github.com/braintrustdata/braintrust-sdk/blob/main/js/src/framework.ts#L790)

core/js/typespecs/common_types.ts
j13huang: done

js/src/logger.ts
j13huang: Not sure if you were looking at an old set of changes or what, but it was added https://github.com/braintrustdata/braintrust-sdk/pull/552/files#diff-22281c487aa6e63328442a1eb413f2a5d5913f854e449155ccc4c2154bec12adR790

j13huang: done

j13huang:
> Also think you need to update [here](https://github.com/braintrustdata/braintrust-sdk/blob/main/js/src/framework.ts#L790)

Not sure if you were looking at an old set of changes or what, but it was added https://github.com/braintrustdata/braintrust-sdk/pull/552/files#diff-22281c487aa6e63328442a1eb413f2a5d5913f854e449155ccc4c2154bec12adR790

core/js/typespecs/common_types.ts
```diff
@@ -114,4 +114,8 @@ export const objectReferenceSchema = z.object({
     .describe("ID of the object the event is originating from."),
   id: z.string().describe("ID of the original event."),
   _xact_id: z.string().describe("Transaction ID of the original event."),
+  created: z
+    .string()
+    .nullish()
+    .describe("Created timestamp of the original event."),
```
ankrgyl, line 120: maybe add "Used to help sort views in the UI."

ankrgyl (approved):

PR #556, opened 2025-02-15
Support bundle/attachment uploads to Azure

ankrgyl:
You need to add a special header

### DIFF
--- a/js/src/functions/upload.ts
+++ b/js/src/functions/upload.ts
@@ -369,12 +369,15 @@ async function uploadBundles({
       bundleStream.on("error", reject);
     });
 
+    const headers = {
+      "Content-Encoding": "gzip",
+    };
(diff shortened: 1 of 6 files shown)


sachinpad (approved):

js/src/functions/upload.ts
```diff
@@ -369,12 +369,15 @@ async function uploadBundles({
       bundleStream.on("error", reject);
     });
 
+    const headers = {
+      "Content-Encoding": "gzip",
+    };
+    addAzureBlobHeaders(headers, pathInfo.url);
```
tara-nagar, line 375: missing function import here

ankrgyl: whoops, not sure how this builds... but will fix

PR #557, opened 2025-02-19
Eval default error score

j13huang:


js/src/framework.ts
```diff
@@ -245,6 +245,11 @@ export interface Evaluator<
    * Optionally explicitly specify the git metadata for this experiment. This takes precedence over `gitMetadataSettings` if specified.
    */
   repoInfo?: RepoInfo;
+
+  /**
+   * Optionally generate 0% value scores for any scorer that errored. For a task that errors, all scores will be given a 0% value.
+   */
+  defaultErrorScore?: boolean;
```
ankrgyl, line 252: maybe it would be clearer to say `scoreZeroOnError`

js/src/framework.ts
```diff
@@ -245,6 +245,18 @@ export interface Evaluator<
    * Optionally explicitly specify the git metadata for this experiment. This takes precedence over `gitMetadataSettings` if specified.
    */
   repoInfo?: RepoInfo;
+
+  /**
+   * Optionally supply a custom function to specifically handle score values when tasks or scoring functions have errored.
+   * If set to true, log a 0 score to the root span for any scorer that was not run. For a task that errored, all scores will to the root span with a 0 value.
+   */
+  unhandledScoresFallback?:
```
ankrgyl, line 253: can you rename this to

```
errorScoreHandler
```

ankrgyl (commented):
Please implement in Python too

js/src/framework.ts
```diff
@@ -245,6 +245,18 @@ export interface Evaluator<
    * Optionally explicitly specify the git metadata for this experiment. This takes precedence over `gitMetadataSettings` if specified.
    */
   repoInfo?: RepoInfo;
+
+  /**
+   * Optionally supply a custom function to specifically handle score values when tasks or scoring functions have errored.
+   * If set to true, log a 0 score to the root span for any scorer that was not run. For a task that errored, all scores will be logged to the root span with a 0 value.
+   */
+  errorScoreHandler?:
```
sachinpad, line 253: should we instead export defaultErrorScoreHandler and let the user set that if they want the default behavior rather than use a union with true?

js/src/framework.test.ts
```diff
@@ -158,3 +158,164 @@ test("expected (read/write) is passed to task", async () => {
...
+          ),
+        },
+        new BarProgressReporter(),
+        [],
+        undefined,
+      );
+
+      expect(
+        out.results.every((r) => Object.keys(r.scores).length === 0),
+      ).toBeTruthy();
```
sachinpad, line 200: should we check if it is equal to true rather than truthy?

js/src/framework.test.ts
```diff
@@ -158,3 +158,164 @@ test("expected (read/write) is passed to task", async () => {
...
+      throw new Error("scorer errored");
+    }
+    return {
+      name,
+      score: 1,
+    };
+  };
+}
+
+describe("runEvaluator", () => {
```
sachinpad, line 177: maybe we should copy this same test in python as well?

js/src/framework.test.ts
```diff
@@ -158,3 +162,166 @@ test("expected (read/write) is passed to task", async () => {
...
+          evalName: "eval",
+          data: [{ input: 1 }],
+          task: async () => {
+            throw new Error("test error");
+          },
+          scores: Array.from({ length: 3 }, (_, i) =>
+            makeTestScorer(`scorer_${i}`),
+          ),
+        },
+        new BarProgressReporter(),
```
sachinpad, line 197: nit: do we need any reporter? ideally don't want to pollute the stdout

js/src/framework.ts
```diff
@@ -245,6 +245,16 @@ export interface Evaluator<
    * Optionally explicitly specify the git metadata for this experiment. This takes precedence over `gitMetadataSettings` if specified.
    */
   repoInfo?: RepoInfo;
+
+  /**
+   * Optionally supply a custom function to specifically handle score values when tasks or scoring functions have errored.
+   * A default implementation is exported as `defaultErrorScoreHandler` which will log a 0 score to the root span for any scorer that was not run.
+   */
+  errorScoreHandler?: (args: {
```
sachinpad, line 253: nit: define the func type separately and export it?

py/src/braintrust/framework.py
```diff
@@ -553,6 +562,9 @@ def _EvalCommon(
     base_experiment_id: Optional[str],
     git_metadata_settings: Optional[GitMetadataSettings],
     repo_info: Optional[RepoInfo],
+    error_score_handler: Optional[
```
sachinpad, line 565: nit: ditto re. typedefs in py

sachinpad (approved):

js/src/framework.test.ts
j13huang: Copied from previous tests, and doesn't seem to print out anything, but I can create a noop reporter

PR #558, opened 2025-02-19
Account for weird types that are returned by error.stack

akulkapoor:
We have seen intermittent issues where trace.split seems to be running on something that is not a string. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack seems to indicate that the behavior may not be consistent. This is an attempt to address that.

### DIFF
--- a/js/src/stackutil.ts
+++ b/js/src/stackutil.ts
@@ -8,7 +8,7 @@ export interface StackTraceEntry {
 
 export function getStackTrace(): StackTraceEntry[] {
   const trace = new Error().stack;
-  if (trace === undefined) {
+  if (typeof trace !== "string") {
     return [];
   }
   const traceLines = trace.split("\n");


jackfischer:
Due to how braintrust seemingly instruments nextjs, these exceptions kill requests and we lose logging and the ability to catch the exception, at least in a straightforward way.

From MDN,
>Non-standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

This appears to be a dark corner that we just have to work around.

js/src/stackutil.ts
```diff
@@ -8,7 +8,7 @@ export interface StackTraceEntry {
 
 export function getStackTrace(): StackTraceEntry[] {
   const trace = new Error().stack;
-  if (trace === undefined) {
+  if (typeof trace !== "string") {
     return [];
   }
   const traceLines = trace.split("\n");
```
ibolmo, line 14: @akulkapoor how about we coerce here:

```ts
String(trace).split('\n')
```

worst case we have an useless trace, and best case we notice what the weird trace is

ibolmo (approved):

PR #567, opened 2025-03-04
Fix deep-copy logic in python logger

manugoyal:
We were using `copy.deepcopy` to deep-copy objects in general. But we found instances in the wild where this doesn't work. E.g. somebody was trying to log an openAI `AsyncHttpxClientWrapper`, which was getting deepcopied, and then partially-deleted before the logger ran.

So instead, we just always resort to a JSON roundtrip unless we the object is a known primitive.


ankrgyl (approved):

ibolmo (approved):
That must have been gnarly.

PR #569, opened 2025-03-04
fix async flush

sachinpad:


js/src/logger.ts
```diff
@@ -2668,7 +2668,7 @@ export function initLogger<IsAsyncFlush extends boolean = true>(
   const {
     projectName,
     projectId,
-    asyncFlush,
+    asyncFlush: asyncFlushArg,
```
manugoyal, line 2671: While we're at it, should we update all the other places where `IsAsyncFlush` is defaulted? E.g. https://github.com/braintrustdata/braintrust-sdk/blob/8fc81c5d906bbeb49e0f3c638b3d518a9e150680/js/src/logger.ts#L3232.

sachinpad: Done

js/src/logger.ts
```diff
@@ -3291,7 +3294,7 @@ function startSpanAndIsLogger<IsAsyncFlush extends boolean = false>(
         components.data.object_type === SpanObjectTypeV3.PROJECT_LOGS &&
         // Since there's no parent logger here, we're free to choose the async flush
         // behavior, and therefore propagate along whatever we get from the arguments
-        !args?.asyncFlush,
+        (args?.asyncFlush === undefined ? false : !args?.asyncFlush),
```
manugoyal, line 3297: A simpler boolean expression would be `args?.asyncFlush === false`.

js/src/logger.ts
```diff
@@ -3301,7 +3304,10 @@ function startSpanAndIsLogger<IsAsyncFlush extends boolean = false>(
     return {
       span,
       isSyncFlushLogger:
-        parentObject.kind === "logger" && !parentObject.asyncFlush,
+        parentObject.kind === "logger" &&
+        (parentObject.asyncFlush === undefined
```
manugoyal, line 3308: same here

manugoyal (approved):
lgtm! I am actually about to deploy something which already bumped the package version to `0.0.187`, so I can just hold off until we merge this too.

js/src/logger.ts
sachinpad: Done

js/src/logger.ts
sachinpad: Done

PR #571, opened 2025-03-06
allow span_attributes to pass through log

sachinpad:


manugoyal (commented):
Can we add to the corresponding type signatures in the JS logger?

manugoyal:
Maybe also good to validate in python that the `span_attributes` object is a dict.

PR #578, opened 2025-03-11
fix prompt pull for long prompts

ankrgyl:


js/src/cli-util/pull.ts
```diff
@@ -219,7 +219,7 @@ function makeFunctionDefinition({
   const promptContents =
     prompt.type === "completion"
       ? `prompt: ${doubleQuote(prompt.content)}`
-      : `messages: ${util.inspect(prompt.messages, { depth: null }).trimStart()}`;
+      : `messages: ${util.inspect(prompt.messages, { depth: null, maxStringLength: Infinity }).trimStart()}`;
```
ibolmo, line 222: Curious why we use inspect vs json.stringify

ibolmo (approved):

js/src/cli-util/pull.ts
ankrgyl: I can't remember, but I'm sure I started with `stringify` 😓

PR #589, opened 2025-03-17
Pass --no-ext-diff flag to git diff

jkimbo:
If the user has an external diff provider setup in git the call to `git diff HEAD` hangs forever. The fix is to always pass the `--no-ext-diff` flag.


ankrgyl:
Thanks so much for contributing this!

PR #590, opened 2025-03-18
Add js support for aborting evaluations

choochootrain:


js/src/framework.ts
```diff
@@ -682,17 +687,22 @@ export async function runEvaluator(
...
+      if (evaluator.signal) {
+        evaluator.signal.addEventListener("abort", () => {
+          reject("evaluator aborted");
+        });
+      }
     });
     return null;
   };
-  const winner = await Promise.race([result, timer()]);
+  const winner = await Promise.race([result, cancel()]);
```
j13huang, line 705: Does this actually abort the eval? I think we would want to do something like `q.kill` here

https://github.com/braintrustdata/braintrust-sdk/blob/945656d9dd3e48ded82ae075761849b0b6de7d99/js/src/framework.ts#L1037

https://caolan.github.io/async/v3/docs.html#QueueObject

Correct me if I'm wrong

j13huang:
Also we should probably have a python implementation as well

choochootrain:
we wouldn't use the python implementation internally so i would only implement if sdk parity for this is important - cc @manugoyal

js/src/framework.ts
choochootrain: it will early return just like timeout will. i suppose we could put some extra effort into ensuring the operation is actually aborted but given that it will only reliably work for `fetch` type operations you would have to expose it to a bunch of other things and downstream users if they write their own functions

the `kill` thing doesn't really do anything that this doesnt, just simpler semantics: https://github.com/caolan/async/blob/master/lib/internal/queue.js#L177

j13huang: Let's chat tomorrow so I can understand

choochootrain: chatted offline - at the very least we need to prevent additional logs from being streamed from the eval

choochootrain: jk you were right, the internal implementation uses `queue` for concurrency and i can dump the pending tasks on abort

js/src/framework.ts
```diff
@@ -1034,7 +1024,37 @@ async function runEvaluatorInternal(
...
+  };
+
+  // wait for tasks to be completed or the evaluator to be cancelled
+  // if the evaluator is cancelled, the remaining tasks that have not been started will be killed
+  try {
+    await Promise.race([q.drain(), cancel()]);
+  } catch (e) {
+    if (
+      e instanceof Error &&
+      (e.message === "Evaluator timed out" || e.message === "Evaluator aborted")
```
j13huang, line 1050: nit: Does it work to have the errors as constants and then check against those constants?

j13huang (approved):
Nice!

js/src/framework.test.ts
```diff
@@ -17,29 +18,84 @@ class NoopProgressReporter implements ProgressReporter {
...
-        data: [{ input: 1, expected: 2 }],
+        data: Array.from({ length: 10 }, (_, i) => ({
+          input: i,
+          expected: i * 2,
+        })),
         task: async (input: number) => {
-          await new Promise((r) => setTimeout(r, 100000));
+          taskStarts.add(input);
+          if (input > 2) {
+            await new Promise((r) => setTimeout(r, 100000));
```
j13huang, line 78: Sorry for being extra anal here but I think this test won't actually verify the queue is actually killed since tasks will never finish past this line even if kill was never called in the code.

Maybe you can do `if (input > 2) {abortController.abort();}` here instead of a timeout

I also found https://jestjs.io/docs/timer-mocks (`jest.runAllTimers`) which may help

j13huang (approved):

PR #592, opened 2025-03-19
Bump py core version

choochootrain:
whoops forgot this one


choochootrain:
nvm this isn't needed since no changes to py core were made

PR #593, opened 2025-03-19
Ensure `permalink` works with no-op spans.

clutchski:
It was throwing errors with no-op spans. Fixes #BRA-816.

This needs a test with real spans as well.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -1335,18 +1335,20 @@ export async function permalink(
     return state.appUrl;
   };
 
-  const components = SpanComponentsV3.fromStr(slug);
-  const object_type = spanObjectTypeV3ToString(components.data.object_type);
-  const [orgName, appUrl, object_id] = await Promise.all([
-    getOrgName(),
-    getAppUrl(),
-    spanComponentsToObjectId({ components, state }),
-  ]);
-  const id = components.data.row_id;
-  if (!id) {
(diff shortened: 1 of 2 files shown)


js/src/logger.ts
```diff
@@ -1335,18 +1335,20 @@ export async function permalink(
...
-    getAppUrl(),
-    spanComponentsToObjectId({ components, state }),
-  ]);
-  const id = components.data.row_id;
-  if (!id) {
-    throw new Error("Span slug does not refer to an individual row");
-  }
-  const urlParams = new URLSearchParams({ object_type, object_id, id });
+  let urlParams = "span=noop-span";
+  // no-op spans have an empty slug, so make sure we handle permalinking them gracefully.
```
manugoyal, line 1339: It may be worthwhile to have the noop span case actually do nothing, including skipping the `getOrgName/getAppUrl` stuff which may log in to braintrust, to further the cause of "the SDK does nothing when you are working with noop spans".

Maybe we could just return a dummy URL like `https://www.braintrust.dev/noop`

manugoyal (commented):
nice! Do we also need to fix this for the python logger?

js/src/logger.ts
clutchski: good point. I didn't realize those logged in.

PR #596, opened 2025-03-20
Use custom HTTPAdapter which retries more errors

manugoyal:
The standard `Retry`-based policy does not necessarily handle all forms of request-related exceptions. For example, some exceptions may be thrown only when downloading the response content, so we must do that ahead of time and retry in case of error. We accomplish this with a custom retry policy.


py/src/braintrust/logger.py
```diff
@@ -400,6 +401,34 @@ def set_http_adapter(adapter: HTTPAdapter) -> None:
...
+                # downloading.
+                if response.content:
+                    pass
+                return response
+            except (urllib3.exceptions.HTTPError, requests.exceptions.RequestException) as e:
+                if num_prev_retries < self.base_num_retries:
+                    # Emulates the sleeping logic in the backoff_factor of urllib3 Retry
+                    sleep_s = self.backoff_factor * (2**num_prev_retries)
+                    print("Retrying request after error:", e)
+                    print("Sleeping for", sleep_s, "seconds")
```
ankrgyl, line 425: rm or print to stderr/warn?

py/src/braintrust/logger.py
```diff
@@ -400,6 +401,34 @@ def set_http_adapter(adapter: HTTPAdapter) -> None:
         _state._api_conn._reset()
 
 
+class LongLivedAdapter(HTTPAdapter):
```
ankrgyl, line 404: wonder if we can make this compose with other adapters?

ankrgyl: Cool, maybe we can add it to the docs too

py/src/braintrust/logger.py
manugoyal: Made it print to stderr

ankrgyl (approved):

PR #601, opened 2025-03-24
WIP Trace anthropic

clutchski:
This adds tracing for Anthropic's python client.

So far I've done about half the API. 

- [ ] batch results
- [x] messages sync
- [x] messages streaming sync
- [x] messages streaming async
- [ ] text completions sync
- [ ] text completions streaming async
- [ ] text completions streaming sync

[[Here's](https://www.braintrust.dev/app/matt-test-org/p/example-anthropic-app/logs?r=79c9ecd8-3666-4c37-84de-dfe65535ec03&s=4df269ad-f067-4cb3-abcc-4b49328de4ad) an example trace and a screenshot.

Screenshot 2025-03-24 at 4 41 37 PM

Notes:

- I think memory logger is a very nice way to unit testing the application, keeping the tests alongside the code & making the library more open to the community.
- Also adds a helper library (only in dev) to do async testing. If there's a better way to do this without the app, please let me know.


py/src/braintrust/logger.py
```diff
@@ -524,11 +526,42 @@ def _check_json_serializable(event):
...
+    @abstractmethod
+    def log(self, *args: LazyValue[Dict[str, Any]]) -> None:
+        pass
+
+    @abstractmethod
+    def flush(self, batch_size: Optional[int] = None):
+        pass
+
+
+class _MemoryBackgroundLogger(_BackgroundLogger):
```
sachinpad, line 539: maybe we should move this out of logger.py into some testonly library?

sachinpad (commented):
Generally LGTM!

re. testing, it's probably best to avoid making actual API calls, especially to a backend not owned by us. I think there are a few possible answers here: a) test in the main repo, where we can leverage our test-proxy; b) add some HTTP record/replay mechanism in this repo for testing

sachinpad:
Also, I don't think we need to worry about batch or text completion APIs. We haven't wrapped the batch API for openai, and text completions is legacy; I actually don't know if we have any anthropic completion models

py/src/braintrust/logger.py
clutchski: I left it in there becuase there is an already a testing one in there.

clutchski:
@sachinpad re: moving this code because it calls the anthropic API.  I disagree for the following reasons:

- Unless there's a very compelling reason, tests should live alongside the code. It makes things greppable, understandable, etc. 
- This is an open source repo, therefore tests should be open source. If they aren't community members won't be able to contribute and also might assume there aren't any tests and the code is low-quality.
- We won't get any benefit. It will still call the API anyway.

py/examples/.gitignore
```diff
@@ -0,0 +1 @@
+run.sh
```
sachinpad, line 1: Is this leftover from your local env?

sachinpad (approved):
Looks good!

PR #602, opened 2025-03-25
Fix python logger retry policy

manugoyal:
The retry policy tries to download the full request payload to ensure we don't hit any download errors. But this seems to fail for redirect-type requests if the content-encoding doesn't match the redirect data. So we just skip this check for redirects.


choochootrain (approved):
for my understanding - why do you have to update the core version here?

sachinpad (approved):

PR #555, opened 2025-02-14
Allow the objectFetcher to use a btql query for initDataset

j13huang:
Things I still need to figure out:
~- [ ] how to pull in btql types into this repo~
~- [ ] python types?~

Any API/interface comments appreciated

PR #607, opened 2025-03-26
Internal btql paginate results

j13huang:
follow up to https://github.com/braintrustdata/braintrust-sdk/pull/555

Fixes https://braintrustdata.slack.com/archives/C07GDHU71D1/p1742946401868119


js/src/logger.ts
```diff
@@ -3743,34 +3744,43 @@ class ObjectFetcher<RecordType>
...
-              ],
-              from: {
-                op: "function",
-                name: {
-                  op: "ident",
-                  name: [this.objectType],
-                },
-                args: [
+        let cursor = undefined;
+        while (true) {
```
ankrgyl, line 3748: may want to limit the # of iterations here (say to 10,000) and throw an exception after that.

js/src/logger.ts
```diff
@@ -3743,34 +3744,43 @@ class ObjectFetcher<RecordType>
...
+                  name: {
+                    op: "ident",
+                    name: [this.objectType],
+                  },
+                  args: [
+                    {
+                      op: "literal",
+                      value: await this.id,
+                    },
+                  ],
```
manugoyal, line 3770: Do we want to add an explicit limit here?

js/src/logger.ts
```diff
@@ -490,13 +490,14 @@ export class BraintrustState {
...
     }
-    const newState = await loginToState({
+    const { state: newState, conn } = await loginToState({
       ...this.loginParams,
       ...Object.fromEntries(
         Object.entries(loginParams).filter(([k, v]) => !isEmpty(v)),
       ),
     });
     this.copyLoginInfo(newState);
+    this.loginReplaceApiConn(conn);
```
manugoyal, line 500: I wonder if instead we should call `this.loginReplaceApiConn` [here](https://github.com/braintrustdata/braintrust-sdk/blob/b432e3b9926b4b379abe068cf156dfc4060810b3/js/src/logger.ts#L410) and leave this as-is.

manugoyal (approved):
lgtm! let's just make sure the internal-repo PR passes all tests before we merge this.

PR #610, opened 2025-03-27
anthropic: ensure the library runs without anthropic

clutchski:
- `wrap_anthropic` would fail if anthropic wasn't installed
-  fixed the nox tests (which weren't quite running properly)
- added some tests to ensure our library can always be imported and run core tests without extra dependencies

### DIFF
--- a/py/noxfile.py
+++ b/py/noxfile.py
@@ -3,11 +3,17 @@ Define our tests to run against different combinations of
 Python & library versions.
 """
 
-
 import nox
 
 LATEST = "__latest__"
 SRC = "src/braintrust"
+
+
+ERROR_CODES = tuple(range(1, 256))
+
+# List your package here if it's not guaranteed to be installed. We'll (try to)
+# validate things work with or with them.
+VENDOR_PACKAGES = ("anthropic", "openai")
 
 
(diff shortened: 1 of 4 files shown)


py/noxfile.py
```diff
@@ -3,25 +3,49 @@
...
 import nox
 
 LATEST = "__latest__"
 SRC = "src/braintrust"
 
 
+ERROR_CODES = tuple(range(1, 256))
+
+# List your package here if it's not guaranteed to be installed. We'll (try to)
+# validate things work with or with them.
```
choochootrain, line 15:
```suggestion
# validate things work with or without them.
```
?

py/setup.py
```diff
@@ -41,6 +41,8 @@
     ],
     "doc": ["pydoc-markdown"],
     "openai-agents": ["openai-agents"],
+    # These should only be installed for linting import errors, not for tests.
+    "lint": ["anthropic"],
```
choochootrain, line 45: i might be reading this wrong but do you still need this for linting if you've removed the import and type annotations below?

clutchski: @choochootrain it's actually for linting the applications in `py/examples`. I could omit them from that and remove any mention of it in imports. 🤷‍♂️

choochootrain (approved):

PR #618, opened 2025-03-29
never use clickhouse for internal_btql

j13huang:


ankrgyl (approved):

js/src/logger.ts
```diff
@@ -3776,6 +3776,7 @@ class ObjectFetcher<RecordType>
                 cursor,
                 limit: INTERNAL_BTQL_LIMIT,
               },
+              use_columnstore: false,
```
ankrgyl, line 3779: If you want to future proof this properly, you can also add

```
brainstore_realtime: true
```

ankrgyl (approved):

PR #621, opened 2025-03-30
Allow propagating a 'strict' argument in the JS SDK to fail invalid prompts

ankrgyl:


core/js/src/object_util.ts
```diff
@@ -164,3 +164,21 @@ export function getRecordKeys<T extends Record<string | number | symbol, any>>(
...
+ * @example getObjValueByPath({ bar: { foo: "Hello, world!" } }, ["bar", "foo"]); // "Hello, world!"
+ */
+export function getObjValueByPath(
+  row: Record<string, unknown>,
+  path: string[],
+): unknown {
+  let curr: unknown = row;
+  for (const p of path) {
+    if (!isObjectOrArray(curr)) {
+      return null;
```
choochootrain, line 179: nit: probably want to return `undefined` to disambiguate from a path with a value of `null`

js/src/prompt.test.ts
```diff
@@ -0,0 +1,133 @@
+import { describe, test } from "vitest";
+import { Prompt } from "./logger";
+import { PromptData } from "@braintrust/core/typespecs";
+
+describe("prompt strict mode", () => {
+  test("strict mode", () => {
+    for (const strict of [true, false]) {
+      for (const shouldFail of [true, false]) {
+        for (const testNull of [true, false]) {
+          testPromptBuild({
```
choochootrain, line 10: should we add a nested object and array path test in the template as well?

ankrgyl:
Not yet.

PR #627, opened 2025-04-01
ignore span order in lanchain-py tests

choochootrain:


sachinpad (approved):

choochootrain:
add a nox test for the package, we can expand it if other updates change behavior

PR #634, opened 2025-04-04
Remove fallback to legacy `/logs` endpoint when logging

manugoyal:
Data planes have supported the `/logs3` endpoint for long-enough that we don't expect users of new SDKs to rely on this fallback. The fallback essentially doubles the number of retries against the data plane from 3 to 6, so this will also bring that down to 3. We can consider re-bumping that if necessary.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,4 +1,8 @@
 # Any changes to deps / build configuration need additional scrutiny.
+#
+# For example, adding a required dependency to core or SDK will make it required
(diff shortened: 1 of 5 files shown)


ankrgyl (approved):

CODEOWNERS
```diff
@@ -1,4 +1,8 @@
 # Any changes to deps / build configuration need additional scrutiny.
+#
```
manugoyal, line 2: Unrelated but thought it would help to add some detail here.

PR #638, opened 2025-04-07
include full error message in `FailedHTTPResponse` string representation

ankrgyl:


js/src/logger.ts
```diff
@@ -617,7 +617,7 @@ export class FailedHTTPResponse extends Error {
   public data: any;
 
   constructor(status: number, text: string, data: any = null) {
```
clutchski, line 619: nit: this is only called in one spot, maybe drop the default value

ankrgyl: good call

clutchski (approved):

PR #645, opened 2025-04-08
Fix positional arg in scorer

ankrgyl:
We were incorrectly treating `*args` as a named argument, when we really should be ignoring it.


py/src/braintrust/framework.py
```diff
@@ -385,8 +385,13 @@ def _call_user_fn_args(fn, kwargs):
     positional_args = []
     final_kwargs = {}
 
+    print("call user fn args", kwargs)
```
clutchski, line 388: remove the two print statements

clutchski (commented):
Remove the print statements but otherwise fine.

PR #647, opened 2025-04-09
process openai details tokens

clutchski:
This PR ensures we collect and store the details tokens noted in:

https://platform.openai.com/docs/api-reference/responses/object#responses/object-usage

We were storing them before in typescript, but this PR makes the naming consistent.

### DIFF
--- a/js/src/wrappers/oai_responses.ts
+++ b/js/src/wrappers/oai_responses.ts
@@ -157,6 +157,17 @@ function responsesStreamProxy(target: any): (params: any) => Promise<any> {
   });
 }
 
+const TOKEN_NAME_MAP = [
+  ["input_tokens", "prompt_tokens"],
+  ["output_tokens", "completion_tokens"],
+  ["total_tokens", "tokens"],
+];
+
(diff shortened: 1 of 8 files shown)


js/src/wrappers/oai_responses.ts
```diff
@@ -170,29 +181,24 @@ function parseMetricsFromUsage(usage: any): Record<string, number> {
...
 
-  const details: string[] = ["input", "output"];
-  for (const src of details) {
-    const details = usage[`${src}_tokens_details`];
-    if (details) {
-      for (const [key, value] of Object.entries(details)) {
-        const metricName = `${src}_${key}` as string;
-        if (typeof value === "number") {
-          metrics[metricName] = value;
+  for (const [oi_name, value] of Object.entries(usage)) {
```
sachinpad, line 186: oai_name?

sachinpad (approved):
lgtm!

js/src/wrappers/oai_responses.ts
clutchski: fixed

PR #648, opened 2025-04-09
change the max window to include zero

ibolmo:


core/py/src/braintrust_core/version.py
```diff
@@ -1 +1 @@
-VERSION = "0.0.58"
```
ibolmo, line 1: not 100% sure this is the right spot

PR #657, opened 2025-04-16
Fix async calls to anthropic.messages.create

clutchski:
These were failing.

### DIFF
--- a/py/src/braintrust/wrappers/test_anthropic.py
+++ b/py/src/braintrust/wrappers/test_anthropic.py
@@ -254,6 +254,8 @@ def test_anthropic_messages_stream_errors(memory_logger):
             raise Exception("fake-error")
     except Exception:
         pass
+    else:
+        raise Exception("should have raised an exception")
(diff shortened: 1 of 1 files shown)


py/src/braintrust/wrappers/test_anthropic.py
```diff
@@ -224,6 +243,25 @@ def test_anthropic_client_error(memory_logger):
...
+    assert not memory_logger.pop()
+
+    client = wrap_anthropic(_get_client())
+    msg_in = {"role": "user", "content": "what is 2+2? (just the number)"}
+
+    try:
+        with client.messages.stream(model=MODEL, max_tokens=300, messages=[msg_in]) as stream:
+            raise Exception("fake-error")
+    except Exception:
+        pass
```
choochootrain, line 256: maybe we should also test that the exception `fake-error` gets raised out of the sdk back to the app?

choochootrain (approved):

PR #662, opened 2025-04-18
Propagate state to ReadonlyAttachment

ankrgyl:


js/src/logger.ts
```diff
@@ -3990,7 +3993,7 @@ export class Experiment
     lazyMetadata: LazyValue<ProjectExperimentMetadata>,
     dataset?: AnyDataset,
   ) {
-    super("experiment", undefined, enrichAttachments);
+    super("experiment", undefined, (r) => enrichAttachments(r, state));
```
ibolmo, line 3996: #nit would love a more descriptive variable name. oh record?

ibolmo (approved):
Admittedly, I'm missing the context (connecting the dots) why with the state the UI just miraculously worked. Perhaps I'm not seeing a separate PR that got it to work, though

PR #664, opened 2025-04-18
Update anthropic_wrapper to log output when invoked via async def create(stream=True)

rmann-nflx:
# About
This changes updates the Braintrust SDK's anthropic_wrapper to log output when `async def client.messages.create(stream=True)`. Before this change the output was dropped completely and only the input was logged. 

This change is "abstractly" sketchy overall as it's a weird domain. We basically have the same problem (logging from a stream) when used across two interfaces:
```
stream_manager = anthropic_client.messages.stream() # This method is not async
async with stream_manager as stream:
    async for msg in stream:
        ...

stream = await anthropic_client.messages.create(stream=True) # This method is async 
async for msg in stream:
    ...
```

The subtle difference is that the stream lifecycle is managed with the context manager when using #stream() and is managed via the iterator lifecycle when using `create(stream=True)`.

This change is "concretely" sketchy because it's using a private function from Anthropic SDK `from anthropic.lib.streaming._messages import accumulate_event` and it's logging on every delta update -- which might pollute Braintrust.

# Testing
- Manually. Tested locally by adding braintrust SDK as an external dep to my existing package which uses the SDK and then updating the code and verifying the logs in the Braintrust UI. Things seem to work. The code was structured to try to leave everything except for the `create(stream=True)` case unchanged so I don't expect anything to be broken.
- I haven't tired to build it at all using the Braintrust build tools. No idea if style check works

### DIFF
--- a/py/src/braintrust/wrappers/anthropic.py
+++ b/py/src/braintrust/wrappers/anthropic.py
@@ -3,7 +3,8 @@ import logging
 from contextlib import contextmanager
 from typing import Any
 
-from braintrust import span_types
+from anthropic.lib.streaming._messages import accumulate_event  # TODO: Let's figure out something better...
+
 from braintrust.logger import NOOP_SPAN, log_exc_info_to_span, start_span
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


py/src/braintrust/wrappers/anthropic.py
```diff
@@ -46,11 +47,24 @@ def __init__(self, messages):
         self.__messages = messages
 
     async def create(self, *args, **kwargs):
+        # Anthropic SDK: -> Message | AsyncStream[RawMessageStreamEvent]
```
rmann-nflx, line 50: Code would be a lot simpiler if the Anthropic SDK returned `Message | AsyncMessageStreamManager` to provide return type symmetry between `.create(stream=True, ...)` and `.stream(...)`

clutchski:
Closed in favor of #665

PR #669, opened 2025-04-21
Solidify prompt cache.

clutchski:
- hash keys for safe filesystem access
- never throw errors into the app because of filesystem errors.


ankrgyl (commented):
Do we need to make the same change in typescript too?

ankrgyl (approved):

PR #676, opened 2025-04-24
Handle json schema response format

ankrgyl:
* Add support for pydantic `response_format`
* Add typespecs for `text` response format (in responses api)

### DIFF
--- a/core/js/typespecs/prompt.ts
+++ b/core/js/typespecs/prompt.ts
@@ -95,6 +95,30 @@ export const responseFormatSchema = z.union([
   z.object({ type: z.literal("text") }).openapi({ title: "text" }),
 ]);
 
+export const responsesAPIJsonSchemaSchema = z.object({
+  type: z.literal("json_schema"),
+  name: z.string(),
+  description: z.string().optional(),
+  schema: z
+    .union([
+      z.record(customTypes.unknown).openapi({ title: "object" }),
+      z.string().openapi({ title: "string" }),
+    ])
+    .optional(),
+  strict: z.boolean().nullish(),
+});
+export type ResponsesAPIJsonSchema = z.infer<
+  typeof responsesAPIJsonSchemaSchema
+>;
+
+export const responsesAPIFormatSchema = z.union([
+  z
+    .object({ type: z.literal("json_object") })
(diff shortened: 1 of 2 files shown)


py/src/braintrust/oai.py
```diff
@@ -746,3 +746,28 @@ def _parse_metrics_from_usage(usage: Dict[str, Any]) -> Dict[str, Any]:
 
 def _is_numeric(v):
     return isinstance(v, (int, float, complex))
+
+
+def prettify_params(params: Dict[str, Any]) -> Dict[str, Any]:
+    ret = {**params}
+    if "response_format" in ret:
+        ret["response_format"] = serialize_response_format(ret["response_format"])
```
ibolmo, line 754: my understanding here is that the `response_format` used to have a pydantic class/model but that wasn't serializable when that param was logged. 

should we attempt to recurse through the objects passed to the arguments try to serialize automatically? i.e. if we see a BaseModel subclass we could `model_json_schema` as default.

I suppose it may be wasteful and may hide problems until someone notices vs. this way it's performant and directed?

ibolmo (approved):

core/js/typespecs/prompt.ts
```diff
@@ -95,6 +95,30 @@ export const responseFormatSchema = z.union([
...
+      z.string().openapi({ title: "string" }),
+    ])
+    .optional(),
+  strict: z.boolean().nullish(),
+});
+export type ResponsesAPIJsonSchema = z.infer<
+  typeof responsesAPIJsonSchemaSchema
+>;
+
+export const responsesAPIFormatSchema = z.union([
```
ibolmo, line 114: when do you decide to add this type to the core typespecs vs. creating a spec where it is used?

py/src/braintrust/oai.py
ankrgyl: We could but I'd rather start with something targeted

core/js/typespecs/prompt.ts
ankrgyl: I have a feeling this is going to get reused in more places. But I wanted it to be next to the other response format definition.

PR #677, opened 2025-04-29
warn on experiment summary failures instead of raising out of Eval()

choochootrain:


js/src/logger.ts
```diff
@@ -4184,17 +4184,25 @@ export class Experiment
...
+          },
+          3,
+        );
 
-      scores = results["scores"];
-      metrics = results["metrics"];
+        scores = results["scores"];
+        metrics = results["metrics"];
+      } catch {
+        console.warn(
```
ankrgyl, line 4200: can you print the error we got from the network, for debugging purposes?

Also, coming soon, we're going to have `x-bt-internal-trace-id` on all requests. I dont want to overload this change, but in the near future, maybe we shoudl refactor all such errors to include that reference.

ankrgyl (approved):

PR #681, opened 2025-05-02
use an async context manager to conform with openai interface

ibolmo:


ibolmo:
looks like @clutchski has a great start here: https://github.com/braintrustdata/braintrust-sdk/pull/682/files

PR #683, opened 2025-05-04
WIP: Remote eval in python

ankrgyl:


mise.toml
```diff
@@ -7,3 +7,6 @@ _.python.venv = { path = "venv", create = true, uv_create_args = ['--seed']}
 
 # See env.example to configure API keys.
 _.file = ".env"
+
+[tools]
+python = "3.11.10"
```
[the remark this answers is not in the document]
clutchski: i think this conflicts with this? https://github.com/braintrustdata/braintrust-sdk/blob/main/.tool-versions#L2

py/src/braintrust/dev/server.py
```diff
@@ -0,0 +1,165 @@
+from dataclasses import asdict
+from typing import Annotated, Any, List, Optional, Union, cast
+
+import uvicorn
+from fastapi import FastAPI
```
clutchski, line 5: this might be a good use case for bottle. we could do this with no extra dependencies (other than perhaps cors?) and even check a copy into the repo. not mandatory, just food for thought. 

https://github.com/bottlepy/bottle

py/src/braintrust/logger.py
```diff
@@ -1219,16 +1288,17 @@ def init_dataset(
     :param use_output: (Deprecated) If True, records will be fetched from this dataset in the legacy format, with the "expected" field renamed to "output". This option will be removed in a future version of Braintrust.
     :returns: The dataset object.
     """
+    state_obj = state or _state
 
     def compute_metadata():
-        login(org_name=org_name, api_key=api_key, app_url=app_url)
+        state_obj.login(FullLoginOptions(org_name=org_name, api_key=api_key, app_url=app_url, force_login=force_login))
```
ibolmo, line 1294: may just to actual kwargs

py/src/braintrust/dev/server.py
ankrgyl: fastapi requires pydantic. pydantic is an untenable dependency because of the v1 / v2 incompatability. therefore, it's probably not a good idea to depend on fastapi (maybe if used as an optional dependency, but probably not)

PR #685, opened 2025-05-07
anthropic.ts: trace messages.beta.create|stream

clutchski:
wrap `anthropic.beta.messages` requests. It has the same api, so we can use the same mechanism as `anthropic.messages` (for now). Added a bunch of tests also. 

This addresses #BRA-2387.

### DIFF
--- a/js/src/wrappers/anthropic.test.ts
+++ b/js/src/wrappers/anthropic.test.ts
@@ -266,7 +266,7 @@ describe("anthropic client unit tests", () => {
       }),
     );
 
-    // validate that the wrapped and unwrapped clients product the same response
+    // validate that the wrapped and unwrapped clients produce the same response
     assertAnthropicResponsesEqual(responses);
 
     // validate we traced the second request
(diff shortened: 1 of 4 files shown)


js/src/wrappers/anthropic.test.ts
```diff
@@ -232,4 +243,254 @@ describe("anthropic client unit tests", () => {
...
   });
+
+  test("test client.beta.messages.create", async () => {
+    let startTime: number = -1;
+    let endTime: number = -1;
+
+    // wrapped has to be last so the timing test works
+    const clients = [anthropic, client];
+    const responses = await Promise.all(
+      clients.map(async (c) => {
```
ibolmo, line 254: #nit c => client ... let the tsc do this for us 😂

js/src/wrappers/anthropic.test.ts
```diff
@@ -232,4 +243,254 @@ describe("anthropic client unit tests", () => {
...
+      }),
+    );
+
+    // validate that the wrapped and unwrapped clients produce the same response
+    assertAnthropicResponsesEqual(responses);
+
+    // validate we traced the second request
+    const spans = await backgroundLogger.drain();
+    expect(spans).toHaveLength(1);
+    const span = spans[0] as any;
```
ibolmo, line 275: I believe we could have many of these. You could try something like:

https://github.com/braintrustdata/braintrust-sdk/blob/89a733ea8deaa0ac4a1bfb99c65e9603ff55f735/integrations/langchain-js/src/test/utils.ts#L35

to convert individual logs to spans (essentially a merge) to then do a proper check against the shape of the spans (i.e. root, child, subchild and so on)

js/src/wrappers/anthropic.test.ts
```diff
@@ -232,4 +243,254 @@ describe("anthropic client unit tests", () => {
...
+          model: TEST_MODEL,
+          messages: [{ role: "user", content: "What's 4*4?" }],
+          max_tokens: 100,
+          system: "Return the result only.",
+          temperature: 0.1,
+        });
+
+        let ttft = 0;
+        let outputText = "";
+        for await (const event of stream) {
```
ibolmo, line 411: you could collect all of the chunks and could do a chunk by chunk comparison (encouraged to use the expect.any() you've been doing). maybe too detailed, though.

would it make easier for future us, though, to see how the shape changes (i.e. if i add a new attribute like reasoning)

ibolmo (approved):

js/src/wrappers/anthropic.test.ts
ibolmo: see example: https://github.com/braintrustdata/braintrust-sdk/blob/89a733ea8deaa0ac4a1bfb99c65e9603ff55f735/integrations/langchain-js/src/BraintrustCallbackHandler.test.ts#L66

js/src/wrappers/anthropic.test.ts
clutchski: I like this idea

js/src/wrappers/anthropic.test.ts
clutchski: Ok, I think this is a good idea, but I'm going to leave it for a future PR because I think this test covers a lot of it. Next time I'll tackle this.

PR #692, opened 2025-05-13
Remove dependency on `braintrust_core`

clutchski:
This simplifies our deployment and ensures our users can't end up in dependency well between `braintrust_core`, `braintrust` and `autoevals`. 

- copy `serialized dataclass` since it's a util
- duck-type of `Score` and `Scorer` objects rather than checking `isinstance`
- add tests for both classes and runs them with and without braintrust core and autoevals installed.

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -31,7 +31,7 @@ from typing import (
 )
 
 import exceptiongroup
-from braintrust_core.score import Score, Scorer
+from braintrust_core.score import Scorer
 from tqdm.asyncio import tqdm as async_tqdm
 from tqdm.auto import tqdm as std_tqdm
 from typing_extensions import NotRequired, TypedDict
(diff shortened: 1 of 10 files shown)


py/src/braintrust/framework.py
```diff
@@ -179,9 +179,12 @@ class EvalScorerArgs(SerializableDataClass, Generic[Input, Output]):
 
 OneOrMoreScores = Union[float, int, bool, None, Score, List[Score]]
 
+# Type representing a scorer-like object with eval_async method
+ScorerLike = Any  # Any object with appropriate eval methods
```
clutchski, line 183: This perhaps could be a python protocol, but these are just type hints, so I think this is ok.

manugoyal: I think if it's part of a public type signature it might be good to make this more specific, because we're essentially weakening the public interface to accept anything as a scorer.

manugoyal (commented):
looks great to me! Just one question about the `ScorerLike` type signature. I suppose we should probably test this with our internal test suite too just in case it catches any inconsistencies between the different definitions of Scorer.

manugoyal (approved):
looks great!

PR #695, opened 2025-05-15
Matt/script py release

clutchski:
testing scripted releases


py/scripts/validate-release-tag.sh
```diff
@@ -0,0 +1,59 @@
...
+    exit 1
+  fi
+
+  TAG_COMMIT=$(git rev-parse "$TAG")
+  MAIN_CONTAINS=$(git branch --contains $TAG_COMMIT | grep -c "main" || true)
+
+  if [ "$MAIN_CONTAINS" -eq "0" ]; then
+    echo "ERROR: Tag $TAG is not on the main branch"
+    # FIXME[matt] remove this once we're onmain
+    # exit 1
```
choochootrain, line 51: im assuming you'll do this in a followup PR?

py/src/braintrust/test_version.py
```diff
@@ -0,0 +1,8 @@
+from braintrust import version
+
+
+def test_version():
+    # Not the most interesting test, but it's a quick sanity check
+    # that the templating during the build process works.
+    assert version.VERSION
+    assert version.GIT_COMMIT
```
choochootrain, line 8: could we test that these aren't equal to the pre-templating values if we're running from a wheel?

py/scripts/template-version.sh
```diff
@@ -0,0 +1,28 @@
...
+
+VERSION_FILE="src/braintrust/version.py"
+
+GIT_COMMIT=$(git rev-parse HEAD)
+
+sed_inplace() {
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+        sed -i '' "$@"
+    else
+        sed -i "$@"
```
choochootrain, line 13: gnu sed is best sed
image

Makefile
```diff
@@ -34,7 +34,7 @@ develop: ${VENV_PRE_COMMIT}
 fixup:
 	source env.sh && pre-commit run --all-files
 
-.PHONY: test test-py test-js nox pylint
+.PHONY: test test-py test-js nox pylint template-version template-version-rc
```
choochootrain, line 37: rm? looks like these were moved to the py Makefile

py/scripts/push-release-tag.sh
```diff
@@ -0,0 +1,66 @@
...
+if [ "$DRY_RUN" = true ]; then
+  exit 0
+fi
+
+echo ""
+echo ""
+echo "Are you ready to release version ${VERSION}? Type 'YOLO' to continue:"
+read -r CONFIRMATION
+
+if [ "$CONFIRMATION" != "yolo" ]; then
```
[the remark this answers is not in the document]
choochootrain: i think diamond is correct here, i vote for all caps YOLO

choochootrain (approved):
thank you for your service 🫡 

left some nits but nothing blocking

py/scripts/validate-release-tag.sh
clutchski: yes

py/src/braintrust/test_version.py
clutchski: Good catch. I actually wasn't testing the wheel. And it only took my 5 hours to figure it out.

PR #697, opened 2025-05-17
Unify / correctly capture token counts from anthropic models + cache breakpoints

ankrgyl:
* Allow setting cache breakpoints in the (unified) message format
* Fix token accounting from anthropic
* Remove anthropic import


js/src/wrappers/anthropic.ts
```diff
@@ -1,13 +1,36 @@
-import Anthropic from "@anthropic-ai/sdk";
```
ankrgyl, line 1: @clutchski i removed this

js/src/wrappers/anthropic.ts
```diff
@@ -1,13 +1,36 @@
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
clutchski, line 1: If we want to enforce this rule, we should add eslint to the build and have it start failing.

js/src/wrappers/anthropic.ts
```diff
@@ -1,13 +1,36 @@
...
 import { Span, startSpan } from "../logger";
 import { SpanTypeAttribute } from "@braintrust/core";
 import { filterFrom, getCurrentUnixTimestamp } from "../util";
 
-export function wrapAnthropic(anthropic: Anthropic): Anthropic {
-  return anthropicProxy(anthropic);
+/**
+ * Wrap an `Anthropic` object (created with `new Anthropic(...)`) to add tracing. If Braintrust is
+ * not configured, this is a no-op
```
clutchski, line 8: I think this comment isn't correct. If it's not an anthropic client, it's a no-op.

js/src/wrappers/anthropic.ts
```diff
@@ -1,13 +1,36 @@
...
+ *
+ * Currently, this only supports the `v4` API.
+ *
+ * @param anthropic
+ * @returns The wrapped `Anthropic` object.
+ */
+export function wrapAnthropic<T extends object>(anthropic: T): T {
+  const anthropicUnknown: unknown = anthropic;
```
clutchski, line 16: nit: i feel like this section would be more readable with a shorter variable like e.g. `a` or `au`. The length of the method makes the short variable name fine.

js/src/wrappers/oai.ts
```diff
@@ -45,8 +47,20 @@ declare global {
  * @returns The wrapped `OpenAI` object.
  */
 export function wrapOpenAI<T extends object>(openai: T): T {
-  if ((openai as any)?.chat?.completions?.create) {
-    return wrapOpenAIv4(openai as any) as T;
+  const openaiUnknown: unknown = openai;
+  if (
+    openaiUnknown &&
```
clutchski, line 52: nit: same as above

js/src/wrappers/anthropic.ts
```diff
@@ -218,8 +252,7 @@ function streamNextProxy(stream: AsyncIterator<any>, sspan: StartedSpan) {
     }
 
     if (result.done) {
-      totals.tokens =
-        (totals["prompt_tokens"] || 0) + (totals["completion_tokens"] || 0);
+      totals = finalizeMetrics(totals);
```
clutchski, line 255: I know I wrote this but it's worth double checking that we always get a "done" result. In some other async tracing code I moved `span.end` up one level because you might not make if there's an error or something.

If so, we might want to do the summing each loop (would require keeping a bit more state) or push it up a level. I can help you look into this if so. Or .. you can say "this is as good as it was before I got here" and I can look it into it when you're done.

js/src/wrappers/oai.ts
```diff
@@ -343,7 +361,7 @@ function parseBaseParams<T extends Record<string, any>>(
 ): StartSpanArgs {
   const { span_info, ...params } = allParams;
   const { metadata: spanInfoMetadata, ...spanInfoRest } = span_info ?? {};
-  let ret: StartSpanArgs = {
+  const ret: StartSpanArgs = {
```
clutchski, line 364: i wonder if there's a lint rule for this also.

js/src/wrappers/anthropic.ts
```diff
@@ -311,20 +346,31 @@ function parseMetricsFromUsage(usage: any): MetricsOrUndefined {
...
 }
 
+function finalizeMetrics(metrics: Metrics): Metrics {
+  const prompt_tokens =
+    (metrics.prompt_tokens || 0) +
+    (metrics.prompt_cached_tokens || 0) +
+    (metrics.prompt_cache_creation_tokens || 0);
+  return {
+    ...metrics,
+    // Anthropic's `input_tokens` does not include cache creation or cached read tokens.
```
clutchski, line 362: almost certainly need to fix in python too.

core/js/typespecs/openai/messages.ts
```diff
@@ -7,18 +7,27 @@ export const messageRoleSchema = z
...
+
 export const chatCompletionContentPartTextSchema = z
   .object({
     text: z.string().default(""),
     type: z.literal("text"),
+    cache_control: cacheControlSchema.optional(),
   })
   .openapi("ChatCompletionContentPartText");
 
+const chatCompletionSystemMessageParamSchema = z.object({
```
clutchski, line 22: do you think it's worth me doing this for anthropic also?

js/src/wrappers/anthropic.test.ts
```diff
@@ -232,14 +232,14 @@ describe("anthropic client unit tests", () => {
...
-    const ccit = "cache_creation_input_tokens";
-    const crit = "cache_read_input_tokens";
     expect(metrics).toBeDefined();
     expect(metrics["prompt_tokens"]).toBe(usage.input_tokens);
     expect(metrics["completion_tokens"]).toBe(usage.output_tokens);
     expect(metrics["tokens"]).toBe(usage.input_tokens + usage.output_tokens);
-    expect(metrics[crit]).toBe(usage[crit]);
-    expect(metrics[ccit]).toBe(usage[ccit]);
+    expect(metrics.prompt_cache_creation_tokens).toBe(
+      usage.cache_creation_input_tokens,
```
clutchski, line 240: nit: I didn't know about it when I originally wrote this, but I think the tests are much more readable/concise when we use `expect.toMatchObject` 

https://github.com/braintrustdata/braintrust-sdk/blob/main/js/src/wrappers/anthropic.test.ts#L443-L471

You can probably ask cursor to change them for you.

clutchski (approved):
I left a few comments, nothing blocking.

js/src/wrappers/anthropic.ts
ankrgyl: Yeah let me tweak this and the OpenAI one too.

js/src/wrappers/anthropic.ts
ankrgyl: happy to!

js/src/wrappers/oai.ts
ankrgyl: done

js/src/wrappers/anthropic.ts
ankrgyl: I think we should investigate this separately and probably write some tests that prove whether or not it would fail. I'll let you take this!

js/src/wrappers/oai.ts
ankrgyl: there is :) that's why my IDE asked to change it

core/js/typespecs/openai/messages.ts
ankrgyl: not yet since we don't parse anthropic inputs anywhere. we parse openai messages to translate _from_ them to various formats.

js/src/wrappers/anthropic.test.ts
ankrgyl: In this case we're comparing diff keys though right? Anyway feel free to take another pass and tweak!

ibolmo: yeah `toBe` and `toEqual` are more strict than `toMatchObject`. meaning if the `expected` keys are a subset then the extra keys in `actual` are ignored.

PR #706, opened 2025-05-21
feat: add ability to pass `Span` into AI SDK wrapper

Kalbi:
Previously the span ID was not accessible from the wrapper. Adding an optional `span` parameter to the wrapper here enables uses cases where we'd want to have a span ID already for feedback tracking. This way, the wrapper will use the existing span ID which can be started by the caller using something like `wrapTraced` or `logger.traced((span) => {...})`, or start a new one if not specified (retaining the current functionality) 

Our use case is to retrieve the specific Braintrust span ID that is generated internally by `wrapAISDKModel` when it instruments a call to an LLM provider (e.g., Gemini via @ai-sdk/google). We need this "inner" span ID to send back to our client for precise feedback tracking associated with that specific AI operation.

### DIFF
--- a/js/src/wrappers/ai-sdk.ts
+++ b/js/src/wrappers/ai-sdk.ts
@@ -1,6 +1,5 @@
 import { ContentPart, Message, Tools } from "@braintrust/core/typespecs";
-import { startSpan } from "../logger";
-import type { Span } from "../logger";
+import { startSpan, type Span } from "../logger";
 import { getCurrentUnixTimestamp, isEmpty } from "../util";
 import {
   LanguageModelV1,
@@ -21,6 +20,10 @@ import {
   X_CACHED_HEADER,
 } from "./oai";
 
+export interface WrapAISDKModelOpts {
+  span?: Span;
+}
+
 /**
  * Wrap an ai-sdk model (created with `.chat()`, `.completion()`, etc.) to add tracing. If Braintrust is
  * not configured, this is a no-op
@@ -34,8 +37,9 @@ import {
  */
 export function wrapAISDKModel<T extends object>(
   model: T,
-  overrideSpan?: Span,
+  opts?: WrapAISDKModelOpts,
 ): T {
+  opts = opts ?? {};
   // eslint-disable-next-line @typescript-eslint/consistent-type-assertions, @typescript-eslint/no-explicit-any
   const m = model as any;
   if (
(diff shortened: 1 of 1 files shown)


js/src/wrappers/ai-sdk.ts
```diff
@@ -1,5 +1,6 @@
 import { ContentPart, Message, Tools } from "@braintrust/core/typespecs";
 import { startSpan } from "../logger";
+import type { Span } from "../logger";
```
ankrgyl, line 3: can you combine this with the import statement above?

js/src/wrappers/ai-sdk.ts
```diff
@@ -24,10 +25,17 @@ import {
...
+ * @param overrideSpan Optional. A Braintrust Span object to use for tracing this model's operations.
+ *                     If provided, this span will be used instead of creating a new one internally.
+ *                     The lifecycle of this span (e.g., calling .end()) should be managed by the caller
+ *                     if it's passed here.
  * @returns The wrapped object.
  */
-export function wrapAISDKModel<T extends object>(model: T): T {
+export function wrapAISDKModel<T extends object>(
+  model: T,
+  overrideSpan?: Span,
```
ankrgyl, line 37: would be nice to have an opts argument that is an object so we can add more options over time. So something like

```
export interface WrapAISDKModelOpts {
  span?: Span
}

export function wrapAISDKModel<T extends object>(model, opts?: WrapAISDKModelOpts) {
  opts = opts ?? {};
  ...
}

File: js/src/wrappers/ai-sdk.ts [hunk 3]
```diff
@@ -36,15 +44,18 @@ export function wrapAISDKModel<T extends object>(model: T): T {
...
+    ) as any as T;
   } else {
     console.warn("Unsupported AI SDK model. Not wrapping.");
     return model;
   }
 }
 
 class BraintrustLanguageModelWrapper implements LanguageModelV1 {
-  constructor(private model: LanguageModelV1) {}
+  constructor(private model: LanguageModelV1, private instanceSpan?: Span) {}
```
ankrgyl, line 58: same thing here (propagate this as the same `opts` object)

ankrgyl (commented):
This is amazing. Thanks so much for the contribution! Left a few comments inline

gingerhendrix:
any progress on this one? We have a similar use-case and wold love to see it merged

PR #717, opened 2025-05-27
Add anthropic message typespecs

colsondonohue:
For use in client LLM viewer. 

Initially, I unioned the anthropic schema into `chatCompletionMessageParamSchema`, but that caused changes that rippled everywhere (I want to constrain the scope of these changes to fix the customer feedback so I can go back to wrapping up prompt editor cleanup). 

Ideally, we'd support anthropic format messages in the SDK, but have that broader type be a superset of e.g. the message schema used by the playground, so that the playground can continue to speak openAI-ish braintrust spec and doesn't need a bunch of extra type coercion. 

I'm keeping these schemas here so that if/when we ultimately add that support, these types are already in the SDK rather than living on the client.

### DIFF
--- a/core/js/typespecs/anthropic/messages.ts
+++ b/core/js/typespecs/anthropic/messages.ts
@@ -224,7 +224,7 @@ export const anthropicContentPartSchema = z.union([
 
 // system isn't technically a role, per Anthropic's docs, but we have users making use of it,
 // so we'll support it for parsing purposes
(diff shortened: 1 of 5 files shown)


ankrgyl (approved):
it would be neat to do a type assertion with the anthropic sdk (eg in a test file)

colsondonohue:
closing in favour of https://github.com/braintrustdata/braintrust-proxy/pull/240

PR #724, opened 2025-05-29
New api

CLowbrow:
Second pass at updating this API. I still need to export the original type as some of the typescript code in the app uses it. 

It worked this time: https://github.com/braintrustdata/braintrust/pull/5946


CLowbrow:
I don't know why the  esbuild bump shows up here it's in main I just rebased main.

CLowbrow:
Should be better

danielericlee (approved):

PR #731, opened 2025-06-03
wip: synchronously create prompts and functions

ankrgyl:


js/src/framework2.ts
```diff
@@ -44,6 +62,51 @@ export class Project {
...
+      GenericFunction<any, any>
+    >,
+  ) {
+    this._publishableCodeFunctions.push(fn);
+    if (globalThis._lazy_load) {
+      globalThis._evals.functions.push(fn);
+    }
+  }
+
+  async publish() {
```
ibolmo, line 89: Do we need to clear up any of the `_publishableCodeFunctions` or `_publishablePrompts` on publish? 

I'm also wondering if we should avoid returning in this function. Technically we still have code functions pending, and it'll likely mean backward incompatible change, for example, if we decide to later return as well the created published code functions. 

I also wonder what this code will do if we have a `publish()` call within a file that also happens in a `push` (bundle) cli call.

js/src/framework2.ts
```diff
@@ -44,6 +62,51 @@ export class Project {
...
+    const functionDefinitions: FunctionEvent[] = [];
+    if (this._publishableCodeFunctions.length > 0) {
+      throw new Error(
+        "Code functions cannot be published directly. Use `braintrust push` instead.",
+      );
+    }
+    if (this._publishablePrompts.length > 0) {
+      for (const prompt of this._publishablePrompts) {
+        const functionDefinition =
+          await prompt.toFunctionDefinition(projectMap);
```
ibolmo, line 101: curious why we couldn't do the functionDefinition within `addPrompt` is it a race condition of the `ProjectNameIdMap` may not have the latest information?

js/src/framework2.ts
```diff
@@ -44,6 +62,51 @@ export class Project {
...
+      globalThis._evals.functions.push(fn);
+    }
+  }
+
+  async publish() {
+    await login();
+    const projectMap = new ProjectNameIdMap();
+    const functionDefinitions: FunctionEvent[] = [];
+    if (this._publishableCodeFunctions.length > 0) {
+      throw new Error(
```
ibolmo, line 94: I wonder if we should warn rather than throw an exception. I could see it annoying since it's all or nothing. i.e. if I have a .create() for functions and prompts, then eventually I hit .publish() nothing will be pushed. Which is safer, but I don't think we got anything out of this change. I would have to split up the file and avoid calling .create() for functions.

ibolmo (approved):
Looks good, but I left a few DX points to consider

js/src/framework2.ts
ankrgyl: sounds good to me

js/src/framework2.ts
ankrgyl: good call. i made it return if lazy loading is on (so it's a no-op)

js/src/framework2.ts
ankrgyl: just hygiene, so all of the network calling happens when you `publish()` rather than while defining.

PR #734, opened 2025-06-04
Add groupBy field to monitor view options schema

CLowbrow:
- Add groupBy as optional string field to monitorViewOptionsSchema
- Enables saving and loading groupBy selections with monitor views


CLowbrow:
This is for https://github.com/braintrustdata/braintrust/pull/6044

j13huang (approved):

PR #739, opened 2025-06-06
ensure traced works with async generators

clutchski:
fixes BRA-1842.

Note, i don't trace the outputs of async generators because the output it's potentially infinite or at least very large.


py/src/braintrust/test_logger.py
```diff
@@ -323,19 +330,6 @@ def test_noop_permalink_issue_1837():
     assert span.link() == "https://www.braintrust.dev/noop-span"
 
 
-def test_permalink_logged_out(with_memory_logger):
```
clutchski, line 326: this is kind of useless

ibolmo (approved):

PR #740, opened 2025-06-08
Fix parallel tool call logging

ankrgyl:


py/src/braintrust/oai.py
```diff
@@ -215,17 +215,21 @@ def _postprocess_streaming_results(cls, all_results: List[Dict[str, Any]]) -> Di
 
             if delta.get("content") is not None:
                 content = (content or "") + delta.get("content")
+
             if delta.get("tool_calls") is not None:
-                if tool_calls is None:
-                    tool_calls = [
+                tool_delta = delta["tool_calls"][0]
```
clutchski, line 220: not sure if this is a possible scenario, but this could fail if tool_calls is empty. maybe safer to do something like .. 

```
tool_calls = delta.get("tool_calls")
if tool_calls:
   tool_delta = tool_calls[0]
```
or just wrap in a try / catch and that's ok too.

clutchski (approved):
a few small comments but fine!

PR #752, opened 2025-06-24
Matt/dev server

clutchski:


js/dev/index.ts
```diff
@@ -1 +1,3 @@
+import { configureNode } from "../src/node";
+configureNode();
```
ankrgyl, line 2: very tiny nit, can you add a newline after this

ankrgyl (approved):

PR #753, opened 2025-06-24
Remove buffer from SpanIdentifierV3

ankrgyl:


core/js/src/span_identifier_v3.ts
```diff
@@ -12,18 +12,56 @@ import { InvokeFunctionRequest } from "typespecs";
...
       throw new Error();
     }
-    return { bytes: Buffer.from(ret), isUUID: true };
+    return { bytes: new Uint8Array(ret), isUUID: true };
   } catch {
     return { bytes: undefined, isUUID: false };
   }
 }
 
+function concatUint8Arrays(...arrays: Uint8Array[]): Uint8Array {
```
manugoyal, line 27: We may want to move these to their own `.ts` file so we can import them in other places that we use buffer, e.g. [here](https://github.com/braintrustdata/braintrust-sdk/blob/adfe5473bf50ce4f88c88826d4055b144ef42bb0/js/src/logger.ts#L1226-L1227) and [here](https://github.com/braintrustdata/braintrust-sdk/blob/adfe5473bf50ce4f88c88826d4055b144ef42bb0/js/src/functions/upload.ts#L303), though the latter seems very node-specific.

ankrgyl: refactored

PR #762, opened 2025-06-30
Fixes linting issue with @traced when used without parenthis

ohmeow:
See this issue here: https://discord.com/channels/1146930518722609238/1389049635515011082/1389049635515011082

If using `@traced` without parenthesis you get the linting error below.  This PR fixes it.

Screenshot 2025-06-30 at 1 09 27 PM

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -1623,6 +1623,16 @@ def _try_log_output(span, output):
 F = TypeVar("F", bound=Callable[..., Any])
 
 
+@overload
+def traced(f: F) -> F:
+    """Decorator to trace the wrapped function when used without parentheses."""
+
+
+@overload
+def traced(*span_args: Any, **span_kwargs: Any) -> Callable[[F], F]:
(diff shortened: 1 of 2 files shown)


py/src/braintrust/logger.py
```diff
@@ -1623,6 +1623,16 @@ def _try_log_output(span, output):
 F = TypeVar("F", bound=Callable[..., Any])
 
 
+@overload
```
ankrgyl, line 1626: lgtm! @manugoyal mind sanity checking?

ankrgyl (approved):

py/src/braintrust/logger.py
clutchski: I mentioned on slack, but a couple of unit tests like will serve as a living record of the contract.  https://github.com/braintrustdata/braintrust-sdk/blob/main/py/src/braintrust/test_logger.py#L596-L632 will serv

ohmeow: yah updated tests ... lmk if there is anything else.

PR #789, opened 2025-07-14
Matt/q

clutchski:


js/src/queue.ts
```diff
@@ -1,14 +1,26 @@
...
+// are using initLogger (aka observing production data) we enforce queue limits to ensure we
+// never OOM.
+let _useQueueSizeLimit = false;
+
+/**
+ * Enable or disable queue size limits globally for all queue instances.
+ * @param enabled - true to use instance maxSize limits, false for unlimited queues (default).
+ */
+export function setQueueSizeLimitEnabled(enabled: boolean) {
+  console.log("setQueueSizeLimitEnabled", enabled);
```
ankrgyl, line 15: rm?

js/src/logger.ts
```diff
@@ -2970,6 +2970,11 @@ export function initLogger<IsAsyncFlush extends boolean = true>(
     project_name: projectName,
     project_id: projectId,
   };
+
+  // When initLogger is called, we actually enable the queue size limit. We do this because
+  // initLogger is used for production obsservability and we don't want to OOM customer's apps.
+  setQueueSizeLimitEnabled(true);
```
ankrgyl, line 2976: it's kind of a bummer that we have to do this globally, rather than per object. I also don't think that this will work in cases where we use the SDK in a multi-tenant fashion (eg our SDK). In those cases, we create an independent `state` variable for each session, and each one has its own background logger. So maybe you could place this state on the background logger itself, and then run this on whatever the object's current `state` variable is?

clutchski: yes agreed. that's a better design.

ankrgyl (approved):

manugoyal (commented):
A few high-level questions:

- Do we want to do the same for python in this PR?
- I am a bit worried about the global boolean-flipping here, since initializing a logger and an experiment in the same process could cause problems. A couple ideas
  - We could use a separate background logger for project loggers vs everything else. Seems unlikely that anybody would care that those data streams are published to the backend independently.
  - Maybe we log a warning if the enforcement boolean is changed after being set the first time?

clutchski:
> A few high-level questions:
> 
> * Do we want to do the same for python in this PR?
> * I am a bit worried about the global boolean-flipping here, since initializing a logger and an experiment in the same process could cause problems. A couple ideas
>   
>   * We could use a separate background logger for project loggers vs everything else. Seems unlikely that anybody would care that those data streams are published to the backend independently.
>   * Maybe we log a warning if the enforcement boolean is changed after being set the first time?

yes and yes but we'll fix those in later PRs.

PR #811, opened 2025-07-24
remove unused imports and enforce with ruff

ibolmo:


py/src/braintrust/test_logger.py
```diff
@@ -15,10 +14,7 @@
     assert_dict_matches,
     assert_logged_out,
     init_test_logger,
-    simulate_login,
     simulate_logout,
-    with_memory_logger,
-    with_simulate_login,
```
ibolmo, line 21: grr

choochootrain: ?

ibolmo: these are pytest fixtures that should stick around

PR #821, opened 2025-07-30
Service token docs

choochootrain:


core/js/typespecs/app_types.ts
```diff
@@ -1219,6 +1239,41 @@ export const createApiKeyOutputSchema = apiKeySchema
...
+    .describe("The organization ID this service token should be scoped to."),
+  account_id: z
+    .string()
+    .describe("The service account ID this service token should belong to."),
+});
+
+export const deleteServiceTokenSchema = z
+  .object({
+    id: z.string().uuid().describe(`Unique identifier for the service token.`),
+    org_id: z
```
manugoyal, line 1258: curious why this is necessary. Shouldn't it be derivable from the id?

core/js/typespecs/app_types.ts
```diff
@@ -1219,6 +1239,41 @@ export const createApiKeyOutputSchema = apiKeySchema
   )
   .openapi("CreateApiKeyOutput");
 
+export const createServiceTokenSchema = z.object({
+  name: z
+    .string()
+    .describe("Name of the service token. Does not have to be unique"),
+  org_id: z
```
manugoyal, line 1246: most other APIs like `createApiSchema` ask for an `org_name` instead of an `org_id`. Should we do that here?

core/js/typespecs/app_types.ts
```diff
@@ -1219,6 +1239,41 @@ export const createApiKeyOutputSchema = apiKeySchema
...
+    .string()
+    .uuid()
+    .describe("The organization ID this service token should be scoped to."),
+  account_id: z
```
manugoyal, line 1250: What does this mean?

manugoyal (approved):

PR #822, opened 2025-07-30
[WIP] openai agents for js

CLowbrow:
The tests pass at least!

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -51,17 +51,19 @@
     "docs": "npx typedoc --options typedoc.json src/index.ts",
     "prepublishOnly": "../../scripts/node_prepublish_sdk.py",
     "postpublish": "../../scripts/node_postpublish_sdk.py",
-    "test": "vitest run --exclude src/wrappers/anthropic.test.ts --exclude src/wrappers/oai.test.ts --exclude src/otel.test.ts --exclude src/otel-no-deps.test.ts  --exclude src/wrappers/ai-sdk-v1.test.ts --exclude src/wrappers/ai-sdk-v2.test.ts",
+    "test": "vitest run --exclude src/wrappers/anthropic.test.ts --exclude src/wrappers/oai.test.ts --exclude src/otel.test.ts --exclude src/otel-no-deps.test.ts  --exclude src/wrappers/ai-sdk-v1.test.ts --exclude src/wrappers/ai-sdk-v2.test.ts --exclude src/wrappers/openai-agents.test.ts",
     "test:anthropic": "vitest run src/wrappers/anthropic.test.ts",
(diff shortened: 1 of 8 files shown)


js/src/wrappers/openai-agents.ts
```diff
@@ -0,0 +1,343 @@
...
+export class BraintrustTracingProcessor {
+  private logger?: Span | Experiment | Logger<any>;
+  private spans: Map<string, Span> = new Map();
+
+  constructor(logger?: Span | Experiment | Logger<any>) {
+    this.logger = logger;
+  }
+
+  onTraceStart(trace: AgentsTrace): Promise<void> {
+    // Check if we already have a span for this trace to avoid duplicates
```
clutchski, line 112: why would this happen?

js/src/wrappers/openai-agents.ts
```diff
@@ -0,0 +1,343 @@
...
+    if (this.spans.has(trace.traceId)) {
+      return Promise.resolve();
+    }
+
+    if (this.logger) {
```
clutchski, line 117: this stanza has a lot of duplicated code. Why not something like `

```
logger = this.logger || globalLogger
logger.startSpan(...)
```

I think we can find a way to do it.

js/src/wrappers/openai-agents.ts
```diff
@@ -0,0 +1,343 @@
...
+    toAgent?: string;
+    triggered?: boolean;
+    data?: Record<string, any>;
+  };
+  startedAt?: string;
+  endedAt?: string;
+  error?: string;
+}
+
+function spanTypeFromAgents(span: AgentsSpan): SpanTypeAttribute {
```
clutchski, line 51: one meta question, how does the data look in the app?

js/package.json
```diff
@@ -117,5 +119,10 @@
   },
   "peerDependencies": {
     "zod": "^3.25.34"
+  },
+  "peerDependenciesMeta": {
```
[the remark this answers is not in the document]
ibolmo: I'm actually thinking we should start doing the right thing and create a new separate project under the sdk/integrations folders and publish it independently of the SDK

js/src/exports-node.ts
```diff
@@ -28,6 +28,7 @@ export * from "./wrappers/oai";
 export { BraintrustMiddleware } from "./wrappers/ai-sdk-v2";
 export { wrapAISDKModel } from "./wrappers/ai-sdk-v1";
 export { wrapAnthropic } from "./wrappers/anthropic";
+export { BraintrustTracingProcessor } from "./wrappers/openai-agents";
```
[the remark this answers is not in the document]
ibolmo: moving this to a separate package will help with these kind of naming concerns

PR #825, opened 2025-08-01
wrap traced for generators

choochootrain:
Introduces a `BRAINTRUST_MAX_GENERATOR_ITEMS` env var (default 1000) which determines how many generator items we will buffer and log as `output`. If more items are generated, the buffer is cleared and no output is logged.

Supports python/js sync/async generators


js/src/logger.ts
```diff
@@ -3493,6 +3493,164 @@ export function traced<IsAsyncFlush extends boolean = true, R = void>(
   }
 }
 
+/**
+ * Check if a function is a sync generator function.
+ * @param fn The function to check.
+ * @returns True if the function is a sync generator function.
+ */
+function isGeneratorFunction(fn: any): boolean {
```
ibolmo, line 3501: May be better to define an actual generated method and to stringify that to get the actual expected string. Looks like a similar library does something similar: https://github.com/inspect-js/is-generator-function/blob/main/index.js#L25

ibolmo (approved):
looks solid!

js/src/logger.ts
choochootrain: added a comment here, we don't need the level of backwards compat that the linked implementation uses. added some tests to verify as well

PR #829, opened 2025-08-05
Add monitor filter options

joemccourt:
Adds filter options
 * `btqlFilters` for multiple btql filters
 * `searchFilter` for free form text input search


j13huang:
Hey, just as some context, we have the search data for the non-monitor view types in the view_data object, which gets saved to the data plane instead of the control plane for information privacy reasons. 

So I would imagine that we would want some similar analog on the monitor side as well

PR #841, opened 2025-08-06
Add a param to eval that makes the SDK not generate an experiment and log locally

CLowbrow:
Yes!

### DIFF
--- a/js/src/framework.test.ts
+++ b/js/src/framework.test.ts
@@ -13,6 +13,7 @@ import {
   EvalScorer,
   runEvaluator,
 } from "./framework";
+import { flush } from "./logger";
 import { configureNode } from "./node";
 import { BarProgressReporter, type ProgressReporter } from "./progress";
 import { InternalAbortError } from "./util";
@@ -511,6 +512,9 @@ test("trialIndex with multiple inputs", async () => {
 });
 
 test("Eval with noSendLogs: true runs locally without creating experiment", async () => {
+  // Spy on flush function to verify it's never called
+  const flushSpy = vi.spyOn({ flush }, "flush");
+
   const result = await Eval(
     "test-no-logs",
     {
@@ -547,4 +551,10 @@ test("Eval with noSendLogs: true runs locally without creating experiment", asyn
   expect(result.summary.experimentUrl).toBeUndefined();
   expect(result.summary.scores.exact_match.score).toBe(1);
   expect(result.summary.scores.simple_scorer.score).toBe(0.8);
+
+  // Most importantly: verify that flush was never called (no logs sent)
+  expect(flushSpy).not.toHaveBeenCalled();
+
+  // Clean up
+  flushSpy.mockRestore();
 });
--- a/js/src/framework.test.ts
+++ b/js/src/framework.test.ts
@@ -9,6 +9,7 @@ import {
 } from "vitest";
 import {
   defaultErrorScoreHandler,
+  Eval,
   EvalScorer,
   runEvaluator,
 } from "./framework";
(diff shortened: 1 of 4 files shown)


js/src/framework.ts
```diff
@@ -479,6 +479,11 @@ export interface EvalOptions<EvalReport, Parameters extends EvalParameters> {
    * A `Reporter` which you can use to summarize progress after an Eval() runs.
    */
   reporter?: ReporterDef<EvalReport> | string;
+  /**
+   * Whether to send logs to Braintrust. When false, the evaluation runs locally
+   * and builds a local summary instead of creating an experiment. Defaults to true.
+   */
+  sendLogs?: boolean;
```
ankrgyl, line 486: i'd prefer to make this `noSendLogs` because convenionally if `sendLogs` is undefined it feels like it'd default to false. Also that'll be consistent with the CLI flag

CLowbrow: Hey what are you doing in here? It says draft! 😄 (ok I'll change it)

py/src/braintrust/test_framework.py
```diff
@@ -237,3 +238,40 @@ def task_with_hooks(input_value: int, hooks: EvalHooks) -> int:
...
+    assert result.results[0].scores["exact_match"] == 1.0
+    assert result.results[0].scores["simple_scorer"] == 0.8
+
+    assert result.results[1].input == "test"
+    assert result.results[1].output == "test world"
+    assert result.results[1].scores["exact_match"] == 1.0
+    assert result.results[1].scores["simple_scorer"] == 0.8
+
+    # Verify it builds a local summary (no experiment_url means local run)
+    assert result.summary.project_name == "test-no-logs"
```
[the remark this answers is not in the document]
CLowbrow: I'm testing it with the assert result.summary.experiment_url is None line. Is there a better way?

the cli also just skips experiment creation when this is set: https://github.com/braintrustdata/braintrust-sdk/blob/b07d82b7fe426edc3ea57eca2e54dd1a7865e61d/py/src/braintrust/cli/eval.py#L125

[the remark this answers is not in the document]
CLowbrow: Ok I was able to make this a little more robust by checking we are never initializing an experiment that could log something.

js/src/framework.test.ts
```diff
@@ -508,3 +509,42 @@ test("trialIndex with multiple inputs", async () => {
...
+          score: args.output === args.expected ? 1 : 0,
+        }),
+        () => ({ name: "simple_scorer", score: 0.8 }),
+      ],
+    },
+    { noSendLogs: true },
+  );
+
+  // Verify it returns results
+  expect(result.results).toHaveLength(2);
```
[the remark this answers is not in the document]
CLowbrow: spying on flush function to make sure it can't be called.

js/src/framework.test.ts
```diff
@@ -547,4 +551,10 @@ test("Eval with noSendLogs: true runs locally without creating experiment", asyn
   expect(result.summary.experimentUrl).toBeUndefined();
   expect(result.summary.scores.exact_match.score).toBe(1);
   expect(result.summary.scores.simple_scorer.score).toBe(0.8);
+
+  // Most importantly: verify that flush was never called (no logs sent)
+  expect(flushSpy).not.toHaveBeenCalled();
```
ibolmo, line 556: sorry for being picky .. could we use https://github.com/braintrustdata/braintrust-sdk/blob/445cba50d36e46a64605ecbca35a76c94a528db3/js/src/logger.test.ts#L21 

then 

```
  await memoryLogger.flush();
  expect(await memoryLogger.drain()).length(0);
```

py/src/braintrust/test_framework.py
```diff
@@ -250,28 +251,33 @@ def exact_match(input, output, expected):
...
-    assert result.results[1].scores["simple_scorer"] == 0.8
-
-    # Verify it builds a local summary (no experiment_url means local run)
-    assert result.summary.project_name == "test-no-logs"
-    assert result.summary.experiment_url is None
-    assert result.summary.scores["exact_match"].score == 1.0
-    assert result.summary.scores["simple_scorer"].score == 0.8
+    # Spy on the logger's init function to verify no experiment is created
+    with patch("braintrust.framework.init_experiment") as mock_init:
+        result = await Eval(
```
ibolmo, line 256: for python we can use something like:

https://github.com/braintrustdata/braintrust-sdk/blob/445cba50d36e46a64605ecbca35a76c94a528db3/py/src/braintrust/test_logger.py#L981-L984

PR #848, opened 2025-08-11
Python Tracing Optimizations

clutchski:
This PR is a pass on the performance of creating spans (it doesn't include queuing/flushing etc), just the things that block user code.

## Summary

Performance Comparison Results:

  Original Version (commit 9a212ee34):

  - Mean time: 1.320s ± 0.019s
  - Range: 1.300s - 1.361s

  Optimized Version (HEAD with all optimizations):

  - Mean time: 612.6ms ± 43.3ms
  - Range: 582.0ms - 718.3ms

  - 53.6% faster (707ms reduction)
  - More than 2x speedup (2.15x faster)
  - Consistent performance improvement across the entire range

  Summary of All Optimizations Applied:

  1. Single-pass data processing - Eliminated multiple iterations
  2. Removed validation overhead - No forbidden key checks, type validation (we shouldn't be throwing errors into userland)
  3. Removed many calls to isinstance calls
  4. In-place mutations - Deep copy once, then mutate in place vs many iterations and many copies
  5. Eliminated JSON serialization check - Removed upfront bt_dumps validation (but unknown types are checked in deep copy)
  6. Removed unused functions - Deleted split_logging_data, _strip_nones
  7. Streamlined data flow - Direct processing without intermediate structures

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -3488,6 +3488,9 @@ class SpanImpl(Span):
                 if v is not None:
                     if isinstance(v, BraintrustStream):
                         lazy_values[k] = LazyValue(lambda stream=v: stream.copy().final_value(), use_mutex=False)
+                    elif k == "span_attributes" and isinstance(v, dict):
+                        # Filter None values from span_attributes
+                        final_record[k] = {sk: sv for sk, sv in v.items() if sv is not None}
                     else:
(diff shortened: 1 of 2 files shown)


AbhiPrasad:
with braintrust-sdk-python now being a thing, going to close this PR. We can re-apply this patch to that repo there.

PR #853, opened 2025-08-12
Use typespecs generated from openapi

manugoyal:
This PR is part of an effort to move ownership of the typespecs in `@braintrust/core/typespecs` into our internal repo, and expose them to our public repos through an openapi spec. Generally this makes sense because most typespecs reflect the APIs provided by our internal services, so they ought to be owned by the internal services.

Aside from the new auto-generated typespecs, most of the changes to existing code are just modifying imports to point to the auto-generated typespecs. And most changes to the typespecs code are just explicitly defining names for schemas that we use in either the JS or PY code, so that they get picked up by the openAPI generator.

The main trickiness involves handling zod constructs which transform data from an input type to an output type, e.g. `.default`, `.transform`, `.preprocess`, etc. Such schemas have a different "input" type than their "output" type, and our current openapi spec generator seems to generate based on input types. So in certain cases, we tack on some properties to the openapi spec to instruct our generator to post-process the spec in certain ways, so that it aligns with the output type. In the future, we can probably refactor our zod -> openapi pipeline to use an output-type-sensitive generator.

### DIFF
--- a/js/tsconfig.json
+++ b/js/tsconfig.json
@@ -10,5 +10,5 @@
     "skipLibCheck": true
   },
   "include": ["."],
-  "exclude": ["node_modules/**", "**/dist/**", "examples/**", "**/*.test.ts"]
+  "exclude": ["node_modules/**", "**/dist/**", "examples/**", "scripts/**", "**/*.test.ts"]
 }


core/js/typespecs/api_types.ts
```diff
@@ -391,6 +391,10 @@ export const datasetEventSchema = z
   .openapi("DatasetEvent");
 export type DatasetEvent = z.infer<typeof datasetEventSchema>;
 
+const functionTypeEnumNullish = functionTypeEnum
```
manugoyal, line 394: Just to preface, none of the schema modifications are actually changing the "logical meaning" of the schema. They are just adding metadata / other workarounds for the sake of better conversion to an openAPI spec. So from a review perspective, just spot-checking that the schema changes are all behavior-preserving should be sufficient.

py/scripts/generate_imported_types.py
```diff
@@ -0,0 +1,108 @@
...
+        check=True,
+    )
+    cleanup_internal_types()
+
+
+def cleanup_internal_types():
+    with open(INTERNAL_TYPES_OUTPUT_PATH, "r") as f:
+        contents = f.read()
+
+    # Replace `NotRequired[...]` with `NotRequired[Optional[...]]` for Python
```
manugoyal, line 54: These cleanups were copied over from our internal script to auto-generate python typeddicts from an openapi spec.

core/js/typespecs/openai/messages.ts
```diff
@@ -22,26 +26,33 @@ export const cacheControlSchema = z.object({
 
 export const chatCompletionContentPartTextSchema = z
   .object({
-    text: z.string().default(""),
+    text: z
+      .string()
+      .default("")
+      .openapi({ [X_OPENAPI_OUTPUT_MARK_REQUIRED]: true }),
```
ibolmo, line 32: I guess this is necessary because at some point we decided that the `text` could be an empty string, which differs from the actual interface from OpenAI? 

Image

core/js/typespecs/openai/messages.ts
```diff
@@ -149,19 +176,25 @@ const chatCompletionAssistantMessageParamSchema = z.object({
     .nullish(),
   // NOTE: It's important to keep these optional, rather than nullish, to stay
   // inline with the OpenAI SDK's type definition.
-  function_call: functionCallSchema.nullish().transform((x) => x ?? undefined),
+  function_call: functionCallSchema
+    .nullish()
```
ibolmo, line 180: and this one because we included `null` as an option but the actual spec has only `undefined`?

Image

ibolmo: I suppose in the future we'll avoid messing with OpenAI types. In the future we'l have a BraintrustChatCompletion schema and that one we can define ourselves how we want it to work. 

If we truly need to refer to the OpenAI spec it self we can refer to its types and/or json schema.

core/js/typespecs/app_types.ts
```diff
@@ -303,6 +303,15 @@ export const playgroundLogsLogIdLiteralSchema = z
   .literal("x")
   .describe("A literal 'x' which identifies the object as a playground log");
 
+// Defined as a separate openapi schema from promptDataSchema, so that the
+// nullish doesn't bleed into the openapi schema for the raw PromptData type.
+const promptDataNullishSchema = promptDataSchema
+  .nullish()
+  .openapi("PromptDataNullish");
+const functionTypeEnumNullish = functionTypeEnum
```
ibolmo, line 311: #nit maybe worth sharing the schema

https://github.com/braintrustdata/braintrust-sdk/pull/853/files#diff-6b08eabf71a46b843175831585e3581b9516702233780ef00f5529cf79b4ba96R394

core/js/typespecs/common_types.ts
```diff
@@ -44,6 +44,15 @@ export const datetimeStringSchema = z
   })
   .openapi({ format: "date-time" });
 
+// Our internal zod -> OpenAPI generator currently generates openapi specs which
```
ibolmo, line 47: 👍🏻

js/src/imported_types.ts
```diff
@@ -0,0 +1,1453 @@
+// Auto-generated file -- do not modify
```
ibolmo, line 1: #nit "imported_types" may be better as "generated_types" and have the `// Auto-generated file` maybe point to the sha/version of the api/version and/or tool that created this

js/scripts/generate_imported_types.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
ibolmo, line 1: I was expecting this to live some where 🤔 I had sdk would just get a simple `ts` file with all the generated stuff updated and we commit the change.

js/src/wrappers/ai-sdk-v1.ts
```diff
@@ -1,4 +1,11 @@
-import { ContentPart, Message, Tools } from "@braintrust/core/typespecs";
+import {
```
ibolmo, line 1: this reminds me that (but lower prio.) the AI proxy also depends on the core/js:

https://github.com/braintrustdata/braintrust/blob/4fb59f67c1d4078ae2bd35e5e6a038832ca6c468/api-ts/src/proxy/proxy.ts#L29

py/src/braintrust/imported_types.py
```diff
@@ -0,0 +1,185 @@
+"""Re-exports generated typespecs that are considered public API."""
```
ibolmo, line 1: I'd be ok if this was `types.py` and we had a `_generated_types.py`

manugoyal: Yeah I wanted to avoid using the name `types.py` because it felt too general. I'm fine with renaming everything to `generated_types`.

js/scripts/generate_imported_types.sh
manugoyal: I wanted to leave the language-specific SDKs responsible for generating types in the way they see fit, and the internal repo to only be responsible for producing an openAPI spec.

core/js/typespecs/app_types.ts
manugoyal: not sure I follow?

core/js/typespecs/openai/messages.ts
manugoyal: yeah I like that direction.

PR #857, opened 2025-08-13
Can fetch prompt versions

CLowbrow:
WIP. Need to test this in integration tests in the main repo.


js/src/logger.ts
```diff
@@ -6169,6 +6170,79 @@ function simulateLogoutForTests() {
...
 }
 
+/**
+ * Get the versions for a prompt.
+ *
+ * @param projectId The ID of the project to query
+ * @param promptId The ID of the prompt to get versions for
+ * @returns Promise containing the version data
+ */
+export async function getPromptVersions(
```
clutchski, line 6180: i'd like to put a moratorium on adding stuff to this file. it's already 6000 lines long (too long for AI to process quickly) and hard to deal with. can we find a better home for this?

py/src/braintrust/logger.py
```diff
@@ -4616,6 +4609,69 @@ def wrapped_fn(*args, **kwargs):
...
+    Args:
+        project_id: The ID of the project to query
+        prompt_id: The ID of the prompt to get versions for
+
+    Returns:
+        List of transaction IDs (_xact_id) for entries where audit_data.action is "upsert"
+    """
+    state = _internal_get_global_state()
+    if not state or not state.logged_in:
+        raise Exception("Must log in first")
```
clutchski, line 4625: i feel like we shouldn't have random "not logged in" methods sprinkled through the code. i feel like this should be done in one way (e.g. `_verify_logged_in()`) or even better ... getting an error that says missing `BRAINTRUST_API_KEY` or something more prescriptive. is there a pattern for this?

CLowbrow: Everything else just calls _state directly so I'm going to do that instead. Probably safe?

clutchski (approved):
looks great

PR #860, opened 2025-08-14
align ruff, .pre-commit-config.yaml and pyproject.toml with monorepo

choochootrain:


pyproject.toml
```diff
@@ -4,10 +4,12 @@ line-length = 119
 [tool.ruff]
 line-length = 119
 
+[tool.ruff.isort]
+known-third-party = ["braintrust_core", "braintrust_local"]
+
 [tool.ruff.lint]
 select = [
     "I",    # isort
-    "F401", # unused imports
```
ibolmo, line 10: this one we can likely keep (been around for a month or so) but no prob

PR #862, opened 2025-08-15
Make the AI SDK v5 wrapper calculate anthropic tokens in the way we want

CLowbrow:
This is the TS side implementation

## Before

image

## After 

image

## OpenAI before

image

## OpenAI after (works the same)

image


manugoyal (approved):
codeowners approval for `package.json`

js/src/wrappers/test-prompts.ts
```diff
@@ -0,0 +1,144 @@
+// Very long system prompt to trigger Anthropic prompt caching (1500+ tokens)
```
ibolmo, line 1: #nit i usually do something like `ai-sdk-v2.fixtures.ts` to point out that it's not an interesting file plus which source code depends on it. not a standard thing, though

js/src/wrappers/ai-sdk-v2.test.ts
```diff
@@ -423,4 +430,124 @@ describe("ai sdk middleware tests", TEST_SUITE_OPTIONS, () => {
...
+        ],
+        maxRetries: 0,
+      });
+
+      const aiSdkSpans = await testLogger.drain();
+      expect(aiSdkSpans).toHaveLength(1);
+      const aiSdkSpan = aiSdkSpans[0] as any;
+
+      // Third call: Direct wrapper with cache hit for comparison
+      await new Promise((resolve) => setTimeout(resolve, 500));
```
ibolmo, line 507: why did we need to sleep?

CLowbrow: This test sucks because we have to wait for anthropic cache. Even with the sleep it sometimes fails for the first time but the tests are configured to try again a couple times and eventually the cache is there. I can take out the sleep but that will possibly make the test even flakier? I'm not sure what the best thing to do here is.

CLowbrow: Oh wait there are two sleeps we don't need this second one it's not going to fix anything. my bad I'll delete it.

PR #864, opened 2025-08-15
ensure vercel spans are valid spans for any otel version

ibolmo:
This is a quick hack to try to help unblock people. 

Proper solution is to move our otel to a separate package and have that package depend on the right version that vercel works out of the box.


js/src/otel.ts
```diff
@@ -305,11 +305,41 @@ export class BraintrustSpanProcessor {
...
+              ]);
+            };
+          }
+
+          // For any other property, pass the access through to the original object.
+          return Reflect.get(target, prop, receiver);
+        },
+      });
     } catch (error) {
+      console.error(error);
```
ibolmo, line 342: #nit we can make this better, but I tripped over this where any problem within this code will cause the same "make sure ... is installed" was was _very_ misleading

ankrgyl (approved):

PR #866, opened 2025-08-18
Improve OpenAI Responses input parsing 

danielericlee:
Currently our TypeScript OpenAI Responses wrapper assumes that `input` will be a string. However, `input` to the `openai.responses.create` invocation may also be [an array of various shapes](https://platform.openai.com/docs/api-reference/responses/create#responses_create-input).

When the input is an array, let's pass the array through in its original form. When the input is a string, we'll continue to transform it into `[{ role: "user", content: params.input }]`.

This PR also augments our existing OpenAI responses example to include a call to `create` that passes an array as input and adds a test that invokes `client.responses.create` with an input array.


danielericlee:
> @danielericlee I think with our move to pass input/output directly we can close this one?

Indeed. Closing.

PR #871, opened 2025-08-19
Add support for tags when using eval hooks

knjiang:
adding support for tags allows filtering eval logs by an array of strings

ref: https://www.braintrust.dev/docs/reference/libs/python#evalhooks-objects

### DIFF
--- a/js/src/framework.test.ts
+++ b/js/src/framework.test.ts
@@ -556,7 +556,7 @@ test("Eval with noSendLogs: true runs locally without creating experiment", asyn
   expect(await memoryLogger.drain()).toHaveLength(0);
 });
 
-test("hooks.tags can be appended and logged to root span", async () => {
+test("tags can be appended and logged to root span", async () => {
   await _exportsForTestingOnly.simulateLoginForTests();
   const memoryLogger = _exportsForTestingOnly.useTestBackgroundLogger();
   const experiment =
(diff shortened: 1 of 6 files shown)


py/src/braintrust/test_helpers.py
```diff
@@ -72,6 +72,31 @@ def init_test_logger(project_name: str):
...
+
+    project_metadata = ObjectMetadata(id=project_name, name=project_name, full_info=dict())
+    experiment_metadata = ObjectMetadata(id=experiment_name, name=experiment_name, full_info=dict())
+    metadata = ProjectExperimentMetadata(project=project_metadata, experiment=experiment_metadata)
+    lazy_metadata = LazyValue(lambda: metadata, use_mutex=False)
+
+    exp = braintrust.init(project=project_name, experiment=experiment_name)
+    exp._lazy_metadata = lazy_metadata  # Skip actual login by setting fake metadata directly
+    return exp
+
```
knjiang, line 99: this function was required for python setup, took it from pr -> https://github.com/braintrustdata/braintrust-sdk/pull/863/files

py/src/braintrust/framework.py
```diff
@@ -1047,6 +1064,14 @@ def span(self) -> Optional[Span]:
     def set_span(self, span: Optional[Span]):
         self._span = span
 
+    @property
+    def tags(self) -> Sequence[str]:
+        return self["tags"]
+
+    @tags.setter
+    def tags(self, tags: Sequence[str]) -> None:
```
choochootrain, line 1072: if the setter only allows sequences when would you need to check `tags is not None` above? or is that specifically to handle the case where there are no tags on the dataset?

py/src/braintrust/test_framework.py
```diff
@@ -280,3 +282,106 @@ def simple_scorer(input, output, expected):
...
+    """ Test that hooks.tags can be set to a list. """
+
+    def task_with_hooks(input, hooks):
+        hooks.tags = tags
+        return input
+
+    evaluator = Evaluator(
+        project_name=__name__,
+        eval_name=__name__,
+        data=[EvalCase(input="hello", expected="hello world")],
```
choochootrain, line 335: should we add a case for when `EvalCase` does not have initial tags AND the task does not touch `hook.tags`? that would express the behavior in my comment above and we can verify it is the same as it was before introducing the tags hook

py/src/braintrust/test_framework.py
```diff
@@ -280,3 +282,106 @@ def simple_scorer(input, output, expected):
...
+    logs = with_memory_logger.pop()
+    assert len(logs) == 3
+
+    # assert root span contains tags
+    root_span = [log for log in logs if not log["span_parents"]]
+    assert len(root_span) == 1
+    assert root_span[0].get("tags") == expected_tags
+
+@pytest.mark.asyncio
+async def test_hooks_tags_with_failing_scorer(with_memory_logger, with_simulate_login, simple_scorer):
```
choochootrain, line 355: do we need this test? the hook logic is the same regardless of the scorers we pass

choochootrain: oh i see in the js tests what you are doing

js/src/framework.test.ts
```diff
@@ -555,3 +555,124 @@ test("Eval with noSendLogs: true runs locally without creating experiment", asyn
...
+  expect(result.results[0].tags).toEqual(expectedTags);
+
+  await memoryLogger.flush();
+  const logs = await memoryLogger.drain();
+  const rootSpans = logs.filter((l: any) => !l["span_parents"]);
+  expect(rootSpans).toHaveLength(1);
+  expect((rootSpans[0] as any).tags).toEqual(expectedTags);
+});
+
+test("tags are persisted with a failing scorer", async () => {
```
choochootrain, line 642: similar to the comment on the python tests - should we add one more test where no initial tags are set on the eval case and we dont do anything with the tags hook?
that should behave the same with and without this change so it's a good regression test

choochootrain (commented):
just some minor things and some test ideas but looks good overall!

py/src/braintrust/framework.py
knjiang: added an extra test case with invalid types for python since the types are just hints.

PR #872, opened 2025-08-19
Support custom masking functions

ankrgyl:


py/src/braintrust/test_logger.py
```diff
@@ -1129,3 +1129,212 @@ async def unlimited_buffer_async_generator():
         os.environ.pop("BRAINTRUST_MAX_GENERATOR_ITEMS", None)
         if original:
             os.environ["BRAINTRUST_MAX_GENERATOR_ITEMS"] = original
+
+
+def test_masking_function_logger(with_memory_logger, with_simulate_login):
+    """Test that masking function is applied to logged data in Logger."""
+
+    def masking_function(data):
```
clutchski, line 1137: i would consider doing in-place replacement rather than a full-copy of the data. in my perf PRs, i'm gradually replacing all of the loops/copies with a single pass, single copy approach. any one of them doesn't make much difference, but all of them added up does.

ankrgyl: This is just a test -- you can do whatever you want in `masking_function` and it is allowed to modify the data inline!

There are certain python types that cannot really be modified inline (like strings) so i think it's important from an interface standpoint that you can return the value.

Also just to be clear, we run masking _after_ a copy. Which is important, because we don't want masking to edit the application's data.

ibolmo (commented):
looks solid but we may want to consider the error handling

js/src/logger.ts
```diff
@@ -92,6 +92,34 @@ import {
...
+  maskingFunction: (value: unknown) => unknown,
+  data: unknown,
+  fieldName: string,
+): unknown {
+  try {
+    return maskingFunction(data);
+  } catch (error) {
+    // Get the stack trace
+    const stack = error instanceof Error ? error.stack : String(error);
+    return `ERROR: Failed to mask data: ${stack}`;
```
ibolmo, line 119: #nit I wonder if the error may include PII.

stupid example but one could imagine they might pass the secret to other functions which would then be part of the stack trace in some way:

Image

safer yet, but I'll defer to you if we just do this or just return the simple message:

Image

ibolmo (approved):

PR #873, opened 2025-08-19
allow non-batch export in braintrust processor

ibolmo:
easier testing since we avoid thread workers


clutchski (approved):
fix the comment though

PR #874, opened 2025-08-19
upgrade typedoc

choochootrain:


ibolmo:
@choochootrain still needed?

choochootrain:
yep ill get around to it

PR #878, opened 2025-08-21
Fix OpenAI wrapping to return a proxied APIPromise

ibolmo:


ibolmo:
ran into some expect test failures, but can't tell if that's the changes here or my local. movign to draft for safety

PR #882, opened 2025-08-22
Add `@braintrust` package as a peer dependency to integration libraries

knjiang:


ibolmo (commented):
Nice!

I recently added a direct dependency in langchain-js for v0.0.7. Two questions:

Does it make sense to have a direct dependency for since the library could depend on specific version functionality. I.e. we fixed someting in 0.1.2 so we want to require at least that version. Is the peer dependency too lenient to catch that?

PR #885, opened 2025-08-25
[Vercel OTEL] Support OTEL v1 style spans

knjiang:


js/src/otel.ts
```diff
@@ -142,7 +143,7 @@ export class AISpanProcessor {
     }
 
     // Always keep root spans (no parent)
-    if (!span.parentSpanContext) {
+    if (!span.parentSpanContext && !span.parentSpanId) {
```
ankrgyl, line 146: might be useful to leave a small comment about how v2 has `parentSpanContext` and v1 has `parentSpanId`

PR #888, opened 2025-08-26
Fixes current_span().start_span not working when nested inside agent tools for oai. 

CLowbrow:
## After Change

image

## Before Change

image
image

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -254,6 +254,11 @@ class Span(Exportable, contextlib.AbstractContextManager, ABC):
         """Set the span as the current span. This is used to mark the span as the active span for the current thread."""
         pass
 
+    @abstractmethod
+    def unset_current(self) -> None:
+        """Unset the span as the current span."""
+        pass
+
 
 class _NoopSpan(Span):
     """A fake implementation of the Span API which does nothing. This can be used as the default span."""
@@ -310,6 +315,9 @@ class _NoopSpan(Span):
         pass
 
     def set_current(self):
+        pass
+
+    def unset_current(self):
         pass
 
     def __enter__(self):


py/src/braintrust/wrappers/openai.py
```diff
@@ -212,18 +213,29 @@ def on_span_start(self, span: tracing.Span[tracing.SpanData]) -> None:
...
-        self._spans[span.span_id] = parent.start_span(
+        created_span = parent.start_span(
             id=span.span_id,
             name=_span_name(span),
             type=_span_type(span),
             start_time=_timestamp_from_maybe_iso(span.started_at),
         )
+        self._spans[span.span_id] = created_span
+
+        # Enter the span context to make it active for current_span() calls
```
manugoyal, line 224: I think we have a `Span.set_current()` method that does this without needing to bring in the context manager.

py/src/braintrust/wrappers/openai.py
```diff
@@ -212,12 +212,16 @@ def on_span_start(self, span: tracing.Span[tracing.SpanData]) -> None:
...
+        created_span = parent.start_span(
             id=span.span_id,
             name=_span_name(span),
             type=_span_type(span),
             start_time=_timestamp_from_maybe_iso(span.started_at),
         )
+        self._spans[span.span_id] = created_span
+
+        # Set the span as current so current_span() calls will return it
+        created_span.set_current()
```
manugoyal, line 224: Cool. But now I'm realizing we may not have an analogous `unset_current` method that resets the context token created by `set_current`. And that unwinding only happens in the `__exit__` method. Maybe we should add that to the `Span` interface/implementation and use it here too?

py/src/braintrust/logger.py
```diff
@@ -3810,6 +3810,10 @@ def set_current(self):
         if self.can_set_current:
             self._context_token = self.state.current_span.set(self)
 
+    def unset_current(self):
```
manugoyal, line 3813: I think we need to add this to the `Span` abstract class and all implementations. Not just `SpanImpl`.

manugoyal (approved):
lgtm. Just need to add `unset_current` to the span abstract class and other implementations.

PR #889, opened 2025-08-27
Improve support for zod v4

patricksmith:
This change updates the imports for zod from `import { z } from "zod";` to `import { z } from "zod/v3";` to improve support for zod v4 in the SDK.

### The problem
My application is using zod v4 (`^4.0.15` specifically) which causes issues when building a prompt via the Braintrust SDK. In my code, I load a prompt and then attempt to build it:

```ts
const prompt = await loadPrompt({ projectId: "my-project-id", slug: "my-prompt-slug" });
const built = prompt.build({ some: "params" });
```

This results in an error on the `prompt.build` line:

```
TypeError: Cannot read properties of undefined (reading '_zod')
    at t._zod.parse (/my-code/index.js:3671:6503)
    at /my-code/index.js:3654:655
    at t.safeParse (/my-code/index.js:3673:4749)
    at Fen.runBuild (/my-code/index.js:3763:12868)
    at Fen.build (/my-code/index.js:3763:11985)
```

### The solution

Starting with zod v3.25.0 (which is already used in most places throughout the SDK), the zod package includes both v3 and v4 of the package available by either import from `"zod/v3"` or `"zod/v4/core"`. Importing directly from `"zod"` will import v3 when a v3.* version is installed and v4 when a v4.* version is installed. Since the v3 and v4 types are incompatible, this is causing issues when a user of the SDK (me) has v4 installed but the SDK is expecting v3.

To fix this, I've updated all imports in the JS SDK to import from `"zod/v3"` instead of `"zod"` directly, [as suggested by the zod docs](https://zod.dev/library-authors?id=how-to-support-zod-4).

### Testing

I tried to test this change locally by pointing my application to a local version of the repo (updating `package.json` to use `"braintrust": "/path/to/code/braintrust-sdk/js"`), but that caused some other issues when building since the `core` package couldn't be found.

Instead, I manually updated the SDK in my `node_modules` (specifically `node_modules/braintrust/dist/index.mjs`) to import from "zod/v3" everywhere instead of "zod" and verified that calling `prompt.build()` in my application code succeeded 🎉 .

### DIFF
--- a/core/js/src/generated_types.ts
+++ b/core/js/src/generated_types.ts
@@ -1,6 +1,6 @@
 // Auto-generated file (internal git SHA 10d02c4e14f09ae288e6bd5af3cebe98cf1dbf83) -- do not modify
 
-import { z } from "zod";
+import { z } from "zod/v3";
 
 export const AclObjectType = z.union([
   z.enum([
--- a/core/js/src/object.ts
+++ b/core/js/src/object.ts
@@ -1,4 +1,4 @@
-import { z } from "zod";
+import { z } from "zod/v3";
 import {
   AsyncScoringControlType as AsyncScoringControl,
   ObjectReference as objectReferenceSchema,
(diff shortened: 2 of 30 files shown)


patricksmith:
@manugoyal Thanks for taking a look! I just pushed the lockfile updates

PR #891, opened 2025-08-28
Switch responses wrapper for js to not do transforms (similar to the agents wrapper)

CLowbrow:
This looks worse until we clean it up in the UI but is strictly better due to not hiding any data. 

## Before
image

## After
image


ankrgyl (approved):
code lgtm

PR #897, opened 2025-08-30
Fix parent precedence ordering

knjiang:


js/src/logger.ts
```diff
@@ -4045,7 +4045,12 @@ function startSpanAndIsLogger<IsAsyncFlush extends boolean = true>(
 ): { span: Span; isSyncFlushLogger: boolean } {
   const state = args?.state ?? _globalState;
 
-  const parentStr = args?.parent ?? state.currentParent.getStore();
+  // Precedence: current span > current parent > logger
```
ankrgyl, line 4048: @manugoyal can you sanity check this reasoning

manugoyal: Would it make sense to implement this whole block of logic in `getSpanParentObject`? Currently that thing returns a `Span` object from which you call `startSpan`, but it could also return a `SpanComponentsV3` object from which you initialize the span directly.

I feel like `getSpanParentObject` ought to consider the parent string, and spreading out the selection between this function and that seems confusing.

That being said, I do think the precedence ordering is sane.

core/js/src/span_identifier_v3.ts
```diff
@@ -239,6 +239,10 @@ export class SpanComponentsV3 {
     }
   }
 
+  public async export(): Promise<string> {
+    return this.toStr();
+  }
+
```
knjiang, line 245: the export methods are required since there are other callers of `getParentObject`

PR #898, opened 2025-08-30
Add strict arg to Prompt.build

dpguthrie:
Adding a `strict` argument to the Prompt.build method.  When `True` it will throw an error for missing variables.

This functionality exists in the typescript sdk:

- https://github.com/braintrustdata/braintrust-sdk/blob/3d5ec4de453ab8ea873a6c8c0d0ef0b2110bd386/js/src/logger.ts#L6012
- https://github.com/braintrustdata/braintrust-sdk/blob/main/js/src/mustache-utils.ts

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -9,6 +9,7 @@ import inspect
 import json
 import logging
 import os
+import re
 import sys
 import textwrap
 import threading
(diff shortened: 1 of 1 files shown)


dpguthrie:
@ankrgyl Added the following:

Helper Function Tests (6 tests)
  - Mustache variable extraction (_extract_mustache_variables)
  - Nested value access (_get_nested_value)
  - Template validation (lint_template)

Integration Tests (7 tests)
  - Strict mode enabled: Validates missing variables throw ValueError
  with descriptive messages
  - Strict mode disabled: Ensures backward compatibility - missing
  variables render as empty strings, no exceptions
  - Nested objects: Tests dot notation like user.profile.email
  - Array access: Validates array bounds checking for items.0, items.1,
  etc.
  - Edge cases: Empty arrays, partial variables, completely missing
  variables

Key Behaviors Verified
  - strict=True: Missing variables → ValueError("Variable 'name' does 
  not exist.")
  - strict=False (default): Missing variables → empty strings, always
  generates valid prompt
  - Backward compatibility: Existing code continues to work unchanged
  - Error messages use original variable names (not normalized array
  indices)

py/src/braintrust/logger.py
```diff
@@ -4291,6 +4297,27 @@ def render_prompt_params(params: Dict[str, Any], args: Any) -> Dict[str, Any]:
...
+def render_mustache(template: str, data: Any, *, strict: bool = False, renderer: Optional[Callable[..., Any]] = None):
+    if renderer is None:
+        renderer = chevron.render
+
+    if not strict:
+        return renderer(template, data=data)
+
+    # Capture stderr to check for missing keys
+    stderr_capture = io.StringIO()
+    with contextlib.redirect_stderr(stderr_capture):
```
ibolmo, line 4309: opted with this to defer as much as we can to the actual template rendering library

PR #905, opened 2025-09-05
properly handle underlying openai rejections in wrapOpenAI

choochootrain:
lazily evaluate the wrapped execution so rejections from openai propagate in the expected place.

added tests for invalid `OPENAI_API_KEY` which causes the underlying rejection in the repro submitted.

### DIFF
--- a/js/src/wrappers/oai.ts
+++ b/js/src/wrappers/oai.ts
@@ -313,10 +313,13 @@ function wrapChatCompletion<
   return (allParams: P & SpanInfo, options?: unknown): APIPromise<C> => {
     const { span_info: _, ...params } = allParams;
 
-    // Cache the execution promise to ensure we only execute once
+    // Lazy execution - we must defer the API call until the promise is actually consumed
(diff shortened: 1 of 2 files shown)


js/src/wrappers/oai.ts
```diff
@@ -381,28 +384,31 @@ function wrapChatCompletion<
...
+        // Handle Promise methods - trigger lazy execution and forward to data promise
+        if (
+          prop === "then" ||
+          prop === "catch" ||
+          prop === "finally" ||
+          prop in Promise.prototype
+        ) {
+          // Create data promise if needed (cache it for efficiency)
+          if (!dataPromise) {
+            dataPromise = ensureExecuted().then((result) => result.data);
```
CLowbrow, line 405: Actually should we still have a catch here? for safety?

choochootrain: discussed offline, we dont want to handle this but instead propagate to the caller

PR #927, opened 2025-09-15
Zod v4

ankrgyl:


js/src/generated_types.ts
```diff
@@ -517,7 +517,7 @@ export const ExperimentEvent = z.object({
   expected: z.unknown().optional(),
   error: z.unknown().optional(),
   scores: z
-    .union([z.record(z.union([z.number(), z.null()])), z.null()])
+    .union([z.record(z.string(), z.union([z.number(), z.null()])), z.null()])
```
manugoyal, line 520: We shouldn't be manually editing this file, since it's auto-generated. If we need to make adjustments to the generated types, we'll have to fix the [generator script](https://github.com/braintrustdata/braintrust-sdk/blob/b8c2d1ebc7b75ea604e9490fd265a6f797add415/js/scripts/generate_types.ts#L1).

PR #930, opened 2025-09-16
Monkey/wrapper based ADK

ibolmo:
The existing ADK integration integrated with OpenTelemetry. Customers asked to see the input/output of the spans. On closer inspection, the Google ADK's use of OpenTelemetry lacked passing the input/output to the traces/spans. We could make a contribution upstream, but based on the other feedback we should have more control over the integration than rely on Google's ADK to keep up with our customer's demand. 

Thus a rewrite to use monkey patching and/or wrapping.


integrations/adk-py/examples/hello_eval.py
```diff
@@ -0,0 +1,22 @@
...
+from braintrust_adk import setup_braintrust
+from hello import main as hello
+
+
+async def main():
+    async def task(input: Any):
+        setup_braintrust()
+        return await hello(input)
+
+    await EvalAsync(
```
ibolmo, line 14: verifies that evals + integration continues to work as expected:

Image

integrations/adk-py/examples/pyproject.toml
```diff
@@ -6,7 +6,7 @@ readme = "README.md"
 requires-python = ">=3.9"
 dependencies = [
     "braintrust-adk",
-    "google-adk>=1.9.0",
+    "google-adk>=1.14.1",
```
ibolmo, line 9: required due to https://github.com/google/adk-python/pull/2383

integrations/adk-py/examples/multi_tool_agent/agent.py
```diff
@@ -56,7 +61,9 @@ def get_current_time(city: str) -> Dict[str, str]:
     return {"status": "success", "report": report}
 
 
-root_agent = LlmAgent(
+init_logger(project="adk-multi-tool")
+
+root_agent = wrap_agent(LlmAgent)(
```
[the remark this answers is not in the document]
choochootrain: kinda weird to have a braintrust example that doesn't log llm calls by default. should we update?

integrations/adk-py/examples/hello_eval.py
```diff
@@ -0,0 +1,22 @@
...
+from typing import Any
+
+from braintrust.framework import EvalAsync
+from braintrust_adk import setup_braintrust
+from hello import main as hello
+
+
+async def main():
+    async def task(input: Any):
+        setup_braintrust()
```
choochootrain, line 11: is this supposed to happen in the task or did you mean to put it at the top level?

integrations/adk-py/src/braintrust_adk/__init__.py
```diff
@@ -13,35 +17,321 @@ def setup_braintrust(
...
+        return wrapper
+
+    Flow.run_async = trace_flow(Flow.run_async)
+
+    def trace_call_llm(wrapped: Any):
+        @wraps(wrapped)
+        async def wrapper(instance: Any, *args: Any, **kwargs: Any):
+            invocation_context = args[0] if len(args) > 0 else kwargs.get("invocation_context")
+            llm_request = args[1] if len(args) > 1 else kwargs.get("llm_request")
+            model_response_event = args[2] if len(args) > 2 else kwargs.get("model_response_event")
```
choochootrain, line 132: do we need to look in both `args` and `kwargs` here and in `trace_run_wrapper`? below in `trace_run_sync_wrapper` and `trace_run_async_wrapper` we only look at `kwargs`. might be something about google adk that i am not familiar with

integrations/adk-py/src/braintrust_adk/__init__.py
```diff
@@ -13,35 +17,321 @@ def setup_braintrust(
...
+    return getattr(obj, "_braintrust_patched", False)
+
+
+def _try_dict(obj: Any) -> Optional[Union[Iterable[Any], Dict[str, Any]]]:
+    if hasattr(obj, "model_dump"):
+        try:
+            obj = obj.model_dump(exclude_none=True)
+        except ValueError as e:
+            if "Circular reference" in str(e):
+                return
```
choochootrain, line 310: why would we want to swallow this error? seems like a good thing to propagate up the call stack and maybe even surface as an error in the span

choochootrain (commented):
left some comments mostly for my understanding

integrations/adk-py/src/braintrust_adk/__init__.py
ibolmo: tl;dr the adk has the same object multiple times i.e. the agent is in multiple objects. i think we can just wait for a customer to complain if something is missing, but i bet we will not hear from them

integrations/adk-py/examples/hello_eval.py
ibolmo: it can go anywhere and multiple calls are noop. other parts explain top level, so i think we could leave it as is

integrations/adk-py/src/braintrust_adk/__init__.py
choochootrain: mind leaving a note about that? otherwise reading this i wouldn't know this is intentional and related to adk including the `agent` reference in here

PR #933, opened 2025-09-17
Add trace-level metadata logging to OpenAI Agents SDK integration

dpdiliberto:
When using the OpenAI Agents SDK, metadata wasn't flowing through Braintrust spans to the root-level span, which made it challenging to track conversation context and user sessions. This PR updates `onTraceEnd()` (JS) and `on_trace_end()` (Python) to include `trace.metadata` in root span logging. A user can implement this using the native OpenAI Agents trace APIs so that they'll appear in Braintrust. Going off the example in Braintrust docs (https://www.braintrust.dev/docs/integrations/openai-agents-sdk#openai-agents-sdk), this would look like the following:

**Typescript**

```
async function main() {
  const agent = new Agent({
    name: "Assistant",
    model: "gpt-4o-mini",
    instructions: "You only respond in haikus.",
  });

  await withTrace(
    { name: "Order Processing", metadata: { conversation_id: "conv-123", user_id: "user-456" } },
    async () => {
      const result = await run(agent, "Process my order");
    }
  );
}
```

**Python**

```
async def main():
    agent = Agent(
        name="Assistant",
        instructions="You only respond in haikus.",
    )
    with trace("Order Processing", metadata={"conversation_id": "conv-123", "user_id": "user-456"}):
        result = await Runner.run(agent, "Process my order")
```


CLowbrow (approved):
lgtm

CLowbrow (approved):

PR #935, opened 2025-09-17
Remove num-incompatible version spec

CLowbrow:


integrations/openai-agents-js/package.json
```diff
@@ -32,6 +32,6 @@
     "zod": "^3.25.34"
   },
   "dependencies": {
-    "braintrust": "workspace:^"
+    "braintrust": ">=0.3.7"
```
ibolmo, line 35: you should keep this as is but instead of using npm use `pnpm publish` it'll automatically resolve the workspace:^ to the right version

PR #940, opened 2025-09-20
try to detect and support esm bundling

### LINKED ISSUE #908: Error: Top-level await is currently not supported with the "cjs" output format
When I try to run something similar to the [Eval via SDK tutorial](https://www.braintrust.dev/docs/start/eval-sdk): 

`npx braintrust eval tutorial.eval.ts`

And in the eval task it runs code imported from another file with a top-level await, for example:

```
import { generateMarkdown } from '@/file-with-top-level-await-in-it';

Eval(...
    task: async (input) => await generateMarkdown(input),
)
```

I get the following error:

`ERROR: Top-level await is currently not supported with the "cjs" output format`

The package.json uses `"type": "module"` and tsconfig uses `"module": "esnext"` and `"moduleResolution": "bundler"` so it's unclear why it is trying to run in cjs mode instead of esm mode.

Any idea how to get this to run as esm with top-level await or otherwise fix the issue?  Thanks!

ibolmo:
fixes: #908

### DIFF
--- a/js/src/cli.ts
+++ b/js/src/cli.ts
@@ -93,11 +93,21 @@ export interface FileHandle {
 function evaluateBuildResults(
   inFile: string,
   buildResult: esbuild.BuildResult,
+  outFile?: string,
 ): EvaluatorFile | null {
   if (!buildResult.outputFiles) {
     return null;
   }
   const moduleText = buildResult.outputFiles[0].text;
+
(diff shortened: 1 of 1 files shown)


fiws:
Why close? Is there a alternative implementation in review or shipped?

PR #947, opened 2025-09-24
Agno Integration

danielericlee:
Braintrust integration for [Agno](https://docs.agno.com/introduction)

The strategy I took here was to minimally transform the payloads returned by Agno such that the Braintrust traces looked nice. Namely, my goal was to ensure that the LLM view worked for both inputs and outputs, the Thread view showed a nice summary of the trace, and the Try Prompt button was enabled for LLM spans.

Olmo and I will collaborate on this PR to get it to a place where it is production-ready.


manugoyal (approved):
CO for `py/setup.py`

PR #948, opened 2025-09-25
add metrics types/schemas

ibolmo:


clutchski (changes requested):
chatted with olmo in person, but

- we shouldn't add these to our public interface (e.g. braintrust.__init__) snice they are internal tools for our integration writers
-  we'll use these to make fucnctins our tests can use (e.g.`assert_metrics_are_valid()`) 
- we'll use them in an integration before merging

ibolmo:
closing due to inactivity

PR #949, opened 2025-09-25
[AI SDK]: Set 'model' from wrapGenerate callback if it cannot be found in the result

danielericlee:
As discussed [here](https://github.com/vercel/ai/issues/8563), sometimes the provider model id is not returned from the result of a call from `wrapGenerate`. However, `wrapGenerate` passes `model` as an argument to its callback. In the event that we cannot find the model in the result, we'll fall-back to looking it up from the callback model.

### DIFF
--- a/js/src/wrappers/ai-sdk-shared.test.ts
+++ b/js/src/wrappers/ai-sdk-shared.test.ts
@@ -3,6 +3,7 @@ import { describe, it, expect } from "vitest";
 import {
   detectProviderFromResult,
   extractModelFromResult,
+  extractModelFromWrapGenerateCallback,
   camelToSnake,
   extractModelParameters,
   getNumberProperty,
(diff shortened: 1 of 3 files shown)


danielericlee:
I was able to verify the fix with this script:

```
import { createAmazonBedrock } from "@ai-sdk/amazon-bedrock";
import { generateText, wrapLanguageModel } from "ai";
import { BraintrustMiddleware, initLogger, wrapAISDKModel } from "braintrust";

initLogger({
  projectName: "Bedrock Bug",
  apiKey: process.env.BRAINTRUST_API_KEY,
});

const bedrock = createAmazonBedrock({
  region: "us-east-1",
  apiKey:
    "<api-key>",
});

const model = wrapLanguageModel({
  model: bedrock("us.anthropic.claude-sonnet-4-20250514-v1:0"),
  middleware: [
    {
      wrapGenerate: async ({ doGenerate, params, model }) => {
        const result = await doGenerate();
        console.log("🔸 Generate result:", JSON.stringify(result, null, 2));
        console.log("🔸 Model:", JSON.stringify(model, null, 2));
        return result;
      },
    },
    BraintrustMiddleware({ debug: true, name: "MyMiddleware" }),
  ],
});

async function main() {
  const result = await generateText({
    model,
    prompt: "What is the capital of California?",
  });

  console.log("Response:", result.text);
}

main().catch(console.error);
```

image

PR #954, opened 2025-09-30
Add `wrapClaudeAgentSDK` to Python & JS SDK

knjiang:


js/src/wrappers/claude-agent-sdk.test.ts
```diff
@@ -0,0 +1,165 @@
...
+import { initLogger, _exportsForTestingOnly } from "../logger";
+import { configureNode } from "../node";
+import { z } from "zod";
+
+// Try to import the Claude Agent SDK - skip tests if not available
+let claudeSDK: unknown;
+try {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore - SDK may not be installed
+  claudeSDK = await import("@anthropic-ai/claude-agent-sdk");
```
ibolmo, line 19: For testing environment we should probably just assume the dependency is installed. I know we've used makefile to conditionally install them. If we want to continue that practice, we should just exclude this file from the sdk core test run and manually install (finally remove) and then run this test. 

Personally, I would rather just add the dev dependency and assume the dep exists and minimize the complexity.

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+      const span = startSpan({
+        name: "Claude Agent",
+        spanAttributes: {
+          type: SpanTypeAttribute.TASK,
+        },
+        event: {
+          input:
+            typeof prompt === "string"
+              ? prompt
+              : { type: "streaming", description: "AsyncIterable<SDKMessage>" },
```
ibolmo, line 94: I expected this to prompt always. Perhaps we just need to serialize the AsynIterable<SDKMessage> (prompt) if it is not a string

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+          metadata: options,
```
ibolmo, line 95: Looks like there's a lot of [non-primitive objects in here](https://docs.claude.com/en/api/agent-sdk/typescript#options) (i.e. AbortController). Does the JS SDK handle these well?

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+
+        if (finalMessageContent) {
+          finalResults.push(finalMessageContent);
+        }
+
+        currentMessages.length = 0;
+      };
+
+      // Create wrapped async generator that maintains span context
+      const wrappedGenerator: AsyncGenerator<SDKMessage, void, unknown> =
```
ibolmo, line 135: When I'm creating new integrations I always try to reuse the `start_span` context manager (in Python) and in JS land use the `traced` method. Did you happen to try `traced` here? I'm curious how it would handle `AsyncGenerator`.

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+
+      let ended = false;
+      const finalResults: Array<{ content: unknown; role: string }> = [];
+
+      // Track messages by message.message.id to group streaming updates
+      let currentMessageId: string | undefined;
+      let currentMessageStartTime = getCurrentUnixTimestamp();
+      const currentMessages: SDKMessage[] = [];
+
+      const endSpan = () => {
```
ibolmo, line 107: #nit I seems silly to have a separate method. I also wonder what span.end(); span.end() will do. Is it not idempotent?

js/src/wrappers/claude-agent-sdk.ts
ibolmo: Oh there's a `wrapTracedAsyncGenerator`

ibolmo: I'm guessing there's just a lot to handle here compared to the default behavior of the traced methods

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+            }
+
+            // Create span for final message group
+            await createLLMSpan();
+
+            // Log final output to top-level span - just the last message content
+            span.log({
+              output:
+                finalResults.length > 0
+                  ? finalResults[finalResults.length - 1]
```
ibolmo, line 177: See [this convo](https://braintrustdata.slack.com/archives/C083GCUTVDZ/p1759186705635059). I would check if you need to do any aggregation on the `finalResults` and metrics. If it looks good for you then you can skip my comment.

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+ *   })
+ * );
+ *
+ * const wrappedTool = wrapClaudeAgentTool(myTool);
+ * ```
+ */
+export function wrapClaudeAgentTool<T>(
+  toolDef: SdkMcpToolDefinition<T>,
+): SdkMcpToolDefinition<T> {
+  const originalHandler = toolDef.handler;
```
ibolmo, line 225: (just an observation) it's interesting that this method seems simpler than the previous (no Proxy method). I had expected something like this in the previous one, but I'm not picky.

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+          metadata: {
+            tool_name: toolDef.name,
+            tool_description: toolDef.description,
+          },
+        });
+
+        const result = await originalHandler(args, extra);
+
+        span.log({
+          output: result.content,
```
ibolmo, line 241: looks like traced already collects the output. Should we just use the whole `result`? (trying to minimize massaging). That avoids the `is_error` in the metadata, for example.

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+          ended = true;
+          span.end();
+        }
+      };
+
+      // Create an LLM span for accumulated messages with the same message ID.
+      // LLM spans can contain multiple streaming message updates. We create the span
+      // when we proceed to a new message ID or when the query completes.
+      const createLLMSpan = async () => {
+        const finalMessageContent = await _createLLMSpanForMessages(
```
ibolmo, line 118: I was expecting that we return the output from the sdk, and for us to add a converter for the claude-agent. Looks like all the information to convert from a claude agent sdk message to an anthropic message is preserved? if so the converter [becomes really simple like it has been for me](https://github.com/braintrustdata/braintrust/pull/7884/files#diff-7595beaf4f95783c025262d76af3b0634db5abeddf440b9037a71ed60b0f7d4dR185). That simplifies things here

ibolmo: btw, I would normally find a way to get closer to the method that calls the anthropic api. this way the creation and handling of the spans becomes simpler. Could avoid the is new message logic, for example, if we found the entrypoint that calls the AI and just traced that method.

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -0,0 +1,505 @@
...
+ * // Tools created with wrapped tool() are automatically traced
+ * const calculator = tool("calculator", "Does math", schema, handler);
+ * ```
+ */
+export function wrapClaudeAgentSDK<T extends object>(sdk: T): T {
+  const cache = new Map<PropertyKey, unknown>();
+
+  return new Proxy(sdk, {
+    get(target, prop, receiver) {
+      if (cache.has(prop)) {
```
ibolmo, line 449: :mind_blown: never thought of doing something like this 😅

ibolmo (approved):
I don't think I found any blockers and I would rather default to approval. 

Lot's of thinkers, though. Potentially we could make the integration even simple.

js/src/wrappers/claude-agent-sdk.ts
knjiang: lgtm

py/src/braintrust/wrappers/_anthropic_utils.py
```diff
@@ -0,0 +1,93 @@
+"""Shared utilities for Anthropic API wrappers."""
```
ibolmo, line 1: you should start a `claude_agent_sdk` folder (include a `__init__.py`) and that way you can move all the things in there. 

```
./claude_agent_sdk
  ./utils.py
  ...
```

py/src/braintrust/wrappers/claude_agent_sdk.py
```diff
@@ -0,0 +1,420 @@
...
+        options = claude_agent_sdk.ClaudeAgentOptions(
+            model="claude-sonnet-4-5-20250929"
+        )
+        async with claude_agent_sdk.ClaudeSDKClient(options=options) as client:
+            await client.query("Hello!")
+            async for message in client.receive_response():
+                print(message)
+        ```
+    """
+    return ClaudeAgentSDKWrapper(sdk)
```
ibolmo, line 417: intersting that you'd wrap the whole import like this. I suppose this works well, but I think users may be confused if they `from claude_agent_sdk import ...` in another file. 

The way I've been doing this is to:

```
def setup_claude_agent_sdk(...init_logger params...):
  # try to import and fail silently the setup
  # if dependency available, then monkey patch

def wrap_claude_sdk_client(Client):
  if is_patched(Client):
     return Client
  
  # patch particular methods I'm interested in

  mark_patched(Client)
  return Client

# and so on
```
 
You can see this [here](https://github.com/braintrustdata/braintrust-sdk/pull/947/files#diff-be274fdb2c2312f12814ab3789c80ab822e06dceaa807026af6cb9f195d9f4d4R36).


This way they get both. If they run the `setup_*` they "instrument" (monkey patch) and if they use the `wrap_*` directly then they get the opt-in behavior.

-- 

I think if you wanted to keep things as is then I would just expect that your example would wrap the particular client instance rather than the whole import module (that's more aligned with existing wrap_*, if i'm not mistaken)

ibolmo (approved):
defaulting to approve to unblock, but left a design decision question for your consideration

js/src/wrappers/claude-agent-sdk.ts
```diff
@@ -284,18 +281,6 @@ export function wrapClaudeAgentTool<T>(
...
-/**
- * Wraps all tools in an array of Claude Agent SDK tool definitions.
- *
- * @param tools - Array of tool definitions
- * @returns Array of wrapped tool definitions with tracing
- */
-export function wrapClaudeAgentTools<T>(
-  tools: Array<SdkMcpToolDefinition<T>>,
-): Array<SdkMcpToolDefinition<T>> {
-  return tools.map(wrapClaudeAgentTool);
```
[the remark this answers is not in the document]
knjiang: i think i just inlined

py/src/braintrust/wrappers/claude_agent_sdk.py
knjiang: yeah, i think that might be better devex, i can try that

py/noxfile.py
```diff
@@ -75,6 +77,17 @@ def test_pydantic_ai(session, version):
...
     _run_core_tests(session)
 
+@nox.session()
+@nox.parametrize("version", CLAUDE_AGENT_SDK_VERSIONS, ids=CLAUDE_AGENT_SDK_VERSIONS)
+def test_claude_agent_sdk(session, version):
+    # claude_agent_sdk requires Python >= 3.10
+    if sys.version_info >= (3, 10):
+        _install_test_deps(session)
+        npm_bin = _install_npm_in_session(session)
+        session.run(npm_bin, "install", "-g", "@anthropic-ai/claude-code", external=True)
```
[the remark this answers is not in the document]
knjiang: no this is correct, we are installing claude code because the python sdk calls out to it

PR #962, opened 2025-10-02
consistent choice of span exporter

clutchski:


py/src/braintrust/span_identifier_v4.py
```diff
@@ -138,6 +138,17 @@ def add_hex_field(orig_val, field_id):
             raw_bytes += bytes(json.dumps(json_obj, separators=(",", ":")).encode())
         return base64.b64encode(raw_bytes).decode()
 
+    @staticmethod
+    def get_version(slug: str) -> int:
```
clutchski, line 142: just used in tests

PR #967, opened 2025-10-04
BRA-2992 move langchain-js braintrust dep to peer

ibolmo:
we recently bumped braintrust minor version which causes a ripple of dependency problems. previously things just worked because we were still on 0.3.x but that was hiding the problem. 

the move to peer dependency is a breaking change. thus, we do need to bump from 0.0.x to minor. we will now require folks to directly depend on the braintrust library. 

with this approach we should be using the exact `currentSpan` (should be just one copy of braintrust), so this will also settle those surprises.

### DIFF
--- a/integrations/langchain-js/package.json
+++ b/integrations/langchain-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@braintrust/langchain-js",
-  "version": "0.0.9",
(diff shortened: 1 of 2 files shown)


manugoyal (approved):
makes sense to me!

Issue #966, opened 2025-10-03
wrapAISDK does not pass through all the return values from ai-sdk methods


kngroo:
Braintrust's wrapAISDK wrapped instances returns a plain object { ...result } which loses:
- Methods like toTextStreamResponse() (on prototype)
- Getters like textStream (non-enumerable)

PR #970, opened 2025-10-07
[ai-sdk]: Preserve result object when returning from wrappedStreamObject

danielericlee:
A potential fix for https://github.com/braintrustdata/braintrust-sdk/issues/966

The problem with returning `{...result}` from the `wrappedStreamObject` is that the spread operator will not included properties from the object's prototype chain. In this case, that means removing things like [toTextStreamResponse](https://ai-sdk.dev/cookbook/next/stream-object#server).

Returning the object directly and using `Object.defineProperty` to override `partialObjectStream` appears to solve the issue.

### DIFF
--- a/js/src/wrappers/ai-sdk-v3.ts
+++ b/js/src/wrappers/ai-sdk-v3.ts
@@ -285,10 +285,13 @@ export function wrapAISDK<T extends AISDKMethods>(
         });
       });
 
-      return {
-        ...result,
-        partialObjectStream: wrapStream,
-      };
+      Object.defineProperty(result, "partialObjectStream", {
+        value: wrapStream,
+        writable: true,
+        enumerable: true,
+        configurable: true,
+      });
+      return result;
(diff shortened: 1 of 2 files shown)


knjiang (approved):

ibolmo (approved):

js/src/wrappers/ai-sdk-v3.ts
```diff
@@ -279,16 +277,36 @@ export function wrapAISDK<T extends AISDKMethods>(
...
-          metrics: { time_to_first_token: (Date.now() - startTime) / 1000 },
-        });
-      });
+      const startTime = Date.now();
+      let receivedFirst = false;
 
-      return {
-        ...result,
-        partialObjectStream: wrapStream,
+      const trackFirstAccess = () => {
```
ibolmo, line 283: left a comment separately in slack for your consideration

PR #1000, opened 2025-10-17
[AI SDK] Fix load prompt

knjiang:


js/src/wrappers/ai-sdk-v2.ts
```diff
@@ -57,6 +57,8 @@ export interface MiddlewareConfig {
   debug?: boolean;
   /** Name identifier for the middleware instance */
   name?: string;
+  /** Additional metadata to attach to spans (e.g., prompt metadata) */
+  metadata?: Record<string, unknown>;
```
ibolmo: seems odd to pass the span metadata 🤔 usually that's specific to a span, and here we're adding it to the constructor for the middleware

PR #1013, opened 2025-10-21
[PY SDK] Wrap anthropic client's `beta` field  

knjiang:
we already do this for JS SDK -> https://github.com/braintrustdata/braintrust-sdk/pull/685


clutchski (approved):
add a test and then yes

PR #1028, opened 2025-10-29
fix python remote evals

clutchski:
Fixes a remote evals bug in python where evals would not be marked as completed.

They evals were done, but the data wasn't sent properly back to the backend.

### DIFF
--- a/py/src/braintrust/version.py
+++ b/py/src/braintrust/version.py
@@ -1,4 +1,4 @@
-VERSION = "0.3.5"
+VERSION = "0.3.6"
 
 # this will be templated during the build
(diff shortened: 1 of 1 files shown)


py/src/braintrust/framework.py
```diff
@@ -1580,7 +1580,7 @@ def build_local_summary(
     }
     return ExperimentSummary(
         experiment_id=None,
-        experiment_name=evaluator.experiment_name,
+        experiment_name=evaluator.eval_name,
```
knjiang: hm.. is this code only used for evals?

PR #1031, opened 2025-10-30
Slight improvements to golden tests

ibolmo:
swap location to city_and_state to confirm params are not coerced during conversion


colsondonohue (approved):
nice

Issue #1027, opened 2025-10-29
ESM build uses dynamic require() for OpenTelemetry packages, causing errors in ESM contexts


saurfang:
## Description

The ESM build of braintrust (`.mjs` file) uses dynamic `require()` calls (via `__require()`) to load OpenTelemetry packages. This doesn't work in ESM contexts and causes the following error:

```
Dynamic require of '@opentelemetry/sdk-trace-base' is not supported
```

This results in the error message "OpenTelemetry packages are not installed" even when the packages are properly installed.

## Reproduction Steps

1. Use braintrust in an ESM context (e.g., with `tsx` in ESM mode)
2. Try to use `BraintrustSpanProcessor`
3. The error occurs during initialization

## Expected Behavior

The ESM build should properly import OpenTelemetry packages using ESM imports rather than dynamic requires.

## Workaround

Currently, we're working around this by forcing the use of the CommonJS build:

```typescript
import { createRequire } from 'module';
const require = createRequire(import.meta.url);
const braintrust = require('braintrust');
```

## Environment

- braintrust version: latest
- Node.js version: 22
- Running with tsx in Docker development environment

## Impact

This prevents proper usage of braintrust in ESM environments, which is increasingly common with modern Node.js applications.

Lorenzo45:
I’m encountering this issue when integrating Braintrust into a Next.js + Bun app using @vercel/otel.

Environment:

```
Next.js: 16.0.0
Bun: 1.3.1
Node API target: Bun ESM runtime
braintrust: 0.4.8
@vercel/otel: 2.1.0
@opentelemetry/api: 1.9.0
@opentelemetry/sdk-trace-base: 2.2.0
@opentelemetry/resources: 2.2.0
@opentelemetry/exporter-trace-otlp-http: 0.207.0
@opentelemetry/semantic-conventions: 1.37.0
typescript: ^5
```

Instrumentation code (copied right from Braintrust onboarding page):

```
import { registerOTel } from "@vercel/otel";
import { BraintrustExporter } from "braintrust";

export function register() {
  registerOTel({
    serviceName: "my-braintrust-app",
    traceExporter: new BraintrustExporter({
      filterAISpans: true,
      parent: `project_name:${process.env.PROJECT_NAME}`,
    }),
  });
}
```

TypeScript error:

```
Type 'BraintrustExporter' is not assignable to type 'SpanExporterOrName | undefined'.
[...]

PR #1032, opened 2025-10-31
Modify Otel imports to work in ESM environments

cpinn:
### Overview
Customers using the JS SDK in ESM-only environments are hitting failures because OpenTelemetry (OTEL) is loaded as a CommonJS module. This forces them to fall back to the CommonJS build.  

[Issue](https://github.com/braintrustdata/braintrust-sdk/issues/1027)

### Changes Made
The Braintrust JS SDK does not require OTEL to be installed.
Attempted to make the following changes:
 - Defer loading OTEL until it’s actually used.
 - **CommonJS-capable environments:** load synchronously via `require`.  
 - **ESM-only environments:** load asynchronously via dynamic `import()` with a timeout.
 - Mark OTEL dependencies as external and peer dependencies
 - Fix typescript errors resulting from mismatched typing between original otel types and minimal types defined for when Otel is not installed

### Testing
- Reproduced the issue using the example: `examples/otel/distributed-tracing.ts`.  
- Verified that pointing the example at the newly built files eliminated the failure to locate otel and other typescript errors


js/src/otel/context.ts
```diff
@@ -29,18 +35,75 @@ interface Context {
...
+  if (OTEL_AVAILABLE !== null) {
+    return;
+  }
+
+  if (otelInitPromise) {
+    return;
+  }
+
+  // CommonJS/Node.js: Load via synchronous require()
+  if (typeof require !== "undefined") {
```
cpinn: Still TODO likely needs to be a better/stronger check!

js/package.json
```diff
@@ -77,6 +77,10 @@
     "@jest/globals": "^29.7.0",
     "@nodelib/fs.walk": "^1.2.8",
     "@openai/agents": "^0.0.14",
+    "@opentelemetry/api": "^1.9.0",
```
clutchski: if we add this do the "make no-otel" tests still run?

js/src/otel.ts
```diff
@@ -1,9 +1,13 @@
 // Conditional imports for OpenTelemetry to handle missing dependencies gracefully
 import { SpanComponentsV4 } from "../util/span_identifier_v4";
 import { SpanObjectTypeV3 } from "../util/span_identifier_v3";
+import { importWithTimeout } from "./import-utils";
 
+// Type definitions that don't depend on OpenTelemetry being installed
 interface OtelContext {
```
clutchski: so, i think it might be nice if we can isolate all of the "is otel installed?" code in oen place and "app code" e.g. our span processor in another file

integrations/langchain-js/src/BraintrustCallbackHandler.test.ts
```diff
@@ -643,7 +643,7 @@ it("should handle LangGraph state management", async () => {
 
   async function sayHello(state: HelloWorldGraphState) {
     const res = await model.invoke("Say hello");
-    return res.content;
+    return { message: res.content };
```
cpinn: Changing otel to be async and optionally loaded when otel has a global context caused issues across tests.

In this case when the test for `should handle LangGraph state management` returned an error: 
InvalidUpdateError: Expected node "sayHello" to return an object or an array containing at least one Command object, received string

Troubleshooting URL: https://langchain-ai.github.io/langgraphjs/troubleshooting/errors/INVALID_GRAPH_NODE_RETURN_VALUE/

The span for that was then found in the next test since hte logs were not flushed. 

Still looking into how to properly reset state but it appears that the validation changed around langchain 0.2.25 (?) (might want to supprot the old and the new).

PR #1044, opened 2025-11-04
fix(py): flatten single-field Pydantic models in parameter JSON schemas

dpguthrie:
## Problem

When using Python remote evals with single-field Pydantic parameter models, the Braintrust Playground UI shows an error: **"must be object - The value will not be saved"** when trying to edit string parameters.

This happens because:
1. The Python SDK's `validate_parameters()` function automatically **unwraps** single-field models (when field is named `value`)
2. But `parameters_to_json_schema()` was generating the **full object schema**
3. The UI validated input against the object schema and expected `{"field": "value"}` instead of just `"value"`

This creates an inconsistency between:
- **TypeScript/Zod**: `z.string()` → UI accepts plain string ✅
- **Python/Pydantic**: Single-field model → UI expects object ❌

## Solution

Detect ANY single-field Pydantic model and flatten the schema to match TypeScript's behavior:

**Before:**
```json
{
  "type": "object",
  "properties": {
    "value": {
      "type": "string",
      "default": "...",
      "description": "..."
    }
  }
}
```

**After:**
```json
{
  "type": "string",
  "default": "...",
  "description": "..."
}
```

Also addresses the TODO comment by extracting `default` and `description` from both single-field and multi-field models.

## Examples That Now Work

```python
class SystemPromptParam(BaseModel):
    value: str = Field(default="...", description="...")

class InstructionsParam(BaseModel):
    instructions: str = Field(default="...")

class PromptParam(BaseModel):
    prompt: str = ...
```

## Testing

Tested with the example from the documentation in `examples/eval_parameters.py` pattern.

Fixes the "must be object" error when editing parameters in the Braintrust Playground.

## Related

- Consistent with TypeScript SDK behavior using Zod schemas
- Follows the pattern documented in remote-evals.mdx

### DIFF
--- a/py/src/braintrust/parameters.py
+++ b/py/src/braintrust/parameters.py
@@ -31,7 +31,9 @@ def _pydantic_to_json_schema(model: Any) -> dict[str, Any]:
         # pydantic 1
         return model.schema()
     else:
-        raise ValueError(f"Cannot convert {model} to JSON schema - not a pydantic model")
+        raise ValueError(
+            f"Cannot convert {model} to JSON schema - not a pydantic model"
+        )
 
 
 def validate_parameters(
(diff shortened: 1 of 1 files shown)


subhashishjs:
@ankrgyl this is exactly the problem we’re struggling with. Since we don’t have this, we’re forced to create a TypeScript bridge between our server and Braintrust. Is there any progress on this?

PR #1047, opened 2025-11-05
Fix concurrent calls in one parent for LangChain Python/jS

ibolmo:
Redo the constructor to avoid auto selecting the current active span

- **handle logger/parent and without context tracing with evals**
- **add coverage for setGlobalHandler use case**
- **allow evals and global handlers to track root spans for seamless concurrent calls**

image

Previously all of the task spans would be in only one root trace.


py/src/braintrust/logger.py
```diff
@@ -5193,9 +5195,21 @@ def submit(self, fn: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
...
 
+        # Capture the current thread-local background logger override
+        # This is necessary because worker threads have their own thread-local storage
+        override_logger = getattr(_state._override_bg_logger, "logger", None)
+
         def wrapped_fn(*args, **kwargs):
-            # Run the function inside the captured context
-            return context.run(fn, *args, **kwargs)
+            # Restore the background logger override in the worker thread
+            if override_logger is not None:
```
clutchski: i wonder if there's an way to make this non-specific t othe override logger ... are there other state variables that need ot be copied?

clutchski: https://github.com/braintrustdata/braintrust-sdk/blob/main/py/src/braintrust/logger.py#L460-L477

I guess I was thinking that maybe there was other state that might need to be copied as well. I wasn't sure if copying the background logger only was enough?

clutchski (approved):

py/noxfile.py
```diff
@@ -56,7 +56,11 @@
 # Test matrix
 ANTHROPIC_VERSIONS = (LATEST, "0.50.0", "0.49.0", "0.48.0")
 OPENAI_VERSIONS = (LATEST, "1.77.0", "1.71", "1.91", "1.92")
-LITELLM_VERSIONS = (LATEST, "1.74.0")
+# litellm latest uses Python 3.10+ type syntax (str | List[str])
```
ibolmo: trying to fix here: https://github.com/braintrustdata/braintrust-sdk/pull/1151

PR #1056, opened 2025-11-07
Add TTFT (Time-To-First-Token) tracing to LangChain Braintrust callback

fj11:
This PR adds Time-To-First-Token (TTFT) tracking to the Braintrust LangChain callback so that streaming model latency can be measured accurately and surfaced in Braintrust metrics.

Details
- Start timestamp captured at LLM / ChatModel invocation
- First token arrival recorded in on_llm_new_token
- TTFT stored in metrics as time_to_first_token
- Works for streaming + async LangChain pipelines
- No behavior change for non-streaming calls


ibolmo:
Here's the PR to replace this one. It's easier for us to merge in your changes and run the tests on our own PRs. Thanks for the contribution. I'll make sure this gets the finish line:

https://github.com/braintrustdata/braintrust-sdk/pull/1056/files

PR #1057, opened 2025-11-08
[JS SDK] Build integration tests and ESM support

cpinn:
### Problem

Recently the js-sdk broke when using otel with ESM. This has been upsetting some customers.

### Changes

- Adds integration validation tests for CJS and ESM environments
This PR introduces two tests that will run on builds testing validation in CJS and ESM environments. One test sends a simple span and logs some information and one test uses the otel node-sdk and tracer to send a span. 

- Adds support in ESM builds for commonjs syntax
This PR also add a top level banner to the esm build file that adds support for using commonjs syntax and creating a require function to allow Otel functionality in ESM environments.
   - Failing integration test build: https://github.com/braintrustdata/braintrust-sdk/actions/runs/19204763731/job/54898437331

  - Fixed build after tsup config: https://github.com/braintrustdata/braintrust-sdk/actions/runs/19211150202/job/54913612966?pr=1057

- Fixes Otel tests that skip when precondition is not met to fail the test

### Testing

The failing ESM integration test and now passing ESM integration test
`OpenTelemetry not available` now printing for the otel-no-deps.test.ts
tests are passing in the ci still


js/package.json
```diff
@@ -102,7 +102,8 @@
     "typedoc-plugin-markdown": "^3.17.1",
     "typescript": "5.4.4",
     "vite-tsconfig-paths": "^4.3.2",
-    "vitest": "^2.1.9"
+    "vitest": "^2.1.9",
+    "simple-git": "^3.21.0"
```
cpinn: This is not a mistake. simple-git is used in gitutils.ts in our src files.

js/tsup.config.ts
```diff
@@ -1,68 +1,135 @@
...
-      // Split DTS generation to reduce memory usage
-      compilerOptions: {
-        skipLibCheck: true,
+const esmNodeSupportBanner = {
+  js: `
+import { createRequire as topLevelCreateRequire } from 'module';
+const require = topLevelCreateRequire(import.meta.url);`,
+};
+
+export default defineConfig(() => {
```
cpinn: The banner cannot be added to the CJS code or it would break the CJS app. Had to split the entries individually in order to comply with the array syntax for tsup.

js/src/otel/otel-no-deps.test.ts
```diff
@@ -74,28 +64,19 @@ describe("OpenTelemetry not installed", () => {
...
-    if (otelInstalled) {
-      // Skip this test if OpenTelemetry is installed
-      return;
-    }
-
-    const { otelContextFromSpanExport } = await import(".");
+    const { otel } = await import(".");
 
-    const result = otelContextFromSpanExport("some-export-string");
+    const result = otel.contextFromSpanExport("some-export-string");
```
cpinn: This was incorrect but the test was being skipped so no one realized.

.github/workflows/js.yaml
```diff
@@ -29,4 +36,100 @@ jobs:
...
+          run-id: ${{ github.run_id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Temporarily point braintrust dependency to local build ${{ matrix.integration }}
+        working-directory: js/integrations/${{ matrix.integration }}
```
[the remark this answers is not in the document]
ibolmo: We should also add a js/integrations/README.md for future us. In particular why we choose to use npm vs. pnpm, for example. I recall we ran into surprises with pnpm caching and magic?

js/tsup.config.ts
```diff
@@ -1,68 +1,139 @@
...
-    external: ["zod"],
-    dts: {
-      // Split DTS generation to reduce memory usage
-      compilerOptions: {
-        skipLibCheck: true,
+/**
+ * This banner is used to support node.js require() in ESM modules.
+ * It is required for the Otel available check to work in ESM modules.
+ */
+const esmNodeSupportBanner = {
```
ibolmo: TIL.. so ESM only have a way to work with cjs-only? 😱

js/integrations/span/simple_span_example.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const spans = await backgroundLogger.drain();
+  try {
+    assert.ok(
+      spans.length > 0,
+      "No spans were captured by the background logger",
+    );
+
+    const spanEvent = spans[0];
+    assert.equal(spanEvent.input, "What is the capital of France?");
```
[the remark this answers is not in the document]
ibolmo: let's add this as a comment inside the file

ibolmo (approved):

PR #1065, opened 2025-11-10
add v5 and v6 ai-sdk Agent automatic tracing

ibolmo:
update the v5 golden tests (v6 pending)

### DIFF
--- a/internal/golden/ai-sdk-v6/ai-sdk.ts
+++ b/internal/golden/ai-sdk-v6/ai-sdk.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/consistent-type-assertions */
 import { wrapAISDK, initLogger, traced } from "braintrust";
 import { openai } from "@ai-sdk/openai";
 import { anthropic } from "@ai-sdk/anthropic";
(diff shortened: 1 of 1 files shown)


arthberman:
really needed!

ibolmo:
hey @arthberman I'd be happy to push this to a finish line. I am having a convo with Vercel on a suggestion to allow dependency injection. Right now as the last commit here one will get this kind of trace: 

image


The screenshot may be confusing. What you're seeing is a golden test trace. The first generateText is from a generateText call, and then the Agent.generate is from the Agent specific generation. Notice how the `generateText` (without Agent) shows the tool call, but the `Agent.generate` lacks this. The end result is the same, but there's a fidelity loss of what happens underhood the `Agent.generate`.

They have a suggestion for me to try, but I'm wondering if you're ok with the current output or if you'd like to wait for us to figure out a solution to show more detail.

ibolmo:
I haven't had a chance. I'll try to make some time tomorrow afternoon. Holidays are upon us so I'll try to get it in sooner rather than later.

ibolmo:
Ok got the tool to show up: 
image


It's a little hacky, but better than no solution so let's go with this for now.

PR #1072, opened 2025-11-11
new @braintrust/otel

ibolmo:
For braintrust to work in any environment, we must move our otel support to a separate package.

### DIFF
--- a/integrations/vercel-ai-sdk/package.json
+++ b/integrations/vercel-ai-sdk/package.json
@@ -31,7 +31,7 @@
   "dependencies": {
     "@ai-sdk/provider": "^0.0.11",
     "ai": "^3.2.16",
-    "braintrust": "workspace:>=0.0.141"
+    "braintrust": "workspace:*"
   },
   "peerDependencies": {
     "braintrust": ">=0.0.141"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -118,7 +118,7 @@ importers:
         specifier: ^3.2.16
         version: 3.2.16(react@18.3.1)(svelte@5.43.12)(vue@3.5.24)(zod@3.25.76)
       braintrust:
-        specifier: workspace:>=0.0.141
+        specifier: workspace:*
         version: link:../../js
     devDependencies:
       '@types/node':
@@ -2346,7 +2346,7 @@ packages:
     engines: {node: ^14.15.0 || ^16.10.0 || >=18.0.0}
     dependencies:
       '@jest/types': 29.6.3
-      '@types/node': 20.19.16
+      '@types/node': 22.19.1
       chalk: 4.1.2
       jest-message-util: 29.7.0
       jest-util: 29.7.0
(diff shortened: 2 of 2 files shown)


integrations/otel-js/examples/compat.ts
```diff
@@ -27,10 +28,8 @@ async function main() {
   await login();
   console.log("🚀 Starting OTEL + Braintrust Integration Demo\n");
 
-  const expectedVersion = process.env.BRAINTRUST_OTEL_COMPAT === "true" ? 4 : 3;
-  console.log(
-    `Expected export version: ${expectedVersion} (BRAINTRUST_OTEL_COMPAT=${process.env.BRAINTRUST_OTEL_COMPAT})\n`,
-  );
+  const expectedVersion = 4;
```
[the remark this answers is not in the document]
ibolmo: i think it is because 

https://github.com/braintrustdata/braintrust-sdk/pull/1072/files/33dd7c0cc3e052127a25620e067671a556f4d800#diff-b23ec238e224b3bbeca7ab77afd31322e064409ffe1968f6fc4659daa3665476R20

indeed if they don't call initOtel it should be v3

integrations/otel-js/examples/otel-v1/aisdk_example.ts
```diff
@@ -0,0 +1,67 @@
+import { NodeSDK } from "@opentelemetry/sdk-node";
+import * as ai from "ai";
+import { openai } from "@ai-sdk/openai";
+import { z } from "zod";
+import { BraintrustSpanProcessor } from "@braintrust/otel";
```
ibolmo: neat that the main sdk is no longer needed 🤔

js/smoke/tests/otel-v1/otel_example.ts
```diff
@@ -1,5 +1,5 @@
 /**
```
ibolmo: non-blocking would be nice if these smoke tests live in integrations/otel-js

Makefile
```diff
@@ -71,6 +71,23 @@ js-docs: js-build
 
 js-verify-ci: js-docs js-test
 
+js-test-otel-docker:
```
ibolmo: don't think we're using this

package.json
```diff
@@ -13,7 +13,7 @@
     "start": "turbo run start",
     "lint": "turbo run lint",
     "clean": "turbo run clean",
-    "test": "turbo run test"
+    "test": "turbo run test --filter=\"!@braintrust/otel\""
```
ibolmo: we should codify this in a turbo.json or whatever file

.github/workflows/js.yaml
```diff
@@ -51,14 +55,56 @@ jobs:
         working-directory: js
         shell: bash
         run: |
-          PACKED_TARBALL=$(npm pack --pack-destination dist)
+          mkdir -p artifacts
+          PACKED_TARBALL=$(npm pack --pack-destination artifacts)
           echo "packed_tarball=$PACKED_TARBALL" >> "$GITHUB_OUTPUT"
 
+      - name: Build and pack @braintrust/otel
```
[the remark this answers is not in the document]
cpinn: Additionally some things I would like to do for the integration tests would be:

- switch to using pnpm the same way otel-js managed to instead of using npm to get around the workspacing issues
- possibly switch to docker but that means not running explicitly on windows

Makefile
ibolmo: np! we can leave it in, just wasn't sure of its purpose

js/examples/otel/custom_otel_example.ts
```diff
@@ -1,19 +1,34 @@
-// custom_otel_example.ts
```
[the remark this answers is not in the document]
cpinn: yeah these are followup v2 examples I would like to add:
- otel_compat
- vercel_ai_sdk
- aisdk
I pushed up the tests I knew were working.

.github/workflows/js.yaml
```diff
@@ -68,7 +114,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        integration: [otel, span]
+        integration: [otel-v1, span]
```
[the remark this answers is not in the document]
cpinn: yes we still need a smoke test for v2

PR #1077, opened 2025-11-14
Set new spans added by on_trace_start as current/unset them

CLowbrow:


py/src/braintrust/wrappers/test_openai.py
```diff
@@ -35,6 +36,35 @@ def memory_logger():
         yield bgl
 
 
+def test_tracing_processor_sets_current_span(memory_logger):
```
ibolmo: would be great to see a real example repro the problem and that with this fix fixes the problem.

dpdiliberto: To repro, I used the following:

```
from agents import set_trace_processors, trace

from braintrust import flush, init_logger, start_span
from braintrust.wrappers.openai import BraintrustTracingProcessor


def run_demo():
    bt_logger = init_logger(project="8375-agents-sdk")

    set_trace_processors([BraintrustTracingProcessor(bt_logger)])

    with trace(workflow_name="workflow"):
        with start_span(name="child-span"):
            pass

    flush()


if __name__ == "__main__":
    run_demo()
```

Before the fix, I got the `workflow` root span and `child-span` in separate traces. 

Screenshot 2025-11-14 at 5 12 47 PM

After this fix and after running 

`pip install --force-reinstall \
  "braintrust[openai-agents] @ git+https://github.com/braintrustdata/braintrust-sdk.git@alex/set-correct-parents#subdirectory=py"`

Screenshot 2025-11-14 at 5 13 01 PM


I got the `child-span` span within the `workflow` trace.

PR #1084, opened 2025-11-19
Less memory

ankrgyl:
• - Eval() and runEvaluator() keep returning full per-row results by default again, but you can now explicitly opt out with returnResults: false when you want the low-memory streaming path. Under the hood, the runner still streams eval data and only materializes rows if that flag is true.
  - The progress reporter API gained an optional setTotal() hook (implemented by both SimpleProgressReporter and BarProgressReporter) so the streaming scheduler can update progress bars as it discovers more datapoints.
  - ObjectFetcher (datasets/experiments) and runEvaluatorInternal both consume async iterables without buffering entire datasets, so BaseExperiment/migration flows don’t blow up memory.
  - The background logger now flushes logs in configurable chunks (BRAINTRUST_LOG_FLUSH_CHUNK_SIZE, default 25) instead of stringifying an entire queue at once, preventing huge log payloads from exhausting the heap.
  - Tests that read result.results opt into retention when necessary, and EvalOptions expose returnResults for callers who want to disable it.

### DIFF
--- a/js/src/framework.ts
+++ b/js/src/framework.ts
@@ -1199,7 +1199,14 @@ async function runEvaluatorInternal(
           state: evaluator.state,
         });
       } else {
-        return await experiment.traced(callback, baseEvent);
+        const result = await experiment.traced(callback, baseEvent);
+        // Flush logs after each task to provide backpressure and prevent memory accumulation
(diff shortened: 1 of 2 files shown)


CLowbrow (approved):
Tested this a bit and lgtm

PR #1088, opened 2025-11-20
fix(py): local eval crash when scorer returns `None`

pmmmwh:


clutchski:
closed in favor of #1241 which has tests. thank you thought!

PR #1120, opened 2025-12-02
Fix typing of BraintrustExporter

cpinn:
ExportResult comes from opentelemetry/core so I shallow copied it to avoid the direct dependency to the core library. 

I build and packed the braintrust-otel package and I was able to remove the `as unknown as SpanExporter` from the exporter_example.ts.

Will update the test on next braintrust/otel release
Screenshot 2025-12-02 at 10 43 31 AM
Screenshot 2025-12-02 at 10 50 49 AM

### DIFF
--- a/integrations/otel-js/src/otel.ts
+++ b/integrations/otel-js/src/otel.ts
@@ -17,6 +17,11 @@ import {
 } from "@opentelemetry/sdk-trace-base";
 import { IDGenerator, type Span as BraintrustSpan } from "braintrust";
 
+interface ExportResult {
+  code: number;
+  error?: Error;
+}
+
 const FILTER_PREFIXES = [
   "gen_ai.",
   "braintrust.",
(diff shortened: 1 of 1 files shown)


integrations/otel-js/src/otel.ts
```diff
@@ -17,6 +17,11 @@ import {
 } from "@opentelemetry/sdk-trace-base";
 import { IDGenerator, type Span as BraintrustSpan } from "braintrust";
 
+interface ExportResult {
```
ibolmo: is this in the opentelemetry/api library? we should keep an eye on this but maybe that's why vercel/ai forces that dependency to avoid this

cpinn: The opentelemetry/api doesn't export the ExportResult unfortunately only opentelemetry/core. 

From what I had read before it seemed like an antipattern to have to rely on core and the api should have the proper interface but it looks like in this case it wasn't true which is likely why we ran into this issue in the first place.

PR #1142, opened 2025-12-05
Adding "x-bt-project-id"

stretpjc:
The header "x-bt-project-id" was added in this commit: https://github.com/braintrustdata/braintrust/commit/5ec2f1e71c68d060e5262ff9f990e6280859a7c8

Adding it to the SDK code to fix cors issues from it missing.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "braintrust",
-  "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


ibolmo (approved):

ankrgyl:
This isn't quite the right final solution. Because it will accept, but not use, the project id.

PR #1150, opened 2025-12-09
Parameterize batch size

ankrgyl:


ibolmo (approved):
approving, but left a few possible gaps in js but otherwise lgtm

PR #1156, opened 2025-12-10
Support jinja templates

cpinn:
Run prompts in the sandbox environment, depabtable but seems recommended for untrusted templates

### DIFF
--- a/py/src/braintrust/prompt.py
+++ b/py/src/braintrust/prompt.py
@@ -79,6 +79,13 @@ class PromptData(SerializableDataClass):
             d2["template_format"] = "jinja"
         return super().from_dict_deep(d2)
 
+    def as_dict(self):
+        """Serialize to dict, converting jinja back to nunjucks for API compatibility."""
+        result = super().as_dict()
+        if result.get("template_format") == "jinja":
+            result["template_format"] = "nunjucks"
+        return result
+
 
 @dataclass
 class PromptSchema(SerializableDataClass):


py/src/braintrust/logger.py
```diff
@@ -4572,19 +4574,155 @@ def _html_escape(x: Any) -> Any:
 
 _custom_render = _create_custom_render()
 
+TemplateFormat = Literal["mustache", "jinja", "none"]
 
-def render_templated_object(obj: Any, args: Any) -> Any:
+
+class _JinjaSafeDict:
```
cpinn: I ran into jinja accessing . as function methods during evaluation. This would prevent that by wrapping the dictionary values. Some users may find this and the sandbox environment to be restrictive but start with safety.

cpinn:
> @cpinn is this one still needed?
Yep, still needed, just needed a small update. I had wanted to do a little bit of testing with the api as well.

py/src/braintrust/logger.py
```diff
@@ -4573,18 +4584,152 @@ def _html_escape(x: Any) -> Any:
...
+    try:
+        template_obj = env.from_string(template)
+        if isinstance(data, dict):
+            wrapped_data = {k: _JinjaSafeDict.wrap(v) for k, v in data.items()}
+            variables = {"input": _JinjaSafeDict(data), **wrapped_data}
+        else:
+            variables = {"input": data}
+        return template_obj.render(**variables)
+    except jinja2.UndefinedError as e:
+        raise ValueError(f"Template rendering failed: {str(e)}") from e
```
ibolmo: #aside Pretty we don't do this, but in the future we should have custom error classes

py/src/braintrust/prompt.py
```diff
@@ -63,10 +63,21 @@ class PromptChatBlock(SerializableDataClass):
...
 class PromptData(SerializableDataClass):
     prompt: PromptBlockData | None = None
     options: PromptOptions | None = None
+    template_format: TemplateFormat | None = None
+
+    @classmethod
+    def from_dict_deep(cls, d: dict):
+        d2 = dict(d)
+        if d2.get("template_format") == "nunjucks":
+            d2["template_format"] = "jinja"
```
ibolmo: i take it we're aliasing jinja <=> nunjucks, I'm not seeing something similar in TS. 🤔

py/src/braintrust/logger.py
```diff
@@ -4714,12 +4866,15 @@ def build(self, **build_args: Any) -> Mapping[str, Any]:
         # Extract strict mode setting from build_args (using get to avoid modifying the original dict)
         strict = build_args.get("strict", False)
 
+        # Use the prompt's template_format
+        template_format = self.template_format
+
         params = self.options.get("params") or {}
```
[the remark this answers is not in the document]
cpinn: If I did that it would not match with the api as template_format is currently top level on the prompt. Maybe that is fine in the python sdk(?) but it wouldn't match the current modelling.

When I had been looking and maybe it was not the best modeling choice it had looked like "options" were really for options of the actual prompt model being used (ie. which model, temperature for the model) and not options for prompt setup.

PR #1168, opened 2025-12-14
chore: Explicit exports

ibolmo:
We have multiple public interfaces:

- `/` which is the main point of entry
- `/dev` for the dev / remote evals
- `/browser` which is a more minimal sdk that can be used in more restricted environments. 

When we worked on 1.x, we found a few cases where we neglected to export new wrappers in the right file. Part of this is that we have complicated export process. 

This makes that a little more explicit. Ideally, we wouldn't have the exports- files and the entrypoints would have the literal exports that we want to support in those environments. 

For now, this moves us away from `export *` and now to explicit exports. This also adds new rules to prevent us from regressing and doing `export *`.


js/src/exports-common.ts
```diff
@@ -1,13 +1,134 @@
 // Common exports shared between browser and node builds
 export { LazyValue } from "./util";
-export * from "./logger";
-export * from "./functions/invoke";
-export * from "./functions/stream";
+
+export type {
```
ibolmo: TIL that `export *` also exported the file's types so we need to specify both the types and objects now separately

PR #1171, opened 2025-12-15
Add expect test for Zod v4 schema serialization (BRA-3619)

mklgallegos:
### Context 

Olmo requested an expect test in https://github.com/braintrustdata/braintrust/pull/9205 to demonstrate the Zod v4 serialization issue.

This test shows that `zodToJsonSchema()` doesn't understand Zod v4's schema structure, causing string parameters to serialize without a type field.

Also see: https://linear.app/braintrustdata/issue/BRA-3619/support-zod-v4-for-remote-eval-parameters

### Description 

- Exported `makeEvalParametersSchema()` from `dev/server.ts` so it can be tested directly
- Added `src/zod-serialization.test.ts` with inline snapshots showing:
 - Zod v3: Serializes correctly with `type: "string"`
 - Zod v4: Missing type field entirely

This demonstrates why the playground UI falls back to JSON/YAML instead of showing the "Text" input option for Zod v4 users.

### Testing 

```
pnpm vitest run src/zod-serialization.test.ts
```

Both tests pass. The snapshots capture the current (broken) behavior for Zod v4, so when a fix is implemented, updating the snapshots will show the before/after difference.

### DIFF
--- a/js/dev/server.ts
+++ b/js/dev/server.ts
@@ -389,7 +389,7 @@ function makeScorer(
   return ret;
 }
 
-function makeEvalParametersSchema(
+export function makeEvalParametersSchema(
   parameters: EvalParameters,
 ): z.infer<typeof evalParametersSerializedSchema> {
   return Object.fromEntries(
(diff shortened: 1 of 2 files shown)


ibolmo:
@mklgallegos I think we're good to close this with the new tests @cpinn added?

PR #1184, opened 2025-12-18
fix: upgrade langchain packages and fix tests for breaking changes

ralfpopescu:
Wanted to make sure the SDK worked with latest langchain versions. Found a breaking change in langgraph state API that affected the tests.


integrations/langchain-js/package.json
ibolmo: thank you! I'd like to see if we can have the integration work with both versions. if impossible, we can just bump the major version to mark this new version as only supporting 1.x

PR #1185, opened 2025-12-18
allow smoke tests to be run locally with `make test-smoke ...`

ibolmo:


cpinn (approved):
very nice. This was something I had wanted.

PR #1193, opened 2025-12-19
Add ability to create prompts with a template format

cpinn:
Ensure the template format on the prompt is used unless a user explicitly overrides the rendering behavior. Regenerated types for the sdk.

A user should be able to create a prompt with a template format:
```
const prompt = new Prompt(
      {
        id: "1",
        _xact_id: "xact_123",
        created: "2023-10-01T00:00:00Z",
        project_id: "project_123",
        prompt_session_id: "session_123",
        name: "test",
        slug: "test",
        prompt_data: {
          options: {
            model: "gpt-4o",
          },
          prompt: {
            type: "completion",
            content: "Complete this: {% if text %}{{text}}{% endif %}",
          },
          templateFormat: "nunjucks",
        },
      },
      {},
      true,
    );
```

A user should then have the template format respected during the build rendering.

### DIFF
--- a/generated_types.json
+++ b/generated_types.json
@@ -5289,6 +5289,17 @@
               "$ref": "#/components/schemas/SavedFunctionId"
             }
           },
+          "template_format": {
+            "type": [
+              "string",
+              "null"
+            ],
+            "enum": [
+              "mustache",
+              "nunjucks",
+              "none"
+            ]
+          },
           "mcp": {
             "type": [
               "object",
@@ -5407,6 +5418,17 @@
               "$ref": "#/components/schemas/SavedFunctionId"
             }
           },
+          "template_format": {
+            "type": [
+              "string",
+              "null"
+            ],
+            "enum": [
+              "mustache",
+              "nunjucks",
+              "none"
+            ]
+          },
           "mcp": {
             "type": [
               "object",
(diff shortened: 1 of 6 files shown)


generated_types.json
```diff
@@ -1867,6 +1867,30 @@
             ],
             "format": "date-time",
             "description": "Date the environment variable was last used"
+          },
+          "metadata": {
```
cpinn: The re-generation of types produced more changes than I was expecting.

js/src/prompt.test.ts
```diff
@@ -134,3 +134,129 @@ function testPromptBuild({
...
+        _xact_id: "xact_123",
+        created: "2023-10-01T00:00:00Z",
+        project_id: "project_123",
+        prompt_session_id: "session_123",
+        name: "test",
+        slug: "test",
+        prompt_data: {
+          options: {
+            model: "gpt-4o",
+            templateFormat: "nunjucks",
```
cpinn: I now know we have been using snake_case. I was seeing some mixed camelCase and snakeCase. Since I made all the other parameter camelCase I kept it camelCase here. Never usually see snake_case in typescript code.

ibolmo (approved):
didn't qa, but changes looks solid

Issue #1134, opened 2025-12-04
braintrust Typescript 0.4.9 breaks AI SDK tools using generator


alechoey:
Braintrust Typescript SDK 0.4.9 significantly rewrote the Vercel AI SDK integration. [[pr]](https://github.com/braintrustdata/braintrust-sdk/pull/1021)

https://www.braintrust.dev/docs/changelog#typescript-sdk-version-0-4-9

This breaks tools using async generators that yield status updates to their caller.

```
const createGreetingTool = (logFn) => ({
  description: "A tool that streams a greeting message",
  inputSchema: ai.jsonSchema({
    type: "object",
    properties: {
      name: { type: "string", description: "The name to greet" },
    },
    required: ["name"],
  }),
  execute: async function* ({ name }) {
    logFn(`[execute] Starting generator for: ${name}`);
    yield { status: "starting", message: "Preparing greeting..." };
    logFn("[execute] Yielded starting status");
    yield { status: "processing", message: `Looking up ${name}...` };
    logFn("[execute] Yielded processing status");
    yield { status: "generating", message: "Generating greeting..." };
    logFn("[execute] Yielded generating status");
    const greeting = `Hello, ${name}!`;
    logFn("[execute] Returning final result");
    const result = { status: "done", greeting };
    yield result;
  },
})
```

I created a [reproduction app](https://github.com/alechoey/braintrust-ai-sdk-repro/tree/main) that demonstrates the difference between a wrapped and unwrapped `ai` instance.

This also happens in 1.0.0-rc2.

When running the repro app against 0.4.9, you'll see output like:
```
│=== Tool Results ===                        ││  }                                         │
│[                                           ││}                                           │
│  {                                         ││                                            │
│    "type": "tool-result",                  ││=== Tool Results ===                        │
│    "toolCallId": "call_Jj2nSvhhOujs1uHzgn0 ││[                                           │
│whNGj",                                     ││  {                                         │
│    "toolName": "greeting",                 ││    "type": "tool-result",                  │
│    "input": {                              ││    "toolCallId": "call_CzK0UvrPMd7eA1fPyiq │
│      "name": "World"                       ││zEGq7",                                     │
│    },                                      ││    "toolName": "greeting",                 │
│    "providerMetadata": {                   ││    "input": {                              │
│      "openai": {                           ││      "name": "World"                       │
│        "itemId": "fc_00f4bcfbcd4399de00693 ││    },                                      │
[...]

PR #1203, opened 2025-12-27
Fix improper handling of async generator as execute input

delner:
## Summary

Async generator tools return empty output when using the Braintrust AI SDK wrapper. (#1134)

### The Problem

AI SDK v5 supports tools with async generator `execute` functions that yield intermediate status updates:

```typescript
const greetingTool = ai.tool({
  description: "A tool that streams a greeting",
  inputSchema: z.object({ name: z.string() }),
  execute: async function* ({ name }) {
    yield { status: "starting", message: "Preparing..." };
    yield { status: "processing", message: `Looking up ${name}...` };
    yield { status: "done", greeting: `Hello, ${name}!` };
  },
});
```

When wrapped with Braintrust's `wrapAISDK()`, the tool result was logged as `{}` instead of the final yielded value `{"status":"done","greeting":"Hello, World!"}`.

### Root Cause

The wrapper was using `await` on the result of calling the execute function:

```typescript
const result = await originalExecute.apply(target, args);
span.log({ output: result });
```

When `originalExecute` is an async generator function, calling it returns an `AsyncGenerator` object immediately—**the generator code doesn't run yet**. You must iterate through the generator to execute its code and get the yielded values.

`await`-ing an AsyncGenerator just returns the generator object itself (not the values). When serialized to JSON, this becomes `{}`.

### Solution

Detect when the execute function returns an AsyncGenerator and handle it specially:

1. Check if the result has async iterator protocol methods (`Symbol.asyncIterator`, `next`, `return`, `throw`)
2. If so, return a wrapper async generator that:
   - Iterates through the original generator
   - Yields all intermediate values (preserving streaming behavior)
   - Captures the final yielded value
   - Logs that final value as the tool output
3. If not an async generator, handle as before with simple `await`

### Test Plan

- [x] Added unit test for async generator tool execution
- [x] Verified test fails without fix
- [x] Verified test passes with fix
- [x] All existing tool tests pass (no regressions)
- [x] Full ai-sdk test suite passes

### DIFF
--- a/js/src/wrappers/ai-sdk/ai-sdk.ts
+++ b/js/src/wrappers/ai-sdk/ai-sdk.ts
@@ -796,6 +796,22 @@ const wrapTools = (tools: any) => {
   return wrappedTools;
 };
 
+/**
+ * Checks if a value is an AsyncGenerator.
+ * AsyncGenerators are returned by async generator functions (async function* () {})
+ * and must be iterated to consume their yielded values.
+ */
+const isAsyncGenerator = (value: any): value is AsyncGenerator => {
+  return (
+    value != null &&
+    typeof value === "object" &&
+    typeof value[Symbol.asyncIterator] === "function" &&
+    typeof value.next === "function" &&
+    typeof value.return === "function" &&
+    typeof value.throw === "function"
+  );
+};
+
 const wrapToolExecute = (tool: any, name: string) => {
   // Only wrap tools that have an execute function (created with tool() helper)
   // AI SDK v3-v6: tool({ description, inputSchema/parameters, execute })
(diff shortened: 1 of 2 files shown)


js/src/wrappers/ai-sdk/ai-sdk.ts
```diff
@@ -811,13 +827,50 @@ const wrapToolExecute = (tool: any, name: string) => {
...
+
+                try {
+                  let lastValue: any;
+                  for await (const value of result) {
+                    lastValue = value;
+                    yield value;
+                  }
+                  // Log the final yielded value as the output
+                  span.log({ output: lastValue });
+                  span.end();
```
ankrgyl: span.end() should be in `finally`, and then you can remove it from both the `try` and `catch` cases.

delner: Makes sense. I'll update.

Issue #1051, opened 2025-11-06
Add support for setting environment during prompt creation


steffendsommer:
In my CI/CD pipeline I sync prompts from code to Braintrust with a simple script that uses the Braintryst TypeScript SDK:

```typescript
const project = braintrust.projects.create({ name: 'my-project' });
project.prompts.create({
    slug: 'my-prompt',
    name: 'My Prompt',
    model: 'gpt-4o',
    messages: ...
    ifExists: "replace",
});
```

As far as I can tell, the SDK don't support setting environment to my prompt during creation. That would be great to be able to do that.

PR #1207, opened 2025-12-28
feat: Add support for setting environment during prompt creation

wojtekPi:
## Description:
This PR adds support for specifying an `environment` field when creating prompts in the Braintrust SDK. This allows users to assign prompts to specific environments (e.g., "production", "staging") during creation.
#### JavaScript/TypeScript:
- Added `environment?: string` to `BaseFnOpts` in `framework-types.ts` and `framework2.ts`
- Updated `CodePrompt` class to store and serialize environment data
- Modified `FunctionEvent` interface to include environment in API payloads
- Updated `toFunctionDefinition()` to include environment in API calls
- Modified `upload.ts` to pass environment through the upload pipeline
- Added comprehensive tests for environment support in `framework.test.ts`
#### Python:
- Added `environment: Optional[str] = None` to `CodePrompt` dataclass
- Updated `PromptBuilder.create()` method with environment parameter in all overloads
- Modified `to_function_definition()` to include environment in serialization
- Updated docstring to document the environment parameter
### Usage Example:
```typescript
const project = braintrust.projects.create({ name: 'my-project' });
project.prompts.create({
    slug: 'my-prompt',
    name: 'My Prompt',
    model: 'gpt-4o',
    messages: [...],
    environment: 'production',  // New parameter
    ifExists: "replace",
});
```

## Testing:
- All existing tests pass
- New tests verify environment storage and serialization
- Both JS and Python SDKs support the feature

This addresses issue #1051 by allowing users to set environment during prompt creation, similar to how metadata is currently supported.

### DIFF
--- a/js/src/cli/functions/upload.ts
+++ b/js/src/cli/functions/upload.ts
@@ -45,6 +45,7 @@ interface BundledFunctionSpec {
   function_schema?: FunctionObject["function_schema"];
   if_exists?: IfExists;
   metadata?: Record<string, unknown>;
+  environment?: string;
 }
 
 const pathInfoSchema = z
@@ -114,6 +115,7 @@ export async function uploadHandleBundles({
               : undefined,
           if_exists: fn.ifExists,
           metadata: fn.metadata,
+          environment: fn.environment,
         });
       }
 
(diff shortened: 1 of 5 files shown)


CLowbrow:
This will pass the parameter, but were you able to push a prompt created like this and get it to show up in the UI with the correct environment when you pushed it? I don't think that functionality exists currently (but I could be wrong).

PR #1211, opened 2025-12-29
Fix ADK and py sdk circ references during serialization

ibolmo:


integrations/adk-py/pyproject.toml
```diff
@@ -40,6 +40,9 @@ members = [
     ".",
 ]
 
+[tool.uv.sources]
+braintrust = { path = "../../py", editable = true }
```
ibolmo: will revert this prior to merging. i'll split the PR as well, so that we get the sdk updated first

PR #1212, opened 2025-12-30
Make zod a peer dependency in the sdk

cpinn:
Customers would like the ability to use the braintrust sdk with zod 4. At the moment this is not possible due to a hard dependency on zod v3. 

This PR starts to decouple zod version from the sdk  by making it a peer dependency and start the upgrade to zod v4.

Breaking Change
- zod will now need to be installed separately 

Typing issues:
- All types need to be kept on zod 3 in the sdk as all generated_types from braintrust are only v3.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -1275,10 +1275,7 @@ export class Attachment extends BaseAttachment {
             signedUrl: z.string().url(),
             headers: z.record(z.string()),
           })
-          .parse(await metadataResponse.json()) as {
-          signedUrl: string;
-          headers: Record<string, string>;
-        });
+          .parse(await metadataResponse.json()));
       } catch (error) {
         if (error instanceof ZodError) {
           const errorStr = JSON.stringify(error.flatten());


js/src/zod/zod-utils.ts
```diff
@@ -0,0 +1,100 @@
+import { zodToJsonSchema as zodToJsonSchemaV3 } from "zod-to-json-schema";
+import { z } from "zod";
```
cpinn: Originally I attempted to do what [openai is doing](https://github.com/openai/openai-node/blob/cb3987b4840aedd3ca8400ca8fa444fd927ada73/src/helpers/zod.ts) explicitly importing from /v3 and /v4.

This caused the the main braintrust build to fail so I have fallen back to my older implementation of checking for the JSONSchema method. It does not appear to respect my target version this way though. It did respect the target version when running directly from v4.

js/src/zod/zod-utils.ts
```diff
@@ -0,0 +1,116 @@
...
+    (zodObject as any)._zod !== undefined
+  );
+}
+
+export function zodToJsonSchema(schema: z4.ZodType | z3.ZodType) {
+  if (isZodV4(schema)) {
+    return z4.toJSONSchema(schema as z4.ZodType, {
+      target: "draft-7",
+    });
+  }
```
ibolmo: i _think_ if you do an else statement here the schema is guaranteed a z3.zodType?

ibolmo (commented):
looks pretty solid. i left a question. i assume openai had to do something similar to normalize the output of their zodtojsonschema?

js/src/logger.ts
```diff
@@ -1273,9 +1275,12 @@ export class Attachment extends BaseAttachment {
         ({ signedUrl, headers } = z
           .object({
             signedUrl: z.string().url(),
-            headers: z.record(z.string()),
+            headers: getZodRecord(z.string()),
```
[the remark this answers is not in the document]
cpinn: Without it some of the tests couldn't run.

 FAIL  src/prompt.test.ts > prompt template_format > template_format applies to completion prompts
TypeError: Cannot read properties of undefined (reading '_zod')
 ❯ Object.inst._zod.parse ../node_modules/.pnpm/zod@4.2.1/node_modules/zod/v4/core/schemas.js:1325:46
 ❯ Module.<anonymous> ../node_modules/.pnpm/zod@4.2.1/node_modules/zod/v4/core/parse.js:33:32
 ❯ _.inst.safeParse ../node_modules/.pnpm/zod@4.2.1/node_modules/zod/v4/classic/schemas.js:37:46
 ❯ Prompt.runBuild src/logger.ts:6602:49
    6600|     }
    6601| 
    6602|     const dictArgParsed = z.record(z.unknown()).safeParse(buildArgs);

js/dev/server.ts
```diff
@@ -412,14 +413,14 @@ function makeEvalParametersSchema(
...
         //
         // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-        const schema = zodToJsonSchema(value as any);
+        const schemaObj = zodToJsonSchema(value as any);
         return [
           name,
           {
             type: "data",
-            schema,
-            default: value.default,
```
cpinn: default and description are already extracted into the json object at the top level. Don't need to parse original schema.

js/dev/server.ts
```diff
@@ -45,7 +45,8 @@ import {
 import { EvalParameters, validateParameters } from "../src/eval-parameters";
 import { z } from "zod/v3";
 import { promptDefinitionToPromptData } from "../src/framework2";
-import zodToJsonSchema from "zod-to-json-schema";
+import { zodToJsonSchema } from "../src/zod/utils";
+import { secureHeapUsed } from "crypto";
```
ibolmo: probably an auto import faux pas

js/src/logger.ts
```diff
@@ -1275,7 +1275,10 @@ export class Attachment extends BaseAttachment {
             signedUrl: z.string().url(),
             headers: z.record(z.string()),
           })
-          .parse(await metadataResponse.json()));
+          .parse(await metadataResponse.json()) as {
```
ibolmo: #nit i didn't think you'd need the type coersion now

js/dev/server.ts
cpinn: yes thank you!

PR #1217, opened 2025-12-31
Fix missing structured output on generate spans

### LINKED ISSUE #1208: Structured output not being logged when wrapping Vercel AI SDK
I am using bt v1.1.1 and the AI SDK v5.0.113. Structured outputs are not logged in the generateObject or doGenerate spans. This works on v0.4.8 so I think there is a regression due to [this](https://github.com/braintrustdata/braintrust-sdk/pull/1021) PR.

### LINKED ISSUE #1214: Wrapped Vercel AI SDK drops prompt-related span_info
https://github.com/braintrustdata/braintrust-sdk/blob/main/js/src/wrappers/ai-sdk/ai-sdk.ts#L150 doesn't seem to include the prompt id. as a result, during evals, all calls using `generateText` and `streamText` etc. don't get linked to the underlying prompt, even if it's managed by BT.

there seems to be similar issues reported before

*  https://github.com/braintrustdata/braintrust-sdk/issues/1081
* https://github.com/braintrustdata/braintrust-sdk/pull/1083

but neither resolved the problem. it's kind of a deal breaker -- would love for this to be fixed.

delner:
## Summary

Fixes #1208 - Structured outputs not being logged in `generateObject`/`doGenerate` spans since v0.4.9.
Fixes #1214 - Wrapped Vercel AI SDK drops prompt-related `span_info`.

### Problem

After upgrading from braintrust v0.4.8 to v0.4.9+, users reported that structured output metadata (particularly `response_format` with the JSON schema) was no longer appearing in spans for `generateObject` and `doGenerate` calls. The v0.4.8 spans showed rich metadata including the full schema definition, while v0.4.9+ showed only minimal metadata (model, provider, braintrust tags).

Additionally, when using Braintrust-managed prompts with the AI SDK wrapper, the `span_info` metadata (containing prompt id, project_id, version) was being dropped, making it impossible to link spans back to their source prompts.

### Root Cause

PR #1021 (ai-sdk rewrite) removed the `extractModelParameters()` call that previously captured all generation parameters, and did not handle `span_info` from compiled Braintrust prompts.

#### Issue 1208: Structured output schema

Based on reviewer feedback, the schema is now captured in the **input** field rather than metadata, following the principle that user input should stay in input:

1. **Added schema serialization in `processInputAttachments`**: When a Zod schema is provided to `generateObject`/`streamObject`, it's converted to JSON Schema and included in the span input.

2. **Added `responseFormat` to `DO_GENERATE_EXCLUDE_KEYS`**: Ensures `responseFormat` stays in input at the `doGenerate`/`doStream` level rather than being moved to metadata.

3. **Added `extractModelParameters` to `doGenerate` and `doStream`**: Captures generation parameters like `temperature`, `maxTokens`, etc. in metadata.

4. **Fixed provider extraction for all wrapper functions**: Changed from `initialProvider` to `effectiveProvider = params.model?.provider || initialProvider` to properly extract the provider from the model object.

5. **Added `finish_reason` to `doGenerate` and `doStream` metadata**: Restores the finish reason capture that was missing in the rewrite.

#### Issue 1214: Braintrust-managed prompt metadata

Added `span_info` support to all four wrapper functions (`generateText`, `streamText`, `generateObject`, `streamObject`):

- `span_info.name` overrides the span name
- `span_info.spanAttributes` merged into span attributes
- `span_info.metadata` (containing `prompt.id`, `project_id`, `version`, `variables`) merged into event metadata

This makes the AI SDK wrapper consistent with the OpenAI wrapper's handling of Braintrust-managed prompts.

## Example output

I tested this locally and generated these spans:

Screenshot from 2026-01-01 11-05-12

Screenshot from 2026-01-01 11-05-46

### DIFF
--- a/js/src/wrappers/ai-sdk/ai-sdk.ts
+++ b/js/src/wrappers/ai-sdk/ai-sdk.ts
@@ -8,40 +8,6 @@ import {
 } from "../attachment-utils";
 import { zodToJsonSchema } from "../../zod/utils";
 
-// Helper to convert camelCase to snake_case for metadata keys
-function camelToSnake(str: string): string {
-  return str.replace(/[A-Z]/g, (letter) => `_${letter.toLowerCase()}`);
-}
-
-// Keys to exclude from model parameters at the doGenerate/doStream level
-const DO_GENERATE_EXCLUDE_KEYS = new Set([
-  "prompt", // Already captured as input
-  "responseFormat", // Already captured as input (user's schema request)
-  "providerOptions", // Internal AI SDK configuration
-  "abortSignal", // Not useful for logging
-  "headers", // Request headers, not model parameters
-]);
-
-/**
- * Extracts model parameters from options, excluding specified keys.
- * Converts camelCase keys to snake_case for consistency.
- */
-function extractModelParameters(
-  params: Record<string, unknown>,
-  excludeKeys: Set<string>,
-): Record<string, unknown> {
-  const modelParams: Record<string, unknown> = {};
-
-  for (const [key, value] of Object.entries(params)) {
-    if (value !== undefined && !excludeKeys.has(key)) {
-      const snakeKey = camelToSnake(key);
-      modelParams[snakeKey] = value;
-    }
-  }
-
-  return modelParams;
-}
-
 // list of json paths to remove from output field
 const DENY_OUTPUT_PATHS: string[] = [
   // v3
@@ -176,9 +142,7 @@ const makeGenerateTextWrapper = (
       spanAttributes: spanInfoAttrs,
     } = span_info ?? {};
 
-    const { model: initialModel, provider: initialProvider } =
-      serializeModelWithProvider(params.model);
-    const effectiveProvider = params.model?.provider || initialProvider;
+    const { model, provider } = serializeModelWithProvider(params.model);
 
     return traced(
       async (span) => {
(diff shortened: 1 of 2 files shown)


js/src/wrappers/ai-sdk/ai-sdk.ts
```diff
@@ -133,9 +167,18 @@ const makeGenerateTextWrapper = (
...
+    const { span_info, ...params } = allParams;
+    const {
+      metadata: spanInfoMetadata,
+      name: spanName,
+      spanAttributes: spanInfoAttrs,
+    } = span_info ?? {};
+
     const { model: initialModel, provider: initialProvider } =
       serializeModelWithProvider(params.model);
+    const effectiveProvider = params.model?.provider || initialProvider;
```
ibolmo: maybe this means that serializeModelWithProvider should receive all of the params to keep the output consistent without the need to do the manual params.model?.provider check

ibolmo (approved):
default to approval. gotta make sure we remove that metadata line, but otherwise looks good. i see this one also tackles the other github issue on span info. 

i think that's ok to move forward, but purist in me would have wanted folks to call currentSpan to override this detail vs. adding more data to the input args. we've mentioned as much to folks before, but I agree we should minimize the user's migration pain point. 

the risk, is one day ai sdk could raise an exception for the extra args passed in. though, that's not a typical practice in js. 

don't forget to bump the version of the library

PR #1219, opened 2026-01-01
rename agents to workflows

edenh:


ankrgyl:
I assume we need to migrate data too? I think it could be challenging to actually overwrite the data because it's in the `logs` table. So we might want to do some kind of dynamic migration thing

colsondonohue (approved):

PR #1222, opened 2026-01-02
mark braintrust middleware as deprecated

cpinn:
BraintrustMiddleware was replaced as part of the wrapAiSDK re-write.

### DIFF
--- a/js/vitest.config.js
+++ b/js/vitest.config.js
@@ -50,6 +50,7 @@ const config = {
       "**/vendor/**",
       "**/.{idea,git,cache,output,temp}/**",
     ],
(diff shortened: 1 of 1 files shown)


js/vitest.config.js
```diff
@@ -50,6 +50,7 @@ const config = {
       "**/vendor/**",
       "**/.{idea,git,cache,output,temp}/**",
     ],
+    testTimeout: 15_000,
```
cpinn: We have a larger test timeout in a few different projects. This should help with the temp failures in the window builds.

ibolmo (approved):

PR #1223, opened 2026-01-02
fix hanging evals

clutchski:
This bug adds timeouts / retries to prevent network errors from hanging evals.

report from customer:

> We have  running eval jobs (500+ rows/PDF docs) on a VM that consistently freezes ~15 minutes after reaching 100% completion and gets stuck in experiment.summarize() . It occurs in fetch_base_experiment() on a POST to /api/base_experiment/get_id (https://github.com/braintrustdata/braintrust-sdk/blob/main/py/src/braintrust/logger.py#L3606C9-L3606C78). This only reproduces with large datasets on the VM and does not occur with smaller datasets or when run locally.
> Root Cause
> We think the issue is that fetch_base_experiment() uses app_conn() (Vercel IP), which is called at experiment start (registration) and then not used again until summarize(). During the long eval run, other logging traffic goes through api_conn() (AWS IP), and leaves app_conn() idle for 15+ minutes.
> Azure NAT gateways have a ~4-minute idle timeout, and would silently drop idle connections. Seems like when summarize() reuses the stale connection, the TCP session has already been removed by the NAT, leading to hangs and eventual connection failure. The customer confirmed via network capture that TCP retransmissions fail at this point, which is consistent with a stale NAT mapping.

### DIFF
--- a/py/src/braintrust/test_util.py
+++ b/py/src/braintrust/test_util.py
@@ -27,6 +27,20 @@ class TestParseEnvVarFloat:
         finally:
             del os.environ["TEST_FLOAT"]
 
+    def test_returns_default_for_inf(self):
+        os.environ["TEST_FLOAT"] = "inf"
+        try:
+            assert parse_env_var_float("TEST_FLOAT", 99.0) == 99.0
+        finally:
+            del os.environ["TEST_FLOAT"]
+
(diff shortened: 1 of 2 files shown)


manugoyal (commented):
Generally looks good. Mainly one question about the use of `.close`

PR #1227, opened 2026-01-06
Improve tags

ankrgyl:
Now that we support merging tags and tagging on subspans:

* Remove errors when you try to log on a subspan
* Update merge logic so that in-SDK merges handle tags correctly


js/util/object_util.ts
```diff
@@ -37,7 +42,18 @@ function mergeDictsWithPathsHelper({
...
+    // Check if this field should use set-union merge (e.g., "tags" at top level)
+    const isSetUnionField =
+      path.length === 0 &&
+      SET_UNION_FIELDS.has(k) &&
+      !mergePaths.has(fullPathSerialized);
+
+    if (isSetUnionField && isArray(mergeIntoV) && isArray(mergeFromV)) {
+      // Set-union merge: combine arrays, deduplicate
+      const combined = new Set([...mergeIntoV, ...mergeFromV]);
+      mergeInto[k] = Array.from(combined);
```
ankrgyl: how should we deal with objects/arrays here?

js/util/object_util.ts
```diff
@@ -66,6 +82,60 @@ export function mergeDicts(
...
+  for (const [fieldPath, valuesToRemove] of Object.entries(arrayDeletes)) {
+    if (!isArray(valuesToRemove)) {
+      continue;
+    }
+
+    // Support both top-level fields and dot-separated paths
+    const pathParts = fieldPath.split(".");
+    let current: unknown = target;
+    let parent: Record<string, unknown> | null = null;
+    let lastKey: string | null = null;
```
ankrgyl: ideally we can do exactly the same thing we do as merge paths. i think that uses arrays

js/util/object_util.ts
```diff
@@ -66,6 +91,68 @@ export function mergeDicts(
   return mergeDictsWithPaths({ mergeInto, mergeFrom, mergePaths: [] });
 }
 
+/**
+ * Mutably removes specified values from array fields in the target object.
+ * @param target - The object to modify
+ * @param arrayDeletes - An array of {path, delete} objects specifying which values to remove from which paths
+ * @example applyArrayDeletes({ tags: ["a", "b", "c"] }, [{ path: ["tags"], delete: ["b"] }]) // { tags: ["a", "c"] }
+ */
+export function applyArrayDeletes(
```
manugoyal: Currently this is only used in our internal data plane. So maybe we should move it to internal until it's actually needed by our SDKs? Then we can avoid duplicating the implementation in python for now too.

PR #1229, opened 2026-01-06
Add endpoints for interacting with env tags in ts. 

CLowbrow:
Test script:

```ts
import * as braintrust from "braintrust";
const prompt = await braintrust.loadPrompt({
  projectName: "Alex-Test-Project",
  slug: "my-prompt-433b",
});
const envs = await prompt.listPromptEnvironments();
await prompt.setPromptEnvironment({
  environmentSlug: "prod",
  version: "e8c5947221ac78b9",
});
await prompt.clearPromptEnvironment({ environmentSlug: "prod" });
```


js/src/framework2.ts
```diff
@@ -581,6 +590,24 @@ export class PromptBuilder {
 
     return prompt;
   }
+
+  public listPromptEnvironments(
```
ibolmo: curious why the double definition vs. a re-export?

js/src/prompt-environments.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  await state.login({
+    orgName,
+    apiKey,
+    appUrl,
+    fetch,
+    forceLogin,
+  });
+
+  const response = await state
```
ibolmo: missing try catch? perhaps better error message

js/src/prompt-environments.ts
```diff
@@ -0,0 +1,209 @@
...
+    .apiConn()
+    .get_json(
+      `environment-object/prompt/${promptId}`,
+      orgName ? { org_name: orgName } : {},
+    );
+
+  return (response.objects ?? []) as PromptEnvironmentAssociation[];
```
ibolmo: would this hide an upstream error?

js/src/framework2.ts
```diff
@@ -581,6 +590,24 @@ export class PromptBuilder {
...
+    options: ListPromptEnvironmentsOptions,
+  ): Promise<PromptEnvironmentAssociation[]> {
+    return listPromptEnvironments(options);
```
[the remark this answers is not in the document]
ibolmo: I'd imagine this environment class would be used in logger instead of the function based approach

js/src/prompt-environments.ts
```diff
@@ -0,0 +1,209 @@
...
+    appUrl,
+    fetch,
+    forceLogin,
+  });
+
+  // Need to use state.fetch since HTTPConnection doesn't have delete method
+  const apiConn = state.apiConn();
+  const queryString = orgName ? `?org_name=${encodeURIComponent(orgName)}` : "";
+  const url = `${apiConn.base_url}/environment-object/prompt/${promptId}/${environmentSlug}${queryString}`;
+  const resp = await state.fetch(url, {
```
ibolmo: ditto here

PR #1232, opened 2026-01-06
vitest evals

cpinn:
### Add ability to create experiments writing tests with vitest

Use existing braintrust datasets or pass their own data to test
Use existing scorers or pass their own custom scorer to tests

Each describe will create an experiment and the tests inside will each be their own span in the experiment.

See golden-ts-vitest-experiment-v* projects for some examples

Example:
```
import { initDataset } from "braintrust";

bt.describe("My experiment", () => {
  const evalData = initDataset({
    project: "llm-evals",
    dataset: "qa-benchmark",
  }).fetchedData();
 // test will also accept the dataset array and it will be expanded or pass the data individually
  bt.test.each(await evalData)(
    "Q&A evaluation",
    {
      scorers: [
        Factuality, // from autoevals
        ({ output, expected }) => ({ // custom scorer
          name: "conciseness",
          score: output.length <= expected.length * 1.2 ? 1 : 0.7,
        }),
      ],
    },
    async (record) => {
      const { input, expected } = record;
      const answer = await llm.answer(input.question);
      return answer; 
    },
  );
});
```

### DIFF
--- a/js/src/wrappers/vitest/index.ts
+++ b/js/src/wrappers/vitest/index.ts
@@ -25,18 +25,17 @@ export type { DatasetOptions, DatasetRecord } from "./dataset-helpers";
  *
  * @example Basic Usage
  * ```typescript
- * import { test, expect, describe, afterAll } from 'vitest';
+ * import * as vitest from "vitest";
  * import { wrapVitest } from 'braintrust';
  *
- * const bt = wrapVitest(
- *   { test, expect, describe, afterAll },
+ * const {test, expect, describe } = wrapVitest(
  *   { projectName: 'my-project' }
  * );
  *
- * bt.describe('Translation Tests', () => {
+ * describe('Translation Tests', () => {
  *
  *   // Tests with input/expected are automatically added to the dataset
- *   bt.test(
+ *   test(
  *     'translates hello',
  *     {
  *       input: { text: 'hello' },
@@ -51,7 +50,7 @@ export type { DatasetOptions, DatasetRecord } from "./dataset-helpers";
  *   );
  *
  *   // Tests without input/expected still run and track pass/fail
- *   bt.test('basic functionality', async () => {
+ *   test('basic functionality', async () => {
  *     const result = await someFunction();
  *     expect(result).toBeTruthy();
  *   });
@@ -60,7 +59,12 @@ export type { DatasetOptions, DatasetRecord } from "./dataset-helpers";
  *
  * @see README.md for full documentation and examples
  */
-export function wrapVitest<VitestContext = unknown, ExpectType = unknown>(
+export function wrapVitest<
+  VitestContext = unknown,
+  ExpectType extends (...args: unknown[]) => unknown = (
+    ...args: unknown[]
+  ) => unknown,
+>(
   vitestMethods: VitestMethods<VitestContext, ExpectType>,
   config: WrapperConfig = {},
 ): BraintrustVitest<VitestContext, ExpectType> {
--- a/js/src/wrappers/vitest/types.ts
+++ b/js/src/wrappers/vitest/types.ts
@@ -94,7 +94,12 @@ interface BaseWrappedDescribe {
 
 export type WrappedDescribe = WithModifiers<BaseWrappedDescribe>;
 
-export interface VitestMethods<VitestContext = unknown, ExpectType = unknown> {
+export interface VitestMethods<
+  VitestContext = unknown,
+  ExpectType extends (...args: unknown[]) => unknown = (
+    ...args: unknown[]
+  ) => unknown,
+> {
   test: TestFunction<VitestContext>;
   it?: TestFunction<VitestContext>;
   expect: ExpectType;
@@ -122,7 +127,9 @@ export interface DatasetRecord {
 
 export interface BraintrustVitest<
   VitestContext = unknown,
-  ExpectType = unknown,
+  ExpectType extends (...args: unknown[]) => unknown = (
+    ...args: unknown[]
+  ) => unknown,
 > {
   test: WrappedTest<VitestContext>;
   it: WrappedTest<VitestContext>;
--- a/js/src/wrappers/vitest/wrapper.ts
+++ b/js/src/wrappers/vitest/wrapper.ts
@@ -1,8 +1,4 @@
-import {
-  initExperiment,
-  type Experiment,
-  type ExperimentSummary,
-} from "../../logger";
+import { initExperiment, type ExperimentSummary } from "../../logger";
 import { SpanTypeAttribute } from "../../../util/index";
 import type {
   TestConfig,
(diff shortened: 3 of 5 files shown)


internal/golden/vitest-v2/vitest-experiment.eval.ts
```diff
@@ -0,0 +1,80 @@
+import { test, expect, describe, afterAll } from "vitest";
+import { wrapVitest, wrapAISDK } from "braintrust";
+import { openai } from "@ai-sdk/openai";
+import * as ai from "ai";
+
+const { generateText } = wrapAISDK(ai);
+
+// indicate the project name the tests will be sent to
+const bt = wrapVitest(
+  { test, expect, describe, afterAll },
```
ibolmo: I'd recommend for users to:

```
import * as vitest from 'vitest';

const { describe, expect, test, ... } = wrapVitest(vitest);
```

simpler, less prone to error, and future proof if we add more functions to our support

js/src/wrappers/vitest/index.ts
```diff
@@ -0,0 +1,166 @@
...
+  return {
+    test: wrappedTest,
+    it: wrappedTest,
+    expect: vitestMethods.expect,
+    describe: wrappedDescribe,
+    beforeAll: vitestMethods.beforeAll || (() => {}),
+    afterAll: vitestMethods.afterAll || (() => {}),
+    beforeEach: vitestMethods.beforeEach,
+    afterEach: vitestMethods.afterEach,
+    logOutputs: (outputs: Record<string, unknown>) => {
```
ibolmo: didn't expect this method 🤔

js/src/wrappers/vitest/index.ts
```diff
@@ -0,0 +1,166 @@
+import { currentSpan } from "../../logger";
+import {
+  wrapTest,
+  wrapDescribe,
+  getExperimentContext,
+  formatExperimentSummary,
+} from "./wrapper";
```
ibolmo: #nit I usually just keep everything in one file.

js/src/wrappers/vitest/wrapper.ts
```diff
@@ -0,0 +1,304 @@
...
+): WrappedTest<VitestContext> {
+  const wrappedTest = function (
+    name: string,
+    configOrFn: TestConfig | ((context: VitestContext) => void | Promise<void>),
+    maybeFn?: (context: TestContext & VitestContext) => void | Promise<void>,
+  ) {
+    const isEnhanced = typeof configOrFn !== "function";
+    const testConfig = isEnhanced ? configOrFn : undefined;
+
+    return originalTest(name, async (vitestContext: VitestContext) => {
```
ibolmo: i'd expect to pass the test config to the originalTest.. perhaps without our bt options.

js/src/wrappers/vitest/wrapper.ts
```diff
@@ -0,0 +1,304 @@
...
+        const { input, expected, metadata, tags } = testConfig;
+        if (input !== undefined || expected !== undefined) {
+          // Add to dataset (will not duplicate if already exists)
+          const exampleId = experimentContext.dataset.insert({
+            input,
+            expected,
+            metadata,
+            tags,
+          });
+          experimentContext.datasetExamples.set(name, exampleId);
```
ibolmo: i think langsmith requires users to have datasets for the experiments to work, but we don't. i wonder if it would be better if we didn't insert datasets and just passed the test context (input/expectec, etc.) directly in the log. the user could move the data to a dataset if they wanted to.

js/src/wrappers/vitest/wrapper.ts
```diff
@@ -0,0 +1,304 @@
...
+
+            // Automatically log pass feedback on success
+            span.log({
+              scores: {
+                pass: 1,
+              },
+            });
+
+            // If test function returns a value, log it as output
+            if (testResult !== undefined) {
```
ibolmo: i think if you `traced(maybeFn || configOrFn, ...)` you may have gotten this automatically?

js/src/wrappers/vitest/wrapper.ts
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// Context for experiment mode - stored per describe block
+interface ExperimentContext {
+  dataset: Dataset<false>;
+  experiment: Experiment;
+  datasetExamples: Map<string, string>; // test name -> example id
+}
+
+// Global context holder (one per describe block)
```
ibolmo: what happens with concurrent calls i.e.

```
it.concurrent(
 describe(..., () => {
 })
);

it.concurrent(
 describe(..., () => {
 })
);
```

did you give `currentExperiment` a try?

ibolmo (commented):
various questions

js/src/wrappers/vitest/types.ts
```diff
@@ -0,0 +1,113 @@
+import type { Span } from "../../logger";
+
+export interface TestConfig {
```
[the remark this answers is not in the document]
cpinn: I am not quite sure I understand this type concern. 

I had to create new types because of the new allowance of passing of input and scores which fundamentally changes the structure? maybe I don't understand what the idea is.

js/src/wrappers/vitest/index.ts
cpinn: This was a method in the original inspiration as a way to add additional data. I thought it could be useful.

internal/golden/vitest-v2/vitest-experiment.eval.ts
```diff
@@ -0,0 +1,78 @@
...
+    {
+      input: { query: "What is 2+2?" },
+      expected: 4,
+      metadata: { category: "math" },
+      tags: ["arithmetic"],
+    },
+    async ({ input, expected }) => {
+      const result = 4;
+      logOutputs({ answer: result });
+      expect(result).toBe(expected);
```
ibolmo: are the expects scorers?

js/src/wrappers/vitest/index.ts
```diff
@@ -0,0 +1,168 @@
...
+  const wrappedDescribe = wrapDescribe(
+    vitestMethods.describe,
+    config,
+    vitestMethods.afterAll,
+  );
+
+  return {
+    test: wrappedTest,
+    it: wrappedTest,
+    expect: vitestMethods.expect,
```
ibolmo: ah so we don't wrap expects. 

like i mentioned earlier, it would be interesting to use the expect(..., 'message')... and the message is the index for the output, or maybe we have a counter/stack that we push expect(output) to then the output might be something like 

```ts
expect(0).toBe(0);
expect('something', 'message').toBe('something');
expect('foo').toBe('bar');
```

then the event could be

```ts
{
  ...
  output: {
    0: output,
    'message': output,
    1: output,
  }, 
  scores: {
    0: 1,
    'message': 1,
    1: 0
  }
 }
}
```

ibolmo (commented):
small suggestions not blocking

js/src/wrappers/vitest/index.ts
cpinn: did a compromise and logged named outputs automatically. Users can still log additional outputs as necessary. Will see what the feedback is like.

PR #1235, opened 2026-01-07
verify we use configured org in links before login.

clutchski:
Our links require an org name, which either needs to be configured via initLogger or env var or obtained via login. This change ensures typescript will use the configured org name (if it exists) before login has finished. 

Fixes BRA-3862

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "braintrust",
-  "version": "2.1.0",
+  "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "braintrust",
-  "version": "2.0.0",
+  "version": "2.1.0",
```
ibolmo: I would also expect your version to be a patch change since it is backward compatible if the user doesn't supply this

PR #1242, opened 2026-01-07
Update js build to run for otel-js

cpinn:
run the js build and therefore the smoke tests when changes are made in otel-js


ibolmo:
i also wonder if the otel-js should do a pnpm build --filter="braintrust" to force the js build

PR #1247, opened 2026-01-08
Handle arrays of images when building prompts

cpinn:
Users would like to be able to reference an array when handling images.

This PR introduces the following changes:

Variables that contain arrays of images or files will now be automatically expanded when used everything is a link or an attachment. Backend changes attachment types to links, if everything goes right you have a full array of links, attachment references could be in the variable if links are not found, unlikely in practice. 

See attachment link resolution tests here for link resolution: https://github.com/braintrustdata/braintrust/pull/9983/changes

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -5527,7 +5527,21 @@ export class ObjectFetcher<RecordType> implements AsyncIterable<
   ): AsyncGenerator<WithTransactionId<RecordType>> {
     const state = await this.getState();
     const objectId = await this.id;
-    const limit = batchSize ?? DEFAULT_FETCH_BATCH_SIZE;
+    const batchLimit = batchSize ?? DEFAULT_FETCH_BATCH_SIZE;
+    const internalLimit = (
+      this._internal_btql as { limit?: number } | undefined
+    )?.limit;
+    const limit =
+      batchSize !== undefined ? batchSize : (internalLimit ?? batchLimit);
+    const internalBtqlWithoutReservedQueryKeys = Object.fromEntries(
+      Object.entries(this._internal_btql ?? {}).filter(
+        ([key]) =>
+          key !== "cursor" &&
+          key !== "limit" &&
+          key !== "select" &&
+          key !== "from",
+      ),
+    );
     let cursor = undefined;
     let iterations = 0;
     while (true) {
@@ -5535,7 +5549,6 @@ export class ObjectFetcher<RecordType> implements AsyncIterable<
         `btql`,
         {
           query: {
-            ...this._internal_btql,
             select: [
               {
                 op: "star",
@@ -5556,6 +5569,7 @@ export class ObjectFetcher<RecordType> implements AsyncIterable<
             },
             cursor,
             limit,
+            ...internalBtqlWithoutReservedQueryKeys,
           },
           use_columnstore: false,
           brainstore_realtime: true,
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -19,3 +19,31 @@ jobs:
         run: pnpm install --frozen-lockfile
       - name: Lint
         run: pnpm run lint
+
+  check-typings:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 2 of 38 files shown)


js/src/template/nunjucks-env.ts
```diff
@@ -1,12 +1,72 @@
...
 const createNunjucksEnv = (throwOnUndefined: boolean): NunjucksEnvironment => {
-  return new nunjucks.Environment(null, {
-    autoescape: true,
+  const env = new nunjucks.Environment(null, {
+    autoescape: false,
     throwOnUndefined,
   });
+
+  // Intercept renderString to wrap objects with custom toString
+  const originalRenderString = env.renderString.bind(env);
```
ibolmo: nit could have used a Proxy

js/src/template/nunjucks-utils.ts
```diff
@@ -2,7 +2,7 @@ import * as nunjucks from "nunjucks";
 
 export function lintTemplate(template: string, context: any): void {
   const env = new nunjucks.Environment(null, {
```
ibolmo: I'd expected to see a getNunjunksEnv here

js/src/template/renderer.test.ts
```diff
@@ -82,4 +82,187 @@ describe("renderTemplateContent", () => {
...
+    const result = renderTemplateContent(
+      "User: {{user}}, Items: {{items}}",
+      varsWithObject,
+      escape,
+      { templateFormat: "mustache" },
+    );
+    expect(result).toBe(
+      'User: {"name":"Alice","age":30}, Items: ["a","b","c"]',
+    );
+    expect(result).not.toContain("[object Object]");
```
ibolmo: i'd encourage a literal comparison, if possible. the not.toContain could be misleading since any value is allowed except `"[object ...`

js/src/template/renderer.test.ts
```diff
@@ -82,4 +82,187 @@ describe("renderTemplateContent", () => {
...
+    };
+    // Nunjucks allows property access on objects
+    const resultWithAccess = renderTemplateContent(
+      "User: {{ user.name }}, Age: {{ user.age }}, Items: {{ items }}",
+      varsWithObject,
+      escape,
+      { templateFormat: "nunjucks" },
+    );
+    // Arrays output as comma-separated (Nunjucks default behavior)
+    expect(resultWithAccess).toBe("User: Bob, Age: 25, Items: 1,2,3");
```
ibolmo: i expected the array to be ["1", "2", "3"] 🤔 (json stringify)

js/src/template/nunjucks-render.test.ts
```diff
@@ -1,19 +1,14 @@
...
 import { lintTemplate } from "./nunjucks-utils";
 import { nunjucks } from "./nunjucks";
+import { getNunjucksEnv } from "./nunjucks-env";
 
 function getEnv() {
-  return new nunjucks.Environment(null, {
-    autoescape: true,
-    throwOnUndefined: false,
-  });
+  return getNunjucksEnv(false);
```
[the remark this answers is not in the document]
cpinn: I believe I resolved this.

js/src/template/renderer.test.ts
ibolmo: looks like mustache does do as I'd expected. 🤔

js/src/logger.ts
```diff
@@ -6635,9 +6648,67 @@ export type DefaultPromptArgs = Partial<
...
+  const varPath = match[1];
+
+  const value = varPath.includes(".")
+    ? getObjValueByPath(variables, varPath.split("."))
+    : variables[varPath];
+
+  if (!Array.isArray(value)) return null;
+
+  // Check if array contains attachments or image URLs (supports mixed arrays)
+  const allValid = value.every(
```
cpinn: The flow is that the backend will attempt to turn all the references into URLs. Now if something goes wrong and one of the attachments isn't turned into a URL this would still allow expansion by ensuring the other values are attachments.

A different flow I experimented with was passing the attachment map from the backend for verification. I did not like that mixing of concerns though so I did not think it was a good idea to introduce that logic here.

Qard (approved):

ibolmo (approved):
LGTM!

PR #1249, opened 2026-01-09
capture ttft metric for gemini instrumentation

realark:


realark:
talked w/ some folks offline. we decided we'll only capture TTFT for streaming requests

PR #1250, opened 2026-01-09
support prompt caching token tracking in langchain

colinbennettbrain:
- Extract `cache_read` and `cache_creation` from LangChain's nested `input_token_details` object
 - Map to Braintrust's standard metric names (`prompt_cached_tokens`, `prompt_cache_creation_tokens`)

  ## Problem
  The LangChain callback handler only extracts top-level fields from `usage_metadata` (input_tokens, output_tokens, total_tokens) but not the nested `input_token_details` containing cache metrics. This causes:
  - Cache tokens showing as 0 in the experiments UI
  - Costs calculated at full uncached rate instead of discounted cache rate

### DIFF
--- a/integrations/langchain-py/src/tests/test_callbacks.py
+++ b/integrations/langchain-py/src/tests/test_callbacks.py
@@ -1128,8 +1128,7 @@ Remember: Testing is not just about finding bugs, it's about building confidence
 
     assert "prompt_cache_creation_tokens" in first_metrics
     assert first_metrics["prompt_cache_creation_tokens"] > 0
-    if "prompt_cached_tokens" in first_metrics:
-        assert first_metrics["prompt_cached_tokens"] == 0
+    assert first_metrics["prompt_cached_tokens"] == 0
 
     res = model.invoke(
         messages + [res, HumanMessage(content="What testing framework is mentioned for Python?")],


cpinn (commented):
This change looks great to me with the added test, was what I was going to take a look into today. Usually for changes like this we also record and save a cassette to have the tests run the tests a little faster. 

cc: @ibolmo should we add a cassette to the change or should we merge as is and add it later?

integrations/langchain-py/src/tests/test_callbacks.py
```diff
@@ -149,6 +151,70 @@ def test_llm_calls(logger_memory_logger: LoggerMemoryLogger):
     )
 
 
+def test_cache_token_metrics_from_input_token_details(logger_memory_logger: LoggerMemoryLogger):
+    logger, memory_logger = logger_memory_logger
+    assert not memory_logger.pop()
+
+    handler = BraintrustCallbackHandler(logger=logger)
+
+    class CachingChatModel(BaseChatModel):
```
ibolmo: let me give this a try. looks like the ai is hallucinating what to do

integrations/langchain-py/src/tests/test_callbacks.py
```diff
@@ -949,3 +949,208 @@ def test_streaming_ttft(logger_memory_logger: LoggerMemoryLogger):
...
+    first_span = llm_spans[0]
+
+    assert "metrics" in first_span
+    first_metrics = first_span["metrics"]
+    assert "prompt_tokens" in first_metrics
+    assert first_metrics["prompt_tokens"] > 0
+
+    assert "prompt_cache_creation_tokens" in first_metrics
+    assert first_metrics["prompt_cache_creation_tokens"] > 0
+    if "prompt_cached_tokens" in first_metrics:
```
cpinn: nit. do we want this if statement in the test because then the assertion may or may not run?

Issue #1251, opened 2026-01-09
Vercel AI SDK Integration not including structured output


jczstudios:
Using Braintrust Node SDK `2.0.1` and Vercel AI SDK `6.0.25`.

When using `wrapAISDK`, Braintrust is not recording the output schema.

Reproduction:
```js

const { streamText } = wrapAISDK(ai)

const schema = z.object({
    scratchpad: z.string()
})

const aiSchema = ai.Output.object({
    schema
})

const response = streamText({
    ...(compiledPrompt as any),
    model: llm,
    output: aiSchema
})
```

The span data in Braintrust shows schema as null:

<img width="530" height="402" alt="Image" src="https://github.com/user-attachments/assets/adb662a6-0ebd-4864-96b7-e8046c6cf06b" />

ibolmo:
thank you for the repro steps! will have a look

ibolmo:
this will go out with v2.0.2 will should go out soon

PR #1255, opened 2026-01-10
ai-sdk: fix input output objects serialization and cover v5 and v6 testing

ibolmo:
the new Output.object etc changed from v5 to v6. new tests harness makes sure we are testing both versions. looks like v3 and v4 do not have the output schema objects. 

fixes: #1251


Qard (approved):

manugoyal (approved):
🙈

PR #1258, opened 2026-01-13
Add a test running the sdk in a browser 

cpinn:
There are use cases where Braintrust needs to run in a browser environment, but we currently do not have any tests that specifically validate this runtime. Browser specific issues could go undetected.

This PR introduces browser coverage in two ways:
1. It adds a new evaluation test that runs as part of all integration tests.
2. It sets up a browser-based environment that runs the existing test suite, ensuring Braintrust works correctly in a browser context.

#### Browser Test
The browser test suite bundles Braintrust dependencies and the test scripts into a single file, which is loaded into an HTML page and executed in the browser via Playwright. The test waits for the results to be exposed on the window object and then reports the run as passed or failed based on those results.

### DIFF
--- a/js/smoke/tests/browser/README.md
+++ b/js/smoke/tests/browser/README.md
@@ -4,16 +4,7 @@ This package contains smoke tests for the Braintrust SDK running in a browser en
 
 ## What It Tests
 
-- **General SDK Tests**: Uses shared test suites to verify:
-
-  - Export verification (all expected exports are available)
-  - Basic logging functionality (initLogger, spans, flush)
-  - Multiple span creation and logging
-
-- **Eval Tests**: Tests the `Eval` function in browser environment:
-  - Running evaluations with test data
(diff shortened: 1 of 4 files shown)


js/smoke/tests/browser/pages/browser-tests.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Braintrust SDK Browser Tests</title>
+  </head>
+  <body>
+    <h1>Braintrust SDK Browser Tests</h1>
+    <div id="output"></div>
+    <script type="module" src="/dist/browser-tests.js"></script>
```
cpinn: for simplicity everything is just currently in this script

PR #1261, opened 2026-01-13
braintrust/temporal

clutchski:
This PR adds braintrust tracing to temporal.

A new package @braintrust/temporal has been added that handles tracing temporal workflows.

### DIFF
--- a/integrations/temporal-js/examples/temporal/mise.toml
+++ b/integrations/temporal-js/examples/temporal/mise.toml
@@ -16,7 +16,7 @@ postinstall = "mise run install"
 
 [tasks.install]
 description = "Install dependencies"
-run = "pnpm install"
+run = "pnpm install --ignore-workspace"
 
 [tasks.server]
 description = "Run temporal server and workers"


js/package.json
```diff
@@ -142,6 +153,28 @@
...
   },
   "peerDependencies": {
+    "@temporalio/activity": "^1.11.0",
+    "@temporalio/client": "^1.11.0",
+    "@temporalio/common": "^1.11.0",
+    "@temporalio/worker": "^1.11.0",
+    "@temporalio/workflow": "^1.11.0",
     "zod": "^3.25.34 || ^4.0"
+  },
+  "peerDependenciesMeta": {
```
Qard: Yarn does not support `peerDependenciesMeta`. It will only yell at you about unsatisfied peer dependencies though but will continue installation anyway. Slight DX issue, but still _works_.

Qard:
Because it has `peerDependenciesMeta` with `optional: true` it is _not_ required to install temporal together with the SDK. Yarn, however, does not support this--only npm and pnpm do--but it doesn't actually enforce peer dependencies at all anyway, it just warns you a peer dependency is missing but then completes the install anyway.

The DX sucks a bit as it will look like it is expecting temporal to be installed when it actually doesn't need it. But users can safely ignore that warning.

js/smoke/tests/otel-v1/tsconfig.json
```diff
@@ -1,17 +1,17 @@
...
-      "forceConsistentCasingInFileNames": true,
-      "skipLibCheck": true,
-      "resolveJsonModule": true,
-      "isolatedModules": true,
-      "types": ["node"]
-    },
-    "include": ["otel_example.ts"],
-    "exclude": ["node_modules"]
-  }
+  "compilerOptions": {
```
ibolmo: #nit i wonder why your setup did this

js/src/wrappers/temporal/index.ts
```diff
@@ -0,0 +1,57 @@
...
+ * This module provides a plugin that automatically creates Braintrust spans
+ * for Temporal workflows and activities, with proper parent-child relationships
+ * across distributed workers.
+ *
+ * @example
+ * ```typescript
+ * import { Client, Connection } from "@temporalio/client";
+ * import { Worker } from "@temporalio/worker";
+ * import * as braintrust from "braintrust";
+ * import { createBraintrustTemporalPlugin } from "braintrust/wrappers/temporal";
```
ibolmo: #nit example needs an update to use `braintrust.`

ibolmo (commented):
going to stop my review since I did find a non-starter.

js/src/wrappers/temporal/interceptors.ts
```diff
@@ -0,0 +1,176 @@
+import type { Context } from "@temporalio/activity";
+import type {
+  ActivityInboundCallsInterceptor,
+  ActivityExecuteInput,
+  Next,
+  ActivityInterceptors,
+} from "@temporalio/worker";
+import type { WorkflowClientInterceptor } from "@temporalio/client";
+import { defaultPayloadConverter } from "@temporalio/common";
```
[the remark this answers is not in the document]
Qard: Ah, right. Yes. Bundling woes. 🙈 

This will indeed fail. Though the peer dependency thing is _fine_ (ish), having imports of these _anywhere_ that gets included in the bundle will indeed make it required as a proper dependency.

js/smoke/tests/otel-v1/tsconfig.json
clutchski: is it prettier that did this?

PR #1265, opened 2026-01-14
Add braintrust.auto_instrument() for one-line auto-instrumentation

clutchski:
Adds a new auto_instrument() function that detects installed AI/ML libraries and automatically instruments them for tracing.

Supported integrations:
- OpenAI (patch_openai/unpatch_openai)
- Anthropic (patch_anthropic/unpatch_anthropic)
- LiteLLM (patch_litellm/unpatch_litellm)
- Pydantic AI (setup_pydantic_ai)
- Google GenAI (setup_genai)
- Agno (setup_agno)
- Claude Agent SDK (setup_claude_agent_sdk)
- DSPy (patch_dspy/unpatch_dspy)

Usage:

```python
import braintrust
braintrust.auto_instrument()
```

### DIFF
--- a/py/src/braintrust/wrappers/test_utils.py
+++ b/py/src/braintrust/wrappers/test_utils.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 import sys
 import textwrap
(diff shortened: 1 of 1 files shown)


py/src/braintrust/auto.py
```diff
@@ -0,0 +1,259 @@
...
+        # Your code now automatically logs to Braintrust
+        import openai
+        client = openai.OpenAI()
+        response = client.chat.completions.create(...)  # Auto-traced!
+        ```
+    """
+    results = {}
+
+    if openai:
+        results["openai"] = _instrument_openai()
```
ibolmo: i'd rather have each register itself with the library, and then have this iterate through registered instrumentations. This allows 3rd (or first) party libraries to register themselves as well.

py/src/braintrust/wrappers/claude_agent_sdk/__init__.py
```diff
@@ -105,7 +105,6 @@ def setup_claude_agent_sdk(
                         setattr(module, "tool", wrapped_tool_fn)
 
         return True
-    except ImportError as e:
-        logger.error(f"Failed to import Claude Agent SDK: {e}")
-        logger.error("claude-agent-sdk is not installed. Please install it with: pip install claude-agent-sdk")
+    except ImportError:
```
knjiang: i dont see tests for claude_agent_sdk and others, i think we can probably add them in as well for `test_auto`.   
i'm wnodering if there's an a way we can also enforce testing auto instrumentation?

py/src/braintrust/auto.py
```diff
@@ -0,0 +1,259 @@
...
+
+    return results
+
+
+def _instrument_openai() -> bool:
+    """Instrument OpenAI if available."""
+    try:
+        from braintrust.oai import patch_openai
+
+        patch_openai()
```
manugoyal: Looks like `patch_openai` swallows import errors, so if we're trying to detect
whether the patch succeeded, maybe `patch_openai` should return the boolean
directly?

PR #1268, opened 2026-01-14
fix(python): add org_id to prompt cache key for cross-org isolation

colinbennettbrain:
- Adds `org_id` to cache key format for proper organization isolation
- Adds tests for cross-org cache behavior

The prompt cache was using a key format of `{project_name}:{slug}:{version}` which did not include the organization ID. This caused a cross-org cache collision bug where two organizations with the same project name and prompt slug could retrieve each other's cached prompts.

## Test plan

- [x] Added `test_handle_different_orgs_with_same_project_and_slug` - verifies prompts from different orgs with same project/slug are isolated
- [x] Added `test_org_id_isolation_with_disk_cache` - verifies isolation works after memory eviction (via disk cache)
- [x] All 15 prompt cache tests pass

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -1678,13 +1678,14 @@ def load_prompt(
             eprint(f"Failed to load prompt, attempting to fall back to cache: {server_error}")
             try:
                 if id:
-                    return _state._prompt_cache.get(id=id)
+                    return _state._prompt_cache.get(id=id, org_id=_state.org_id)
                 else:
                     return _state._prompt_cache.get(
                         slug,
                         version=str(version) if version else "latest",
                         project_id=project_id,
                         project_name=project,
+                        org_id=_state.org_id,
                     )
             except Exception as cache_error:
                 if id:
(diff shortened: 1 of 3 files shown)


colinbennettbrain:
Hey Olmo, this came up because a customer realized the wrong prompt was getting sent to OpenAI.  I am attempting to fix the issue with this.  I also have a repro script if that's helpful.

py/src/braintrust/logger.py
```diff
@@ -1714,6 +1715,7 @@ def compute_metadata():
                 _state._prompt_cache.set(
                     prompt,
                     id=id,
+                    org_id=_state.org_id,
```
ibolmo: same here combine with the next statement

py/src/braintrust/prompt_cache/prompt_cache.py
```diff
@@ -15,22 +15,32 @@
...
         return f"id:{id}"
 
     prefix = project_id or project_name
     if not prefix:
         raise ValueError("Either project_id or project_name must be provided")
     if not slug:
         raise ValueError("Slug must be provided when not using ID")
+
+    # Include org_id in cache key if available to ensure cross-org isolation
+    if org_id:
```
ibolmo: looks like login is called L1696 before updating/setting the prompt. you should be able to assume org_id exists .. perhaps raise if not set

PR #1270, opened 2026-01-15
Fix the release ci job

ibolmo:


py/src/braintrust/test_framework.py
```diff
@@ -241,7 +241,6 @@ def task_with_hooks(input_value: int, hooks: EvalHooks) -> int:
     assert sorted(input_2_trials) == [0, 1]
 
 
-@pytest.mark.vcr
 @pytest.mark.asyncio
```
ibolmo: reverts just this change: https://github.com/braintrustdata/braintrust-sdk/commit/a751e399b66d396cdf13599c93b1445d81934ddc#diff-04ffd10c18b19c5e639ba456d01eaebccdcf130e657c0407b7cd8be40359762dL243

PR #1272, opened 2026-01-15
add a force option to force tag replacement

ibolmo:


py/scripts/push-release-tag.sh
```diff
@@ -44,6 +55,11 @@ echo "commit:       ${COMMIT}"
 echo "code:         ${REPO_URL}/commit/${COMMIT}"
 echo "changeset:    ${REPO_URL}/compare/${LAST_RELEASE}...${COMMIT}"
 
+if [ "$FORCE" = true ]; then
+  echo ""
+  echo "⚠️  WARNING: Force mode enabled - will overwrite existing tag if present"
```
clutchski: I'm not sure we need this warning? the user is opting in?

clutchski (approved):
I think it would be nice to assert that no pypi build exists with that version before replacing.

PR #1274, opened 2026-01-15
Braintrust 3.0.0 version

cpinn:
## Braintrust 3.0.0 Version

This new version separates out nunjucks templating into a new package and creates a new separate package just for a braintrust browser version.

Cloudflare Workers were encountering compatibility issues with certain dependencies in the Braintrust SDK. These issues surface particularly in restrictive runtime environments and how some dependencies behave in Vite-based builds.

### Changes
- Add additional test coverage for cloudflare workers:
   - braintrust, braintrust/browser and nodejs_compat flag
   - vite cloudflare test
- introduce browser specific configuration to drop nunjucks template support in browser
- introduce build verification and improved auto selection via node exports for different environments
- introduces the braintrust/browser package that includes the additional [als-browser ](https://github.com/apm-js-collab/als-browser) package for AsyncLocalStorage support in browsers
- creates additional test structure where integrations can have their own smoke tests that integrate with the main build.


js/smoke/tests/cloudflare-worker/src/browser.ts
```diff
@@ -0,0 +1,4 @@
+import * as braintrust from "braintrust/browser";
+import { createWorker } from "./worker";
+
+export default createWorker(braintrust, "cloudflare-worker-browser-no-compat");
```
ibolmo: duplicate?

js/smoke/tests/cloudflare-worker/src/worker.ts
```diff
@@ -0,0 +1,116 @@
...
+  fetch(request: Request, _env: Env): Promise<Response>;
+} {
+  const { initLogger, _exportsForTestingOnly } = braintrust;
+
+  async function runSharedTestSuites(): Promise<TestResponse> {
+    try {
+      const adapters = await setupTestEnvironment({
+        initLogger,
+        testingExports: _exportsForTestingOnly,
+        canUseFileSystem: false,
```
ibolmo: looks like this should be true? https://developers.cloudflare.com/workers/runtime-apis/nodejs/fs/

js/smoke/tests/deno/browser/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "braintrust-deno-browser-smoke-test",
+  "version": "1.0.0",
+  "description": "Smoke test for Braintrust browser build in Deno runtime",
+  "private": true,
+  "scripts": {
+    "test": "deno test --no-config --import-map=./deno.json --lock=deno.lock --allow-env --allow-read=.,../build --allow-net shared_suite_test.ts"
```
ibolmo: curious what this package is doing, if anything

js/smoke/tests/deno/browser/deno.json
```diff
@@ -0,0 +1,11 @@
+{
+  "imports": {
+    "braintrust/browser": "../build/braintrust/dist/browser.mjs",
+    "uuid": "npm:uuid@^13.0.0",
```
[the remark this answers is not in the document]
ibolmo: curious why we're not seeing the other maps for other npm libraries we have, for example

js/smoke/tests/deno/node/deno.json
```diff
@@ -0,0 +1,14 @@
+{
+  "imports": {
+    "braintrust": "../build/braintrust/dist/index.mjs",
+    "uuid": "npm:uuid@^13.0.0",
+    "zod/v3": "npm:zod@^3.24.1",
+    "zod/v4": "npm:zod@^4.1.12",
+    "nunjucks": "npm:nunjucks@^3.2.4",
+    "simple-git": "npm:simple-git@^3.21.0"
+  },
+  "tasks": {
```
ibolmo: curious why these tasks differ compared to the other deno.json

js/smoke/tests/vite-react-hono/src/worker/index.ts
```diff
@@ -11,12 +12,13 @@ import {
   runImportVerificationTests,
   runPromptTemplatingTests,
   type TestResult,
-} from "../../../shared/dist/index.mjs";
+} from "../../../../shared/dist/index.mjs";
 
-import * as braintrust from "braintrust";
+import * as braintrust from "braintrust/browser";
```
ibolmo: we might need two versions. one that is browser and one that is node.

js/smoke/tests/vite-react-hono/wrangler.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "node_modules/wrangler/config-schema.json",
+  "name": "braintrust-vite-react-hono-smoke-test",
+  "main": "./src/worker/index.ts",
+  "compatibility_date": "2025-01-15",
+  "compatibility_flags": ["nodejs_compat"]
```
ibolmo: same here did we want to try without compat? maybe the same results

js/src/browser-config.ts
```diff
@@ -30,6 +30,12 @@ export function configureBrowser() {
     return process.env[name];
   };
 
+  iso.renderNunjucksString = () => {
```
ibolmo: did we still need nunjucks-env.browser.ts and nunjucks-utils.browser.ts?

js/smoke/tests/deno/browser/deno.json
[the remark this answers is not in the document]
cpinn: We aren't seeing the other maps because I added them to `noExternal`. I am not sold on if we want to do that or not but it was able to remove some of the imports for deno, it did increase the build size a bit but cut down on the dependencies required in deno environments.

I didn't add uuid, simple-git or nunjucks though because those packages internally use node functions (ie like child-process for simple-git). Those internal functions aren't wrapped by the new eslint-plugin-node-import so those packages end up being required in the deno import map (unless we do some special tsup things TM).

js/smoke/tests/deno/node/deno.json
```diff
@@ -0,0 +1,14 @@
+{
+  "imports": {
+    "braintrust": "../build/braintrust/dist/index.mjs",
+    "uuid": "npm:uuid@^13.0.0",
+    "zod/v3": "npm:zod@^3.24.1",
+    "zod/v4": "npm:zod@^4.1.12",
+    "nunjucks": "npm:nunjucks@^3.2.4",
+    "simple-git": "npm:simple-git@^3.21.0"
```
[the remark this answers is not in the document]
cpinn: see here: https://github.com/braintrustdata/braintrust-sdk/pull/1274#discussion_r2700528337

Also the internal message in the slack thread on noExternal

js/smoke/tests/cloudflare-worker/run-test.mjs
```diff
@@ -1,11 +1,83 @@
...
+    nodejsCompat: true,
+  },
+  "node-no-compat": {
+    config: "wrangler.node-no-compat.toml",
+    entrypoint: ENTRYPOINT.node,
+    nodejsCompat: false,
+  },
+};
+
+function shouldExpectStartupFailure(variant) {
```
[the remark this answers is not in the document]
cpinn: #nit the last time I made scenarios fail in the build pipeline people didn't like it

js/smoke/tests/vite-react-hono/src/worker/index.ts
cpinn: The node one will always fail. Nunjucks and vite do not work together. if you use vite you need to use browser.

js/src/browser-config.ts
cpinn: no they were from some issue I thought I was running into with the deno build and I had been waiting to see if this build finally passed.

js/smoke/tests/vite-react-hono/wrangler.json
cpinn: we can. I did that in the other tests. 

I do wonder how useful it is without the compat in general since then you lose all the hierarchy. I think that would be a pretty annoying span experience.

js/smoke/tests/vite-react-hono/src/worker/index.ts
cpinn: see: https://github.com/braintrustdata/braintrust-sdk/actions/runs/21075727902/job/60617791890

js/smoke/tests/deno/browser/package.json
cpinn: Currently it is needed as part of run-scripts for "test discovery". I think there is something that could be done to discover differently for deno but at the moment that was why there appeared to be two.

cpinn:
okay I removed a couple of the dead files and removed the noExternal change I had made after we made the deno test changes. 

once smoke-v2 is merged in we can update the compatibility dates for the cloudflare tests

js/src/browser-config.ts
cpinn: removed.

js/tsup.config.ts
```diff
@@ -19,7 +22,15 @@ export default defineConfig([
...
     outDir: "dist",
+    removeNodeProtocol: false,
     external: ["zod"],
+    noExternal: [
+      "mustache",
+      "eventsource-parser",
+      "zod-to-json-schema",
+      "@vercel/functions",
+    ],
+    platform: "browser",
```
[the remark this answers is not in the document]
cpinn: resolved in newer commit

js/smoke/scenarios/deno-node/deno.json
```diff
@@ -2,7 +2,7 @@
   "imports": {
     "@std/assert": "jsr:@std/assert@^1.0.14",
     "@braintrust/smoke-test-shared": "jsr:@braintrust/smoke-test-shared",
-    "braintrust": "npm:braintrust@^2.0.2"
+    "braintrust": "npm:braintrust"
```
ibolmo: avoid us having to update this

js/smoke/scenarios/deno-node/Makefile
```diff
@@ -4,17 +4,15 @@ setup:
...
+	if [ -n "$(BRAINTRUST_TAR)" ]; then \
+		echo "==> Extracting SDK tarball for workspace link"; \
+		TARBALL_PATH="$$(cd ../.. && pwd)/$(BRAINTRUST_TAR)"; \
+		echo "==> Resolved tarball path: $$TARBALL_PATH"; \
+		cd ../../.. && tar -xzf "$$TARBALL_PATH" --strip-components=1 package/dist; \
+	else \
+		echo "==> Building SDK for workspace link"; \
+		cd ../../.. && pnpm exec turbo build --filter=braintrust; \
+	fi; \
 	deno install
```
ibolmo: `deno install` seems to ignore the `links` directive and tries to grab from npm.

js/src/template/registry.ts
```diff
@@ -0,0 +1,161 @@
...
+ * }
+ * ```
+ */
+export const getTemplateRenderer = templateRegistry.get.bind(templateRegistry);
+
+// Built-in mustache plugin
+const jsonEscape = (v: unknown) =>
+  typeof v === "string" ? v : JSON.stringify(v);
+
+const mustachePlugin: TemplateRendererPlugin = {
```
ibolmo: #nit

would have expected this in a separate file like `template/plugins/mustache.ts`

js/src/browser.ts
```diff
@@ -1,7 +1,16 @@
...
+ *   import * as braintrust from '@braintrust/browser';
+ */
+
 import { configureBrowser } from "./browser-config";
 
 configureBrowser();
 
-// eslint-disable-next-line no-restricted-syntax -- already enforced in exports
 export * from "./exports";
-export * as default from "./exports";
```
ibolmo: we don't need for the default export?

js/src/browser.ts
```diff
@@ -1,7 +1,16 @@
...
+ * For optimal browser support with AsyncLocalStorage polyfill, consider:
+ *   npm install @braintrust/browser
+ *   import * as braintrust from '@braintrust/browser';
+ */
+
 import { configureBrowser } from "./browser-config";
 
 configureBrowser();
 
-// eslint-disable-next-line no-restricted-syntax -- already enforced in exports
```
ibolmo: huh i wonder why we don't need the no-restricted-syntax (the custom eslint) i added

js/src/logger.ts
```diff
@@ -4396,8 +4399,12 @@ export async function login(
     return _globalState;
   }
 
+  if (!_globalState) {
+    _internalSetInitialState();
```
ibolmo: i think i was expecting this to be done by a configure* method?

js/src/node.ts
```diff
@@ -9,16 +9,19 @@ import iso from "./isomorph";
 import { getRepoInfo, getPastNAncestors } from "./gitutil";
 import { getCallerLocation } from "./stackutil";
 import { _internalSetInitialState } from "./logger";
-import { promisify } from "util";
-import * as zlib from "zlib";
+import { promisify } from "node:util";
```
ibolmo: something to consider for the future (or perhaps the ai is quick to do this).

i previously added a lint rule to disallow import from cli in the sdk/js. 

what if we had a sdk/js/node and sdk/js/browser ... we then move files in two each of those. we could say that nothing is allowed to depend on sdk/js/node/** unless the it stems from node.ts, and same with browser.ts

integrations/browser-js/README.md
```diff
@@ -0,0 +1,75 @@
...
+- Evaluation framework
+- OpenTelemetry integration
+
+## Documentation
+
+For full documentation, visit [https://www.braintrust.dev/docs](https://www.braintrust.dev/docs)
+
+## License
+
+Apache-2.0
```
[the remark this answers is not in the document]
cpinn: This is actually something I have been meaning to ask about. 

We keep declaring our license as MIT but our license is actually Apache 2.0 (so it would overwrite and publish as apache 2.0). I have been pretty confused which license we were intending to go with. 

See actual license used during publishing to npm: https://github.com/braintrustdata/braintrust-sdk/blob/main/LICENSE

* I got pretty confused when I was publishing and noticing a totally different license was being included.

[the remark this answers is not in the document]
ibolmo: ah no you're right then i think we are apache then 👍🏻

js/src/browser.ts
cpinn: This is a downside of our current linting approach.

Each package needs its own linting and own configuration. I would like to do this but I haven't gotten around to it. Personally would like to move everything to oxlint which is what I was working on in a separate branch but oxlint and custom rules like this actually don't work that great at the moment.

I really didn't want an overhaul of our entire linting to be a part of this PR.

js/src/symbols.ts
```diff
@@ -0,0 +1 @@
+export const BRAINTRUST_STATE_KEY = Symbol.for("braintrust-state");
```
Qard: You actually don't need to export and import this. Anywhere that you do `Symbol.for(string)` with the same string contents will produce the exact same symbol instance.

Qard (approved):
Not a huge fan of all the infrastructure necessary for supporting smoke testing, but otherwise looks fine. We can address the smoke tests another time--it'd be a major refactor to get all that to something cleaner.

A few minor notes, but nothing blocking. Ship it!

js/src/node.ts
```diff
@@ -9,16 +9,25 @@ import iso from "./isomorph";
...
+import * as zlib from "node:zlib";
+
+let nodeConfigured = false;
 
 export function configureNode() {
+  if (nodeConfigured) {
```
cpinn: It is okay to setup/call iso multiple times per module but these are just safeguards as it is unnecessary work.

ibolmo:
looks like the auto import the right entrypoint may not be working with workerd (at least my local run):

image


getting a context loss 🤔  

image

js/package.json
```diff
@@ -24,14 +24,27 @@
...
+      "default": "./dist/index.mjs"
     },
     "./browser": {
       "import": "./dist/browser.mjs",
-      "module": "./dist/browser.mjs",
-      "require": "./dist/browser.js"
+      "require": "./dist/browser.js",
+      "default": "./dist/browser.mjs"
+    },
+    "./node": {
```
ibolmo: we're missing the new "./worked" export

js/package.json
```diff
@@ -24,14 +24,27 @@
     "./package.json": "./package.json",
     ".": {
       "types": "./dist/index.d.ts",
+      "edge-light": "./dist/browser.mjs",
+      "workerd": "./dist/browser.mjs",
```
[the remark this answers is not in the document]
cpinn: yes. I never added a specific separate workerd export. 

My original plan had really just been to create and use the new template registry. That was what I had originally wanted from this PR. 

Then the plan entirely changed into the completely new browser package and also a cloudflare package (?) if there was a cloudflare package there is no need for workerd. 

I am still really confused as to what we want here. It seems more confusing with the new browser package in general.

js/src/cli/functions/load-module.ts
```diff
@@ -31,7 +31,7 @@ export function loadModule({
     };
     globalThis._lazy_load = true;
     const state = _internalGetGlobalState();
-    (globalThis as any)[BRAINTRUST_STATE_KEY] = state;
+    (globalThis as any)[Symbol.for(BRAINTRUST_STATE_SYMBOL_NAME)] = state;
```
Qard: Doesn't need the shared string either.

```js
Symbol.for("foo")
// elsewhere...
Symbol.for("foo") // same symbol instance, despite different string instances, because string content is same
```

PR #1276, opened 2026-01-16
init-dataset-with-id

ankrgyl:
Make it easier to initialize an experiment's dataset reference without calling `initDataset`

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -98,6 +98,14 @@ from .xact_ids import prettify_xact
 Metadata = dict[str, Any]
 DATA_API_VERSION = 2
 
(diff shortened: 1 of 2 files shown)


py/src/braintrust/logger.py
```diff
@@ -1294,7 +1294,7 @@ def init(
     project: str | None = None,
     experiment: str | None = None,
     description: str | None = None,
-    dataset: Optional["Dataset"] = None,
+    dataset: Optional[Union["Dataset", dict[str, str]]] = None,
```
ibolmo: `Optional['Dataset'] | dict[str, str]` (preferably a `TypedDict`)

PR #1278, opened 2026-01-16
Python trace scoring candidate

CLowbrow:


py/src/braintrust/functions/invoke.py
```diff
@@ -243,6 +243,8 @@ def init_function(project_name: str, slug: str, version: str | None = None):
     :param version: Optional version of the function to use. Defaults to latest.
     :return: A function that can be used as a task or scorer.
     """
+    # Disable span cache since remote function spans won't be in the local cache
+    _internal_get_global_state().span_cache.disable()
```
ibolmo: may be too indirect. if the remote function spans are not in the cache what's the big deal in checking the cache?

py/src/braintrust/logger.py
```diff
@@ -401,6 +401,11 @@ def default_get_api_conn():
             ),
         )
 
+        from braintrust.span_cache import SpanCache
```
ibolmo: #nit i think we can move this to the top of the file

py/src/braintrust/trace.py
```diff
@@ -0,0 +1,385 @@
...
+
+    def get_configuration(self) -> dict[str, str]:
+        """Get the trace configuration."""
+        return {
+            "object_type": self._object_type,
+            "object_id": self._object_id,
+            "root_span_id": self._root_span_id,
+        }
+
+    async def get_spans(self, span_type: Optional[list[str]] = None) -> list[SpanData]:
```
ibolmo: give the option to disable the cache

PR #1288, opened 2026-01-20
Reapply braintrust/temporal integration from PR (#1261)

cpinn:
Reapply the changeset for braintrust/temporal from this PR: https://github.com/braintrustdata/braintrust-sdk/pull/1261.

The devDependencies on braintrust/temporal originally had a workspace dependency. I thought this might be causing the issue but then I realized we forcefully ignored the dependencies for otel-v1 and otel-v2. I ended up doing the same for braintrust/temporal.

### DIFF
--- a/integrations/temporal-js/package.json
+++ b/integrations/temporal-js/package.json
@@ -35,7 +35,7 @@
     "@temporalio/worker": "^1.14.1",
     "@temporalio/workflow": "^1.14.1",
     "@types/node": "^22.15.21",
-    "braintrust": ">=2.0.0",
+    "braintrust": "workspace:*",
     "tsup": "^8.5.0",
     "typescript": "5.5.4",
     "vitest": "^2.1.9"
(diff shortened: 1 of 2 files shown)


integrations/temporal-js/package.json
```diff
@@ -0,0 +1,71 @@
...
+  },
+  "author": "Braintrust Data Inc.",
+  "license": "MIT",
+  "devDependencies": {
+    "@temporalio/activity": "^1.14.1",
+    "@temporalio/client": "^1.14.1",
+    "@temporalio/common": "^1.14.1",
+    "@temporalio/worker": "^1.14.1",
+    "@temporalio/workflow": "^1.14.1",
+    "@types/node": "^22.15.21",
```
cpinn: hmm I thought we would want temporal dependencies as dev dependencies but I don't want them installed in the workspace so I have taken them out.

PR #1292, opened 2026-01-20
expose wrapAgentClass and fix private name access

ibolmo:
Customers may need to wrap their own agent class that they've implemented (all they need to do is implement an interface) and this allows them to have their agent class get traced


knjiang (approved):
seems good. curious how we will document! feel free to assign me that too

ibolmo:
> seems good. curious how we will document! feel free to assign me that too

what is this documentation you speak of? haha. maybe we should add ref docs (autogenerated) of our integrations too

PR #1295, opened 2026-01-21
Modify package json exports for browser environments

cpinn:
This PR updates default entrypoint selection so each runtime resolves the correct build. Node environments select the Node entrypoint, while browser and worker environments default to the browser build.

Worker runtimes no longer implicitly resolve the Node build. Importing braintrust/node must be done explicitly and is expected to fail in workers (e.g. cloudflare-vite-hono-vite-dev-node-esm), rather than failing due to accidental default resolution.

PR changes:

The change uses Node’s conditional exports to ensure the correct build is selected based on runtime capabilities, instead of relying on ambiguous defaults.

Each scenario now checks that the appropriate isomorphic build type (Node vs browser) is selected and that module resolution lands on the correct output files for that environment.

### DIFF
--- a/js/smoke/shared/src/suites/import-verification.ts
+++ b/js/smoke/shared/src/suites/import-verification.ts
@@ -912,10 +912,9 @@ function buildSuccessMessage(
  */
 export async function runImportVerificationTests(
   module: BraintrustModule,
-  options?: {
-    checkBuildResolution?: boolean;
-    expectedBuild?: "browser" | "node";
-    expectedFormat?: "cjs" | "esm";
+  options: {
+    expectedBuild: "browser" | "node";
+    expectedFormat: "cjs" | "esm";
   },
 ): Promise<TestResult[]> {
   const results: TestResult[] = [];
@@ -934,17 +933,13 @@ export async function runImportVerificationTests(
   results.push(await testStateManagementExports(module));
   results.push(await testExperimentExports(module));
   results.push(await testEvalExports(module));
-
-  // Optionally check which build was resolved
-  if (options?.checkBuildResolution) {
-    results.push(
-      await testBuildResolution(
-        module,
-        options.expectedBuild,
-        options.expectedFormat,
-      ),
-    );
-  }
+  results.push(
+    await testBuildResolution(
+      module,
+      options.expectedBuild,
+      options.expectedFormat,
+    ),
+  );
 
   return results;
 }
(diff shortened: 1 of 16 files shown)


cpinn:
@ibolmo I think this is ready for a review now. I updated the older vite-hono dev test to run all the expected tests and I added a new test to explicitly pull in the braintrust/node package and run into the build error.

js/package.json
```diff
@@ -24,15 +24,28 @@
     "./package.json": "./package.json",
     ".": {
       "types": "./dist/index.d.ts",
```
ibolmo: pulled down the node.js repo   https://gist.github.com/ibolmo/9a1d923447657c02d69893cb4f7d0ecf[4:29 PM]the code doesn't do any kind of evaluation. it uses a list of conditions and one of them is the well known conditions so yeah  i think the way you have it makes a lot of sense

js/smoke/shared/src/suites/import-verification.ts
```diff
@@ -706,5 +935,16 @@ export async function runImportVerificationTests(
   results.push(await testExperimentExports(module));
   results.push(await testEvalExports(module));
 
+  // Optionally check which build was resolved
+  if (options?.checkBuildResolution) {
```
ibolmo: looks like they're all true?

js/package.json
```diff
@@ -24,15 +24,28 @@
...
+      "edge-light": "./dist/browser.mjs",
```
ibolmo: should we always use:

```ts
{
  "import": ...
  "require": ...
}
```

?

ibolmo (approved):
now that i'm seeing all of the updated tests where we assume workerd will use browser it is really nice that the naive user will likely have a working braintrust trace. 

if they can they should use /node

js/smoke/scenarios/deno-browser/deno.json
```diff
@@ -2,7 +2,7 @@
   "imports": {
     "@std/assert": "jsr:@std/assert@^1.0.14",
     "@braintrust/smoke-test-shared": "jsr:@braintrust/smoke-test-shared",
-    "braintrust/browser": "npm:braintrust@^2.0.2/browser"
+    "braintrust": "npm:braintrust/browser"
```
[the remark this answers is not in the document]
cpinn: deno will by default pull in the braintrust node build. 

I tried looking around online but it seems like the only way to make this be a specific test for the browser was to pull in the brower import? 

If i change to pull in just npm:braintrust it would be the same as the deno-node test.

cpinn:
I fixed the small cleanup. Can discuss  whether we want the other small changes in the other PR before the release.

PR #1303, opened 2026-01-24
make python context vars more robust to threads

ibolmo:
in the PR:
- new `setup_threads`, `wrap_thread`, `wrap_thread_pool_executor` to automatically propagate our context vars
- new `BRAINTRUST_INSTRUMENT_THREADS` to run `setup_threads` . figured to do this as opt-in for now. 
- [lots of tests](https://github.com/braintrustdata/braintrust-sdk/pull/1303/changes#diff-0132eb216c0ab9720ab927c06c7c192f055b4efa8f6473f64c9b7889b39f39ab) (this is where I'd like reviewer's time should be spent). did i miss anything? 
- logger `.end()` etc., will stop raising an exception.. possibly risky, but seems sensible. at worst case the hierarchy is lost or spans are set to `NOOP_SPAN` (worst case) but at least user's code may proceed. 
  - as a follow-up we should add more helpful logging for customers to help debug things

up next:

I'm going to check on our integrations. based on the tests, it seems like we can remove a lot of dead code now.

### DIFF
--- a/py/src/braintrust/test_context.py
+++ b/py/src/braintrust/test_context.py
@@ -2,129 +2,77 @@
 Context Propagation Tests for Braintrust SDK
 
 This test suite validates context propagation behavior across various concurrency patterns.
-Threading tests run TWICE - with and without auto-instrumentation - to demonstrate:
-1. WITHOUT auto-instrumentation: Context is lost (xfail)
-2. WITH auto-instrumentation: Context is preserved (pass)
 
-ALL XFAILS ARE INTENTIONAL DOCUMENTATION:
-- Every xfail is the "[no_auto_instrument]" parameterized version of a test
-- The same test PASSES with "[with_auto_instrument]"
-- This proves auto-instrumentation solves the problem!
+TEST ISOLATION STRATEGY:
+- Tests use pytest-forked to run each test in an isolated process
+- This ensures setup_threads() patches don't leak between tests
+- Use unpatched(scenario) for xfail tests (documents context loss)
+- Use patched(scenario) for tests that prove setup_threads() fixes it
 
-BREAKDOWN:
-- 13 threading patterns × 2 (with/without auto-instrument) = 26 tests
-  → 13 PASS (with auto-instrument), 13 XFAIL (without)
-- 27 other tests (async patterns, auto-instrument specific, etc.) = 27 PASS
-- TOTAL: 40 PASSED, 13 XFAILED ✅
+Example:
+    def _threadpool_scenario(test_logger, with_memory_logger):
+        # test logic...
 
-Threading tests are parameterized to run BOTH with and without auto-instrumentation,
-demonstrating that our auto-instrumentation solution solves the context loss problem.
+    test_threadpool_loses_context = unpatched(_threadpool_scenario)
+    test_threadpool_with_patch = patched(_threadpool_scenario)
 
-REALISTIC INTEGRATION TESTS (Good Intent, Wrong Implementation):
-Each tests a different way integrations can get context propagation wrong:
-
-1. ✅ Library doing context RIGHT
-   - Uses copy_context() at the right time
-   - Passes WITH and WITHOUT auto-instrumentation
-   - Proves: Auto-instrumentation doesn't break well-behaved libraries
-
-2. ✅ Forgot context propagation entirely
-   - Developer didn't know about context
-   - XFAIL without auto-instrumentation, PASS with it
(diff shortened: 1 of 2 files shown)


py/src/braintrust/logger.py
```diff
@@ -4060,10 +4060,21 @@ def set_current(self):
...
+        This is safe - context_manager.unset_current_span() handles None.
+        """
         if self.can_set_current:
-            # Pass the stored token to context manager for cleanup
-            self.state.context_manager.unset_current_span(self._context_token)
-            self._context_token = None
+            try:
+                self.state.context_manager.unset_current_span(self._context_token)
+            except Exception:
+                # Cleanup should never break the application
```
clutchski: we should log debug here at least.

py/src/braintrust/logger.py
```diff
@@ -4074,8 +4085,16 @@ def __exit__(self, exc_type, exc_value, tb) -> None:
...
-            self.end()
+            try:
+                self.unset_current()
+            except Exception:
+                pass
+
+            try:
+                self.end()
+            except Exception as e:
+                # Span ending failures are serious - log as warning
```
clutchski: remove this comment

py/src/braintrust/test_context.py
```diff
@@ -0,0 +1,1869 @@
...
+- 27 other tests (async patterns, auto-instrument specific, etc.) = 27 PASS
+- TOTAL: 40 PASSED, 13 XFAILED ✅
+
+Threading tests are parameterized to run BOTH with and without auto-instrumentation,
+demonstrating that our auto-instrumentation solution solves the context loss problem.
+
+REALISTIC INTEGRATION TESTS (Good Intent, Wrong Implementation):
+Each tests a different way integrations can get context propagation wrong:
+
+1. ✅ Library doing context RIGHT (e.g., LangChain)
```
clutchski: remove langchain here

py/src/braintrust/test_context.py
```diff
@@ -0,0 +1,1869 @@
...
+    parent_log = next((l for l in logs if l["span_attributes"]["name"] == "parent"), None)
+    assert parent_log is not None
+
+
+@pytest.mark.asyncio
+async def test_fastapi_background_task_pattern(test_logger, with_memory_logger, auto_instrument):
+    """
+    Test: FastAPI background tasks pattern.
+
+    FastAPI uses loop.run_in_executor() for background tasks:
```
clutchski: i feel like all the tests have a bit too much comments . docuemntation and the test code is probably sufficient to document it self. maybe one line comemnts?

py/src/braintrust/test_context.py
```diff
@@ -0,0 +1,1869 @@
...
+    parent_log = next(l for l in logs if l["span_attributes"]["name"] == "pool_explicit_parent")
+    assert parent_log is not None
+
+
+# ============================================================================
+# SUMMARY TEST
+# ============================================================================
+
+
+def test_print_comprehensive_summary():
```
clutchski: this seems not valuable

py/src/braintrust/test_context.py
```diff
@@ -0,0 +1,1869 @@
...
+        thread.join()
+
+    # Verify context was preserved (worker saw the parent)
+    assert parent_seen_by_worker is not None
+    assert parent_seen_by_worker.id == parent_id, "Auto-instrumentation should preserve span context"
+
+    test_logger.flush()
+    logs = with_memory_logger.pop()
+
+    # Memory logger is thread-local, so only parent log appears in tests
```
clutchski: i don't think this is true anymore and if it is , we should fix it so taht we can actualy verify the span/child relationships are created.

py/src/braintrust/test_context.py
```diff
@@ -0,0 +1,1869 @@
...
+        setup_threads()
+    return request.param
+
+
+# ============================================================================
+# CONTEXT MANAGER PATTERN: with start_span(...)
+# ============================================================================
+
+
+def test_threadpool_context_manager_pattern(test_logger, with_memory_logger, auto_instrument):
```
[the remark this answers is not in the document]
ibolmo: are you worried that the test are some how not working as intended because they're not isolated? i think the xfail here would have caught that, but i'll see what the ai thinks. https://github.com/braintrustdata/braintrust-sdk/pull/1303#discussion_r2738427946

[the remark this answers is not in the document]
ibolmo: never opposed to more robust testing

PR #1307, opened 2026-01-27
braintrust/browser package and use of async local storage in browser

cpinn:
## New braintrust/browser package

This changset adds a new package, `braintrust/browser`, which introduces the browser-compatible AsyncLocalStorage als-browser package.

## Use the new package

`pnpm add @braintrust/browser braintrust`

Users can export anything they would usually export from braintrust from braintrust/browser.

## Using the older braintrust build

Users that install `braintrust` will now receive a message if they are attempting to use the build in a browser environment. This will inform them to install the new `braintrust/browser` package for improved compatibility.

Importing directly from braintrust in a browser environment will continue to work. The new exports for edge-light, workerd, and browser exports will point to isomorph builds with no-op or deprecated behavior.


js/smoke/scenarios/browser-main-package/src/browser-message-test.ts
```diff
@@ -0,0 +1,58 @@
...
+      hasNewId: boolean;
+      hasTraceable: boolean;
+    };
+  }
+}
+
+// Capture console.info messages BEFORE importing braintrust
+const capturedMessages: string[] = [];
+const originalConsoleInfo = console.info;
+console.info = (...args: any[]) => {
```
cpinn: I am not really sure what/how to send the user a proper message. All I have so far is a console message to inform them of the deprecation? Not sure if there are any other ideas on what we would want to do here. I don't want the package to entirely break so I think having a noop here makes sense.

integrations/browser-js/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "als-browser": "^1.0.1"
+  },
+  "devDependencies": {
+    "@types/node": "^20.10.5",
+    "braintrust": "workspace:*",
+    "tsup": "^8.3.5",
+    "typescript": "^5.3.3",
+    "vitest": "^2.1.9"
+  },
+  "peerDependencies": {
```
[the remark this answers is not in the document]
cpinn: cc: @Qard  

@ibolmo and I wanted to get your opinion on this packaging change in particular. Let me know if you have thoughts.

integrations/browser-js/src/browser-config.ts
```diff
@@ -19,12 +20,17 @@ export function configureBrowser() {
   // Set build type indicator
   iso.buildType = "browser";
 
+  // Try to use native AsyncLocalStorage if available (edge runtimes)
+  // Otherwise fall back to als-browser polyfill
   try {
     if (typeof AsyncLocalStorage !== "undefined") {
```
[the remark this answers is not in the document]
Qard: Presumably not if it's truly browser-only and not browsers plus typically bundled runtimes like FaaS providers.

js/edge-light.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Browser-compatible build of the Braintrust SDK for edge runtime environments.
+ *
+ * This build uses a noop isomorph that provides browser-safe implementations
+ * for Node.js-specific features.
+ *
+ * Edge runtimes (Cloudflare Workers, Vercel Edge, Deno Deploy) have native
+ * AsyncLocalStorage support. For optimal support with built-in polyfill, consider:
+ *   npm install @braintrust/browser
```
Qard: Is this correct that the edge-light.ts is used through the browser build? I don't see it in the package.json, is this actually exposed?

cpinn: Yes this would be a new export to properly route "edge" environments "somewhere". 

Currently users in the edge environment are importing directly from "braintrust/browser" which is a bit odd as that package was originally intended for just the real browser.  

I am not quite sure what this environment needs. In vercel edge AsyncLocalStorage is attached as a global variable and the environment actually allows access to the node AsyncLocalStorage.

integrations/browser-js/package.json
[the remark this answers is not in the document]
cpinn: hmm. I haven't used yarn in a while but I felt it warned you just didn't install by default. 

Background: 
Originally I had this change such that braintrust was bundled as a different package called braintrust/browser from the source of braintrust. That way users wouldn't need a version of braintrust and the package just had everything they needed since we don't have a concept of braintrust/core (or some other base package name). I wasn't certain about the tight coupling in that case so I modified the changeset to be more flexible with the base package. 

Without a peer dependency though wouldn't we need to pin the version of braintrust for this package?  

Not sure which option you like or dislike more let me know!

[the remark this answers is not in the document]
ibolmo: I'm also worried about having two installs like:

```
braintrust: 2.1.0
@braintrust/browser: 1.0.0
```

if they have two installs it's possible for the user to have two ALS vars going and it may be very surprising for the user. 

part of the motivation for peer dependency is that the @braintrust/browser should use `braintrust` ALS. 

alternatively, we need to say they need to remove `braintrust` 🤔 i wonder what our docs will look like 😅

integrations/browser-js/README.md
```diff
@@ -0,0 +1,86 @@
...
+- Experiments and datasets
+- Prompt management
+- AI provider wrappers (OpenAI, Anthropic, Google)
+- Evaluation framework
+- OpenTelemetry integration
+
+## Environment Support
+
+- Modern browsers (Chrome, Firefox, Safari, Edge)
+- Edge runtimes (Cloudflare Workers, Vercel Edge)
```
ibolmo: ditto here remove

integrations/js/artifacts/braintrust-templates-nunjucks-js-0.0.1.tgz
cpinn: This snuck in during some local testing.

integrations/templates-nunjucks/smoke/scenarios/nextjs/src/instrumentation.ts
```diff
@@ -1,8 +1,10 @@
-import { registerTemplatePlugin } from "braintrust";
-import { nunjucksPlugin } from "@braintrust/templates-nunjucks-js";
-
 export async function register() {
   if (process.env.NEXT_RUNTIME === "nodejs") {
```
ibolmo: i was expecting a check for `NEXT_RUNTIME === 'edge'` 🤔 i wonder why the test works

js/src/browser-isomorph.ts
```diff
@@ -0,0 +1,62 @@
...
+export function configureBrowserIsomorph(): void {
+  if (browserConfigured) {
+    return;
+  }
+
+  // Show informational message once
+  if (!messageShown && typeof console !== "undefined") {
+    console.info(
+      "Braintrust SDK Browser Build\n" +
+        "You are using a browser-compatible build from the main package.\n" +
```
ibolmo: I was thinking more like:

```
This entrypoint is no longer supported. 

You should be using entrypoints: 

- `/worked` (cloudflare envs), 
- `/edge-lite` (next-js or other edge envs) 

If you'd like to use braintrust in the browser use the dedicated package: @braintrust/browser
```

js/edge-light.ts
ibolmo: yeah this is more of a stub, but if users use this import we shouldn't show the warning message. perhaps we need an option in configureBrowserIsomorph to warn or not warn.

integrations/templates-nunjucks/smoke/scenarios/nextjs/src/instrumentation.ts
cpinn: I had wanted this test to be for the nodejs environment which should easily work with templates nunjucks. 

That was why I was confused as to why the test began failing at all after the packaging change to create this new package. 

I still have to figure this out which was why I wrote the note to myself to undo this change once the actual direction of this package has been decided.

I think it would be easier to release a 3.0.0 version with just the pull out of the templates nunjucks package without this change entirely.

integrations/browser-js/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Differences from Main Package
+
+This package:
+
+- **Includes** `als-browser` polyfill for AsyncLocalStorage (bundled)
+- **Requires** `braintrust` as a peer dependency
+- **Optimized** for browser bundle size
+- **Auto-configures** browser environment on import
+- **Supports** edge runtimes with native AsyncLocalStorage (Cloudflare Workers, Vercel Edge)
```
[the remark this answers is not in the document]
ibolmo: well it'll be mostly to avoid the same coupling we had with `/browser`. 

i could also see `@braintrust/cloudflare` maybe having a vite plugin wrapper/fix (or other unique code only affecting that package)

js/package.json
```diff
@@ -11,15 +11,13 @@
   "main": "./dist/index.js",
   "module": "./dist/index.mjs",
   "types": "./dist/index.d.ts",
-  "browser": {
-    "./dist/index.js": "./dist/browser.js",
-    "./dist/index.d.ts": "./dist/browser.d.ts",
```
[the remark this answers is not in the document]
cpinn: I undid some of these changes to keep them minimal we can decide on the final configuration in the main PR.

PR #1308, opened 2026-01-27
Add review span type

tara-nagar:


generated_types.json
```diff
@@ -3849,6 +3849,18 @@
           "replace"
         ]
       },
+      "ImageRenderingMode": {
```
j13huang: I think this shouldn't be a part of this pr? or am I missing something

PR #1318, opened 2026-01-29
add set filtering function

ibolmo:
a general purpose to help customers decide if and when a span will go out. doesn't make any opinions about root/child, etc.

pending:  I haven't given a lot of thought is what happens with root/child/ancestor. if root goes away, or if child goes away. i'll send more tests

### DIFF
--- a/js/src/logger.test.ts
+++ b/js/src/logger.test.ts
@@ -1384,4 +1384,110 @@ describe("filtering functionality", () => {
     expect(loggedNames).not.toContain("sibling"); // Filtered
     expect(loggedNames).not.toContain("descendant"); // Also filtered (cascaded)
   });
+
+  /**
+   * Test what happens when the ROOT span is filtered.
+   *
+   * Given:
+   *     root  <-- FILTERED
+   *       parent
+   *         child
+   *         sibling
+   *           descendant
+   *
+   * Result: All spans still logged, but parent becomes orphan.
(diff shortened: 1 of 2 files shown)


js/src/logger.ts
```diff
@@ -2498,6 +2383,14 @@ export class TestBackgroundLogger implements BackgroundLogger {
     this.maskingFunction = maskingFunction;
   }
 
+  setFilteringFunction(
```
cpinn: What happens when someone sets this up multiple times in their project looks like it would just overwrite the function?

In that case would it make more sense to have this as an option during logger initialization?

ibolmo: 🤔 maybe. i followed the masking pattern but maybe that one needs a tweak too

js/src/logger.test.ts
```diff
@@ -1197,3 +1198,159 @@ describe("sensitive data redaction", () => {
...
+    _exportsForTestingOnly.clearTestBackgroundLogger();
+  });
+
+  test("filter out events with langsmith:hidden tag", async () => {
+    const filteredSpanIds = new Set<string>();
+
+    const filteringFunction = (event: any): any | null => {
+      const spanId = event.span_id;
+      if (spanId && filteredSpanIds.has(spanId)) {
+        return null;
```
clutchski: I feel like null is an odd choice. Maybe keep/drop or True/false? Then they can be chained together etc

PR #1323, opened 2026-01-30
add root_span_id and span_id to the updateSpan log if they were exported

j13huang:
I believe these don't get used to actually update the span since the id is provided, but it helps our realtime events in the UI


ankrgyl (commented):
can you make the equivalent change in python?

lforst:
Superseded by https://github.com/braintrustdata/braintrust-sdk-javascript/pull/1353

PR #1332, opened 2026-02-03
Fix init logger state isolation

stretpjc:
When passing `state=BraintrustState()` to `init_logger()`, the logger was not fully isolated because `_compute_logger_metadata()` was still using the global `_state` instead of the passed state parameter.

Changes:
- Modified `_compute_logger_metadata()` to accept a required `state` parameter  and use it for `org_id` and `app_conn()` instead of the global `_state`
- Updated `init_logger()` to pass the state to `_compute_logger_metadata()`
- Updated `_span_components_to_object_id_lambda()` to explicitly call `login()` and pass `_state` when computing project IDs for deserialized span components
- Updated state so that if `api_key` or `app_url` exists then `BraintrustState()` is passed implicitly
- Added tests

This enables multi-tenant logging where separate loggers can be created for different organizations, each with their own isolated BraintrustState.

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -1766,11 +1766,12 @@ def init_dataset(
     )
 
 
-def _compute_logger_metadata(project_name: str | None = None, project_id: str | None = None):
-    login()
-    org_id = _state.org_id
+def _compute_logger_metadata(
+    state: "BraintrustState", project_name: str | None = None, project_id: str | None = None
+):
+    org_id = state.org_id
     if project_id is None:
-        response = _state.app_conn().post_json(
+        response = state.app_conn().post_json(
             "api/project/register",
             {
                 "project_name": project_name or GLOBAL_PROJECT,
@@ -1783,7 +1784,7 @@ def _compute_logger_metadata(project_name: str | None = None, project_id: str | 
             project=ObjectMetadata(id=resp_project["id"], name=resp_project["name"], full_info=resp_project),
         )
     elif project_name is None:
-        response = _state.app_conn().get_json("api/project", {"id": project_id})
+        response = state.app_conn().get_json("api/project", {"id": project_id})
         return OrgProjectMetadata(
             org_id=org_id, project=ObjectMetadata(id=project_id, name=response["name"], full_info=response)
         )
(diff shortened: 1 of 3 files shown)


py/src/braintrust/logger.py
```diff
@@ -1819,7 +1832,13 @@ def init_logger(
     :returns: The newly created Logger.
     """
 
-    state = state or _state
+    # If no explicit state is provided but api_key or app_url is specified,
+    # create an isolated state to avoid conflicts with the global state
+    if state is None:
```
clutchski: i feel like this should go into a helper function

clutchski:
I think (a) the tests need to pass (b) we should somehow have tests verify the correct behaviour (e.g. data is sent to more than one org) vs just asserting internal detals about the logger

py/src/braintrust/logger.py
```diff
@@ -1429,6 +1429,22 @@ def _internal_get_global_state() -> BraintrustState:
...
+
+    If an explicit state is provided, use it. Otherwise, if api_key or app_url
+    is specified, create a new isolated BraintrustState to avoid conflicts with
+    the global state. If neither is specified, use the global state.
+    """
+    if state is not None:
+        return state
+    if api_key is not None or app_url is not None:
+        return BraintrustState()
+    return _state
```
ibolmo: the `api_key` and `app_url` is a bit odd for this func. i'm worried we would make an isolated state each time causing unusual

py/src/braintrust/logger.py
```diff
@@ -3400,7 +3417,12 @@ def _span_components_to_object_id_lambda(components: SpanComponentsV4) -> Callab
         raise Exception("Impossible: compute_object_metadata_args not supported for experiments")
     elif components.object_type == SpanObjectTypeV3.PROJECT_LOGS:
         captured_compute_object_metadata_args = components.compute_object_metadata_args
-        return lambda: _compute_logger_metadata(**captured_compute_object_metadata_args).project.id
+
+        def compute_project_id():
+            login()
```
ibolmo: why should we always login?

ibolmo (commented):
> Modified _compute_logger_metadata() to accept a required state parameter and use it for org_id and app_conn() instead of the global _state

based on the PR description that seems sensible. is there a requirement to do all of these _resolve_state and possibly create a random BraintrustState()? 

if we need to, we should error if the user is not logged in

lforst:
PR is stale, python code lives somewhere else now

PR #1337, opened 2026-02-04
Add sandbox registration and types

nselvidge:
adds utility to register a sandbox with braintrust

Data plane APIs are not yet available, so these utilities won't do anything. merging them so they'll be ready when the data plane APIs are available. 

this doesn't include any changes to the CLI, but that will come in a follow up PR soon. 
<s>
I suspect some of this API is going to change over the next few weeks as we nail down supported integrations and workflows, so I'm marking things as experimental and preventing docs from getting generated for now. </s>

This workflow should be stable going forward, but we'll likely add new ways to create sandboxes in addition to these.

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -38,8 +38,7 @@ import {
   VALID_SOURCES,
   isArray,
   isObject,
-} from "../util/index";
-import { slugify } from "../util/string_util";
+} from "./util";
 import {
   type AnyModelParamsType as AnyModelParam,
   AttachmentReference as attachmentReferenceSchema,
@@ -66,7 +65,6 @@ import {
   type PromptSessionEventType as PromptSessionEvent,
   type RepoInfoType as RepoInfo,
   type PromptBlockDataType as PromptBlockData,
-  type IfExistsType,
 } from "./generated_types";
 
 const BRAINTRUST_ATTACHMENT =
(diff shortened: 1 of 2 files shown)


js/src/logger.ts
```diff
@@ -3739,6 +3741,136 @@ export function initDataset<
...
+ * ```
+ */
+export async function experimental_registerSandbox(
+  options: ExperimentalRegisterSandboxOptions,
+): Promise<ExperimentalRegisterSandboxResult> {
+  const state = _internalGetGlobalState();
+  await login({
+    apiKey: options.apiKey,
+    appUrl: options.appUrl,
+    orgName: options.orgName,
```
ankrgyl: nowadays it's convention to accept the state as an optional param and then default it to the global state if nothing is passed in

py/src/braintrust/logger.py
```diff
@@ -1775,6 +1777,167 @@ def compute_metadata():
...
+
+        result = experimental_register_sandbox(
+            name="My Sandbox",
+            project="My Project",
+            entrypoints=["./my-eval.eval.py"],
+            sandbox=ExperimentalSandboxConfig(provider="modal", snapshot_ref="sb-xxx"),
+        )
+        print([f.id for f in result.functions])
+    """
+    state = _state
```
ankrgyl: same comment RE state

js/src/logger.ts
```diff
@@ -38,7 +38,8 @@ import {
   VALID_SOURCES,
   isArray,
   isObject,
-} from "./util";
+} from "../util/index";
```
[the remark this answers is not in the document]
nselvidge: oops, was a holdover from previous version. will remove

PR #1341, opened 2026-02-05
Add per-input trialCount support to Eval()

dflynn15:
## Why?

Braintrust's `Eval()` function supports a `trialCount` parameter that runs each input multiple times to measure variance in non-deterministic LLM outputs. However, this setting applies globally to all inputs, which creates some (but minimal) friction in some evaluation workflows. For example:

1. **Targeted Debugging is Expensive**: When investigating a single flaky test case, you want to run it 10-20 times to understand the variance pattern. With global `trialCount`, this means running your entire suite 10-20 times, multiplying costs and wait time unnecessarily.

2. **Mixed Determinism is Common**: Real evaluation suites contain a mix of deterministic scenarios (math problems, factual lookups) and non-deterministic ones (creative writing, open-ended reasoning). Forcing the same trial count on both wastes resources.

3. **Cost Scales Linearly**: Every additional trial means another LLM API call. A global `trialCount: 5` on a 100-item dataset means 500 API calls, even if only 10 items actually need variance analysis.

In order to address this, we've created a custom solution that I want to propose as a contribution. Specifically it, allows each data item to specify its own `trialCount`, overriding the global default. This gives users fine-grained control over where to invest their evaluation budget.

---

## What?

```typescript
Eval("My Project", {
  data: [
    { input: "stable query", expected: "..." },                    // Uses global (3)
    { input: "flaky query", expected: "...", trialCount: 10 },     // Override to 10
    { input: "deterministic", expected: "...", trialCount: 1 },    // Override to 1
  ],
  task: myTask,
  scores: [Factuality],
  trialCount: 3, // Global default
});
```

There is a corollary Python PR up to match it here: #1342

### DIFF
--- a/js/src/framework.test.ts
+++ b/js/src/framework.test.ts
@@ -528,6 +528,99 @@ test("trialIndex with multiple inputs", async () => {
   expect(input2Trials).toEqual([0, 1]);
 });
 
+test("per-input trialCount overrides global trialCount", async () => {
+  const trialData: Array<{ input: number; trialIndex: number }> = [];
+
+  const { results } = await runEvaluator(
+    null,
+    {
+      projectName: "proj",
+      evalName: "eval",
+      data: [
+        { input: 1, expected: 2 },
+        { input: 2, expected: 4, trialCount: 5 },
+        { input: 3, expected: 6, trialCount: 1 },
+      ],
+      task: async (input: number, { trialIndex }) => {
+        trialData.push({ input, trialIndex });
+        return input * 2;
+      },
+      scores: [],
+      trialCount: 2,
+    },
+    new NoopProgressReporter(),
+    [],
+    undefined,
+    undefined,
+    true,
+  );
+
+  expect(results).toHaveLength(8);
+  expect(trialData).toHaveLength(8);
+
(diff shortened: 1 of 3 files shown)


lforst:
I have trouble fetching your fork because of a weird GH error:

```
2026-04-14 14:50:59.791 [error] [Review+0] Checkout failed #{"stdout":"","stderr":"ERROR: Access to this repository has been disabled by GitHub staff due to excessive resource use.\nPlease contact support via https://support.github.com/contact to restore access to this repository.\nRead about how to decrease the size of your repository:\n  https://docs.github.com/articles/what-is-my-disk-quota\n\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\n","exitCode":128,"gitErrorCode":"RemoteConnectionError","gitCommand":"fetch","gitArgs":["fetch","dflynn15","feature/js-per-input-trial-count:pr/dflynn15/1341"]}
```

Gonna re-implement it and tag you as a co-author. Ty!

lforst:
Superseded by: https://github.com/braintrustdata/braintrust-sdk-javascript/pull/1814

PR #1342, opened 2026-02-05
Add per-input trial_count support to Eval()

dflynn15:
## Why?

Braintrust's `Eval()` function supports a `trialCount` parameter that runs each input multiple times to measure variance in non-deterministic LLM outputs. However, this setting applies globally to all inputs, which creates some (but minimal) friction in some evaluation workflows. For example:

1. **Targeted Debugging is Expensive**: When investigating a single flaky test case, you want to run it 10-20 times to understand the variance pattern. With global `trialCount`, this means running your entire suite 10-20 times, multiplying costs and wait time unnecessarily.

2. **Mixed Determinism is Common**: Real evaluation suites contain a mix of deterministic scenarios (math problems, factual lookups) and non-deterministic ones (creative writing, open-ended reasoning). Forcing the same trial count on both wastes resources.

3. **Cost Scales Linearly**: Every additional trial means another LLM API call. A global `trialCount: 5` on a 100-item dataset means 500 API calls, even if only 10 items actually need variance analysis.

In order to address this, we've created a custom solution that I want to propose as a contribution. Specifically it, allows each data item to specify its own `trialCount`, overriding the global default. This gives users fine-grained control over where to invest their evaluation budget.

---

## What?

```python
Eval(
    "My Project",
    data=[
        {"input": "stable query", "expected": "..."},                     # Uses global (3)
        {"input": "flaky query", "expected": "...", "trial_count": 10},   # Override to 10
        EvalCase(input="deterministic", expected="...", trial_count=1),   # Override to 1
    ],
    task=my_task,
    scores=[Factuality],
    trial_count=3,  # Global default
)
```

There is a corollary JS PR up to match it here: #1341

### DIFF
--- a/py/src/braintrust/framework.py
+++ b/py/src/braintrust/framework.py
@@ -83,6 +83,9 @@ class EvalCase(SerializableDataClass, Generic[Input, Output]):
     _xact_id: str | None = None
     created: str | None = None
 
+    # The number of times to run the evaluator for this specific input. 
+    trial_count: int | None = None
+
 
 class _EvalCaseDictNoOutput(Generic[Input], TypedDict):
     """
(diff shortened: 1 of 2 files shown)


lforst:
Hi, this unfortunately became stale because we moved the python sdk into its own repo: https://github.com/braintrustdata/braintrust-sdk-python

We can think about re-introducing the change there.

PR #1343, opened 2026-02-05
Prevent plugins from being overwritten in CLI bundle function

joshuawootonn:


joshuawootonn:
https://github.com/user-attachments/assets/d78977d2-e04c-4fd3-9419-e79f7e9a08d4

PR #1347, opened 2026-02-05
add new cli tests and fix eval to with with cjs, esm, deno, and bun

ibolmo:
We received two reports that our cli evals (and push) don't play well with `esm`. 

This adds a new `cli-tests` and scenarios, and updates our `eval` to run the evals with particular runners based on the user's code.

PR #1349, opened 2026-02-06
add framework tests (i.e. vitest)

ibolmo:
this should eventually help answer do our evals work in `vitest`, `jest` etc.

### DIFF
--- a/js/tsconfig.json
+++ b/js/tsconfig.json
@@ -16,6 +16,7 @@
     "**/examples/**",
     "node_modules/**",
     "**/smoke/**",
+    "cli-tests/**",
     "scripts/**",
     "vendor/**",
     "eslint.config.ts",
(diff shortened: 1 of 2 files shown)


js/framework-tests/scenarios/deno/tests/top-level-await.eval.ts
```diff
@@ -0,0 +1,23 @@
+import { Eval } from "braintrust";
+
+// Top-level await - ESM-only feature
+// This proves Deno can handle async module evaluation
+const config = await Promise.resolve({ prefix: "Result: " });
```
ankrgyl: does this need to be duplicated for each runtime? or can we make the runtime a separate axis?

ankrgyl: even in that change there are several instances of essentially the same file, https://github.com/braintrustdata/braintrust-sdk/pull/1347/changes#diff-a294c96468f18c6afc2645b696eeb0dc99b74ae03ca2ec61d8d8238f125c5b31

PR #1355, opened 2026-02-10
Move JS version to plain text VERSION file

clutchski:
## Summary
- Adds `js/VERSION` as the single source of truth for the JS SDK version, avoiding CODEOWNERS conflicts on `js/package.json` for version bumps
- Sets `js/package.json` version to `0.0.0-dev` placeholder; real version is injected at publish time via a new `_sync-version` Makefile target
- Updates all release/publish scripts and CI workflow to read from `VERSION` instead of `package.json`

## Test plan
- [ ] `cd js && cat VERSION` shows `2.2.0`
- [ ] `cd js && make _sync-version && node -p "require('./package.json').version"` shows `2.2.0`
- [ ] `cd js && git checkout package.json` restores placeholder after testing
- [ ] `cd js && ./scripts/push-release-tag.sh --dry-run` reads version from VERSION file
- [ ] Verify CI workflow dispatch pre-release path reads from VERSION


js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "braintrust",
-  "version": "2.2.1",
+  "version": "0.0.0-dev",
```
ibolmo: not recommended. what are we trying to solve?

PR #1359, opened 2026-02-11
Add public name property to Span interface

cjgalione:
## Summary

- Add a public `name` property to the `Span` abstract class so users can read the current span's name for logging, diagnostics, and conditional logic
- Implement in `SpanImpl` (backed by `self._name`, kept in sync by `set_attributes`)
- Implement in `_NoopSpan` (returns `None`)
- No schema changes, no API changes, purely additive

## Context

Users working with `traced()` and `start_span()` can *set* a span name via `span.set_attributes(name="...")` but cannot *read* it back. This adds `span.name` so patterns like `current_span().name` work for logging and debugging.

Pylon-Issue-ID: 6761467d-c2cf-4cac-bb43-db10abde5ca1

## Test plan

- [x] `test_span_name_returns_explicit_name` — name passed to `start_span()` is readable
- [x] `test_span_name_returns_inferred_root_name` — root span defaults to `"root"`
- [x] `test_span_name_returns_inferred_subspan_name` — child span gets caller-location name
- [x] `test_span_name_updated_by_set_attributes` — `.name` reflects `set_attributes()` updates
- [x] `test_span_name_consistent_with_logged_data` — property matches logged `span_attributes["name"]`
- [x] `test_noop_span_name_returns_none` — `NOOP_SPAN.name` returns `None`
- [x] `test_current_span_name_accessible` — `current_span().name` works
- [x] `test_traced_decorator_span_name` — `@traced` auto-sets name to function name

### DIFF
--- a/py/src/braintrust/logger.py
+++ b/py/src/braintrust/logger.py
@@ -162,7 +162,7 @@ class Span(Exportable, contextlib.AbstractContextManager, ABC):
 
     @property
     @abstractmethod
-    def name(self) -> str | None:
+    def name(self) -> str:
         """Name of the span, for display purposes only."""
 
     @abstractmethod
@@ -4067,7 +4067,7 @@ class SpanImpl(Span):
         return self._id
 
     @property
-    def name(self) -> str | None:
+    def name(self) -> str:
         return self._name
 
     def set_attributes(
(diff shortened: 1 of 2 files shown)


py/src/braintrust/logger.py
```diff
@@ -4055,12 +4066,18 @@ def __init__(
     def id(self) -> str:
         return self._id
 
+    @property
+    def name(self) -> str | None:
```
clutchski: same here

PR #1361, opened 2026-02-12
Add child span handling to OpenAIAgentsTraceProcessor

colinbennettbrain:
The fix updates OpenAIAgentsTraceProcessor.onTraceEnd() to end any child spans still tracked for the trace before deleting trace state, so spans don’t stay “in progress” if onTraceEnd runs before deferred onSpanEnd. It also adds regression tests covering the race path, late onSpanEnd no-op behavior, and multiple remaining child spans being closed correctly.


lforst:
superseded by #1813

PR #1363, opened 2026-02-12
Initialize parameter values to default values

joshuawootonn:
Related PR: https://github.com/braintrustdata/braintrust/pull/10788

After some recent discussions, I've realized that parameter data should be initialized to your default values. This is largely because when we open a parameter within our UI, we want to show the default values within forms without rendering forms in a dirty state. 

Here is a demo of parameter creation:

https://github.com/user-attachments/assets/0bd28948-843b-41c0-bf0d-e1a2d39ccce2

Bumped the SDK version because I would like to release and share this with customers ASAP.

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "braintrust",
-  "version": "2.2.1",
+  "version": "2.2.2",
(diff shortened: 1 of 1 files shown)


ibolmo (approved):
CO

PR #1370, opened 2026-02-12
Fix ObjectFetcher internal BTQL cursor override during pagination

colinbennettbrain:
## Summary
- Prevent `_internal_btql.cursor` from overriding the loop-managed pagination cursor in `ObjectFetcher` BTQL fetches.
- Preserve `_internal_btql.limit` overrides while keeping default `batchSize` behavior for normal fetches.
- Add regression tests for internal limit behavior and multi-page cursor advancement.
- Include existing branch dependency metadata updates in `js/package.json` and `pnpm-lock.yaml` (`jiti` resolution changes).

### DIFF
--- a/js/package.json
+++ b/js/package.json
@@ -99,6 +99,7 @@
     "async": "^3.2.5",
     "autoevals": "^0.0.131",
     "cross-env": "^7.0.3",
+    "jiti": "^2.6.1",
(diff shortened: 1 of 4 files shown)


lforst:
Closing because we likely do not want to add any additional stuff to internal. We want our backend to use only public apis of the SDK.

PR #1371, opened 2026-02-12
replace PEP 585 tuple annotations in pydantic_ai

colinbennettbrain:
Replace PEP 585-style `tuple[...]` return annotations in `py/src/braintrust/wrappers/pydantic_ai.py` with `Tuple[...]` to preserve Python 3.8 compatibility for existing users.  Adds `Tuple` import and updates five return annotations only.


ibolmo:
we have a min. of 3.10 .. 3.8 and 3.9 are past EOL.

colinbennettbrain:
Resolved as not needed: SDK minimum Python version is 3.10+, so older-version compatibility workaround is out of scope. Closing.

PR #1372, opened 2026-02-12
replace PEP 585 tuple annotations in pydantic_ai

colinbennettbrain:
## Summary

- Replace PEP 585-style `tuple[...]` return annotations in `py/src/braintrust/wrappers/pydantic_ai.py` with `Tuple[...]`.
- Add `Tuple` import and update five return annotations only.
- This is a no-behavior-change Python 3.8 compatibility fix.

Commit: `0045dda1`


colinbennettbrain:
Closing: min supported Python is 3.10+, so 3.8/3.9 compatibility fix is no longer needed.

PR #1383, opened 2026-02-17
Small pending improvements to bt 3.x

ibolmo:


cpinn:
I guess I got confused as I thought we agreed to not have the duplicated empty files that had the same behavior and just expose the different entry points. 

I am okay with this as it is is similar to what I originally had here: https://github.com/braintrustdata/braintrust-sdk/pull/1307/changes/BASE..249df97b5faa9faf817e15b8e1407c5ae0c75abf#diff-a0627437f0d10f5ff9edd541155f12f047785417bd803a4dab353b4733d51ea8

ibolmo:
> I guess I got confused as I thought we agreed to not have the duplicated empty files that had the same behavior and just expose the different entry points.
> 
> I am okay with this as it is is similar to what I originally had here: https://github.com/braintrustdata/braintrust-sdk/pull/1307/changes/BASE..249df97b5faa9faf817e15b8e1407c5ae0c75abf#diff-a0627437f0d10f5ff9edd541155f12f047785417bd803a4dab353b4733d51ea8

yeah exactly. not sure when we dropped that, but i'm hugely in favor. there's more surface area to upkeep, but it's a very low touch area. might as well be explicit here.

PR #1388, opened 2026-02-18
Remove als-browser from the newest release

cpinn:
Remove the als-browser from the newest release


ibolmo (approved):
i think we're saying we're going to revisit the als browser in the future, but for now avoiding this right now

PR #1396, opened 2026-02-20
test: Fix hanging google-genai tests

AbhiPrasad:
# Background

`google-genai` `v1.64.0` (released Feb 19) added `aiohttp` as a dependency. When `aiohttp` is available, the async requests use `aiohttp` instead of `httpx`. VCR.py's `aiohttp` stubs have a bug where:

1. `response_stream.content` returns a new `MockStream` every time it's accessed (it's a `@property`, not `@cached_property`)
3. `google-genai`'s streaming code calls `await self.response_stream.content.readline()` in a `while True` loop
5. Each access creates a fresh stream that starts from the beginning, so `readline()` keeps returning the first line forever — infinite loop

This makes the Python `google-genai` tests hang. The sync tests pass (they use `httpx` which VCR handles fine), but the async streaming tests cause an infinite loop.

Why it started failing
-  Last passing run (Feb 17): installed `google-genai<=1.63.0`, which only used `httpx` — VCR works fine
- First failing run (Feb 19): installed `google-genai==1.64.0`, which brings in `aiohttp` — triggers the VCR bug

# Fix

`py/src/braintrust/conftest.py` — added `_patch_vcr_aiohttp_stubs()` that monkey-patches VCR.py's aiohttp stubs at import time. Three fixes:

1. **`MockClientResponse.content`** — changed from `@property` (creates new `MockStream` every access) to a cached version. This fixes the infinite loop where `google-genai`'s `while True: await response.content.readline()` never reached EOF.

2. **`MockClientResponse.text()` and `read()`** — added gzip decompression. Cassettes store gzip-compressed bodies; VCR's httpx stubs decompress automatically but the aiohttp stubs don't, causing `UnicodeDecodeError`.

3. **`MockStream.set_exception()`** — made it a no-op. `build_response()` calls `response.close()` which sets a `ClientConnectionError` on the stream, poisoning it for subsequent reads.

The patch is safe:
- Guarded behind `try/except ImportError` — no-op if aiohttp/vcr aren't installed
- Idempotent — `_bt_patched` flag prevents double-patching
- Only affects VCR mock objects, not real aiohttp behavior

# Aside

After implementing this fix, another issue with the temporal integration was found. This was also fixed, but is unrelated to the `google-genai` related changes.

All 12 tests pass on both temporalio 1.19.0 and 1.23.0. Here's a summary of both fixes:

`py/src/braintrust/contrib/temporal/__init__.py` — temporalio 1.23.0 compat.

`SimplePlugin.__init__()` merged `client_interceptors` + `worker_interceptors` into a single `interceptors` parameter in temporalio 1.23.0. The fix inspects the signature at runtime and uses the appropriate parameter, maintaining compatibility with both old (1.19.0+) and new (1.23.0) versions.

### DIFF
--- a/py/src/braintrust/conftest.py
+++ b/py/src/braintrust/conftest.py
@@ -39,22 +39,18 @@ def _patch_vcr_aiohttp_stubs():
             return gzip.decompress(body)
         return body
 
-    # No-op set_exception so close() doesn't poison the stream.
     aiohttp_stubs.MockStream.set_exception = lambda self, exc: None
 
-    # Override text() to decompress gzip before decoding.
     async def patched_text(self, encoding="utf-8", errors="strict"):
         return _decompress_body(self._body).decode(encoding, errors=errors)
 
     aiohttp_stubs.MockClientResponse.text = patched_text
 
-    # Override read() to decompress gzip.
     async def patched_read(self):
         return _decompress_body(self._body)
 
     aiohttp_stubs.MockClientResponse.read = patched_read
 
-    # Cache content stream per instance and feed decompressed data.
     @property
     def cached_content(self):
         if not hasattr(self, "_cached_content"):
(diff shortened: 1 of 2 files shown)


nselvidge (approved):
minor nitpick about ai comments. I see you commented on the issue in the vcrpy repo, maybe keep an eye out in case that issue gets resolved so we can remove this? (not the most robust way to keep track but seems like the best we can do). I'll also follow the issue.

PR #1397, opened 2026-02-21
Evals(py): Allow passing tags for evals and experiments

Mahhheshh:
#1389 for Python


AbhiPrasad:
Hey @Mahhheshh, thanks for opening a PR! CI atm is a bit red, so doing some work to unblock that. As soon as that is good, we can get this merged in!

PR #1403, opened 2026-02-24
meta: Remove Python related code from this repo

AbhiPrasad:
This PR removes all js related code from this codebase, as it was a remanent of the transition from https://github.com/braintrustdata/braintrust-sdk.

This PR also replaces pre-commit with husky, which means we can drop all python usage in this repo.

ref https://github.com/braintrustdata/braintrust/pull/11071

plan: https://www.notion.so/braintrustdata/Splitting-up-braintrust-sdk-30df7858028980298731d3827a23cb70


.github/workflows/lint.yaml
```diff
@@ -1,20 +0,0 @@
...
-    runs-on: ubuntu-latest
-    timeout-minutes: 10
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0 # Fetch full history for proper diff
-      - uses: actions/setup-python@v3
-      - uses: pre-commit/action@v3.0.0
-        with:
-          extra_args: --from-ref origin/${{ github.base_ref || 'main' }} --to-ref HEAD
```
Qard: Does the pre-commit lint flow not apply to JS too?

PR #1481, opened 2026-03-03
feat: Add node test runner integration

AbhiPrasad:
resolves https://github.com/braintrustdata/braintrust-sdk-javascript/issues/1478

Adds `initNodeTestSuite`, a composable integration for `node:test` that tracks test results as Braintrust experiments. This is a little different than `wrapVitest` from the vitest integration as it doesn't require you to wrap all the exports. Instead we elected for an approach where you call `initNodeTestSuite` for a describe block, and then use the `suite.eval` helper function in tests.

```js
import { test, describe, after } from "node:test";
import { initNodeTestSuite } from "braintrust";

describe("My Evaluation Suite", () => {
  const suite = initNodeTestSuite({
    projectName: "my-project",
    after, // Auto-flush when tests complete
  });

  test(
    "evaluates output",
    suite.eval(
      {
        input: { text: "hello" },
        expected: "hola",
      },
      async ({ input }) => {
        const result = await translate(input.text);
        return result; // Return value is logged as output
      },
    ),
  );

  // Untracked tests work normally — no Braintrust involvement
  test("sanity check", () => {
    assert.strictEqual(1 + 1, 2);
  });
});
```

After tests run, you get an experiment summary. This is formatted similarly to the vitest integration's experiment summary.

```
┌─ Braintrust Experiment Summary ─────────────────┐
│ Experiment: my-project-2026-03-03T16:13:31.826Z
│
│ Scores:
│   exact_match: 100.00%
│   pass: 100.00%
│
│ View results: https://www.braintrust.dev/app/.../experiments/...
└──────────────────────────────────────────────────┘
```

Unlike the vitest integration, the suite does not expose `logOutputs()`, `logFeedback()`, and `getCurrentSpan()`. This is because users can just easily use `currentSpan().log` instead. 

It doesn't import `node:test` itself. It only uses `t.name` from the test context and accepts `after` as a config param so it works with any `node:test`-compatible runner. This means we can easily add runners for deno and bun 🔜.

You'll notice that we moved some utilities into shared helpers. Here is a Claude summary:

- **Traced eval loop** — run test fn, run scorers, log pass/fail + output + errors. Scorers run concurrently via `Promise.all` and execute once regardless of pass/fail (not duplicated in try + catch).
- **Flush** — summarize + flush, avoiding double-flush since `summarize()` calls `flush()` internally.
- **Scorer runner, formatter, types** — moved from vitest-specific to shared.

On the way, Claude caught some bug fixes:

- **vitest: `displaySummary: false` skipped flushing entirely** — The `afterAll` hook was gated on `displaySummary`, meaning experiments were never flushed when summary display was off.
- **scorers: array results not validated** — `normalizeScores` now applies `isScore()` to array elements instead of only filtering nulls.
- **Error logging** — Test failures now use the existing `logError()` utility to write to the top-level `error` field on spans, consistent with all other wrappers in the codebase.

### DIFF
--- a/js/src/wrappers/node-test/test-helpers.ts
+++ b/js/src/wrappers/node-test/test-helpers.ts
@@ -31,9 +31,9 @@ export async function setupNodeTestEnv(): Promise<TestBackgroundLogger> {
   return bgLogger;
 }
 
-export async function teardownNodeTestEnv(): Promise<void> {
-  await _exportsForTestingOnly.clearTestBackgroundLogger();
-  await _exportsForTestingOnly.simulateLogoutForTests();
+export function teardownNodeTestEnv(): void {
+  _exportsForTestingOnly.clearTestBackgroundLogger();
+  _exportsForTestingOnly.simulateLogoutForTests();
 }
 
 export function mockTestContext(name: string): MinimalTestContext {


js/src/wrappers/shared/flush.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Summarize and flush an experiment.
+ *
+ * `summarize()` calls `flush()` internally
+ */
+export async function summarizeAndFlush(
+  experiment: Experiment,
+  options: { displaySummary?: boolean },
+): Promise<void> {
+  const shouldDisplay = options.displaySummary ?? true;
+  const summary = await experiment.summarize();
```
lforst: It seems like that previously we flushed explicitly and also had some try-catching. Was this change intentional?

lforst (approved):

cpinn (approved):

PR #1499, opened 2026-03-04
ref: Use diagnostic channels for OpenAI SDK wrapper

lforst:
Moves the implementation for the `wrapOpenAI` wrapper over to use diagnostic channels that are picked up by the plugins.

Ref: https://github.com/braintrustdata/braintrust-sdk-javascript/issues/1483

### DIFF
--- a/js/src/wrappers/oai.ts
+++ b/js/src/wrappers/oai.ts
@@ -189,7 +189,7 @@ function wrapBetaChatCompletionParse<
     );
     return channel.tracePromise(
       // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-      () => completion(params as P),
+      async () => await completion(params as P),
       { arguments: [params], span_info },
     );
   };


js/src/wrappers/oai.ts
```diff
@@ -1,16 +1,20 @@
...
-} from "../logger";
-import { getCurrentUnixTimestamp, isEmpty } from "../util";
-import { mergeDicts } from "../../util/index";
-import { responsesProxy, parseMetricsFromUsage } from "./oai_responses";
+  LEGACY_CACHED_HEADER,
+  parseCachedHeader,
+  X_CACHED_HEADER,
+} from "../openai-utils";
+import { responsesProxy } from "./oai_responses";
+import { tracingChannel } from "dc-browser";
```
Qard: We should not be import `dc-browser` directly like this anywhere. See my note in #1501.

lforst: Oh right, for some dumb reason I thought it was already isomorphic. Makes sense.

PR #1544, opened 2026-03-11
chore: Update release process to use gha trigger

AbhiPrasad:
This PR makes some small improvements to our release process, notably making it easier to publish stable releases via the github UI (action triggers). It also addresses an error scenario where it will only publish a tag on the repo after a successful npm publish. `PUBLISHING_JS.md` has the full set of details of the new publish workflow.

Publishing a main release is now gated behind an approver for the action. This is the current list of folks who can approve a release. Wasn't 100% sure of this, but figured it's nice for safety now that both pre-releases and stable releases go through github triggers. Approvers don't have to get another approver to run their release.

image

This is done in prep for https://github.com/braintrustdata/braintrust-sdk-javascript/issues/1543

### DIFF
--- a/.github/workflows/publish-js-sdk.yaml
+++ b/.github/workflows/publish-js-sdk.yaml
@@ -18,6 +18,10 @@
 
 name: Publish JavaScript SDK
 
+concurrency:
+  group: publish-js-sdk-${{ inputs.release_type }}-${{ inputs.branch }}
+  cancel-in-progress: false
+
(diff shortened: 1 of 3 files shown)


AbhiPrasad:
yeah I'll think about this after the nightly. We can probably just use something like https://github.com/googleapis/release-please eventually and remove most of the custom stuff here.

(in repo changelog also super useful).

PR #1553, opened 2026-03-12
Add support for classifier "scorers"

aswink:


AbhiPrasad (commented):
This also needs a bunch more tests, but we can tackle that after this.

lforst:
I looked into this PR a bit from a user perspective and I found it quite confusing to have both `id` and `label`. What is the behavior of label? What if the label doesn't match `id`? IMO it would make most sense to have one singular required param named `slug`. How locked in is this design?

Edit: Also how does `name` fit into the story? makes it even more confusing...

js/util/score.ts
```diff
@@ -1,3 +1,23 @@
+/**
+ * The result returned by a classifier function. Unlike `Score`, `id` is
+ * required and the span will be recorded as a classifier span.
+ */
+export interface Classification {
+  name: string;
+  id: string;
+  label?: string;
```
lforst, line 23: Would love jsdoc for all of these fields. As a user I'd have a tough time knowing what to provide for them.

lforst:
I think the way we are changing `Evaluator` is a breaking change 🤔

js/util/score.ts
aswink: added a comment!

lforst (approved):

PR #1558, opened 2026-03-13
fix: Make sure params get passed through for compiled prompt

AbhiPrasad:
resolves https://linear.app/braintrustdata/issue/BT-4378/typescript-sdk-promptbuild-fields-response-format-tool-choice

This fixes a bug with typings where params were not being passed through correctly.

```ts
import OpenAI from "openai";
import { Prompt } from "braintrust";

declare const prompt: Prompt;

const params = prompt.build(
  { topic: "TypeScript" },
  { flavor: "chat" },
);

// Before the fix, these would fail
// Property 'temperature' does not exist on type 'CompiledPrompt<"chat">'
const temperature = params.temperature;
const responseFormat = params.response_format;
const toolChoice = params.tool_choice;
```

I didn't add a test for this because it's just a types change, but lmk and I can add a set of type check tests.


AbhiPrasad:
Ah okay we are fine. I over though it. We did break `params.response_format.json_schema.schema`, but that would have been busted before anyway. So this change is safe.

PR #1567, opened 2026-03-15
fix: resolve edge AsyncLocalStorage from runtime builtins

dmmulroy:
## Summary
- resolve `AsyncLocalStorage` for the `workerd` and `edge-light` builds from either the global or `process.getBuiltinModule(\"node:async_hooks\")`
- add `js/src/edge-runtime-bootstrap.test.ts` as a minimal repro + regression test, then harden it to cover accessor-backed `getBuiltinModule` and reset the symbol-backed global state between cases
- flip the affected Cloudflare compat smoke scenarios from expected failure to passing for ALS-backed parent propagation

## Minimal Repro
I included the minimal repro inline here and as the committed regression test in `js/src/edge-runtime-bootstrap.test.ts`.

```js
import * as asyncHooks from "node:async_hooks";

// Reproduce an edge/workerd-style runtime where AsyncLocalStorage is not
// exposed as a global constructor.
Reflect.deleteProperty(globalThis, "AsyncLocalStorage");

// Reproduce the alternative runtime path this patch now supports:
// AsyncLocalStorage is only reachable through process.getBuiltinModule().
Object.defineProperty(process, "getBuiltinModule", {
  configurable: true,
  get: () => (id) =>
    id === "node:async_hooks" ? asyncHooks : undefined,
});

// Import after patching the runtime surface so the SDK bootstrap sees the
// same environment shape that was previously failing.
const braintrust = await import("braintrust/workerd");

braintrust._exportsForTestingOnly.setInitialTestState();
await braintrust._exportsForTestingOnly.simulateLoginForTests();

const logger = braintrust.initLogger({
  projectName: "als-repro",
  projectId: "test-project-id",
});

const root = logger.startSpan({ name: "root", type: "task" });
const parent = await root.export();

const result = await braintrust.traced(
  async (span) => {
    // Before the fix, currentSpan() could still resolve to NOOP_SPAN here
    // even though traced() passed a real span into the callback.
    const active = braintrust.currentSpan();
    const child = braintrust.startSpan({ name: "child", type: "task" });
    child.end();

    return {
      isNoop: Object.is(active, braintrust.NOOP_SPAN),
      sameObject: Object.is(active, span),
      childParents: child.spanParents,
    };
  },
  { parent, name: "chat", type: "task" },
);

// After the fix, this logs something like:
// { isNoop: false, sameObject: true, childParents: ["<chat span id>"] }
// Before the fix, the same script could log:
// { isNoop: true, sameObject: false, childParents: [] }
console.log(result);
```

`childParents` contains the active `chat` span id, so the exact string varies per run. The important part is that it goes from an empty array before the fix to a single parent span id after the fix.

Before this patch, `currentSpan()` could still resolve to `NOOP_SPAN` in this setup, so `sameObject` was false and `childParents` was empty even though `traced()` passed a real span into the callback. After this patch, `currentSpan()` resolves to the active span and the child span inherits the expected parent/root context.

## Verification
- `pnpm --dir js exec vitest run src/edge-runtime-bootstrap.test.ts`
- `pnpm --dir js exec tsc --noEmit`
- `pnpm --dir js build`

### DIFF
--- a/js/smoke/scenarios/cloudflare-worker-browser-compat/README.md
+++ b/js/smoke/scenarios/cloudflare-worker-browser-compat/README.md
@@ -1,20 +1,20 @@
 # Cloudflare Worker Browser Compat Scenario
 
-Tests Braintrust SDK in Cloudflare Workers with `braintrust/browser` entrypoint with `nodejs_compat_v2` enabled.
+Tests Braintrust SDK in Cloudflare Workers with the top-level `braintrust` import resolving to the `workerd` build, with `nodejs_compat_v2` enabled.
 
 ## Design Decisions
 
 ### Import Path
 
-Uses `braintrust/browser` (browser entrypoint) to test browser-compatible code with Node.js APIs available.
+Uses `braintrust` and asserts that Cloudflare resolves it to the `workerd` build in this runtime.
 
 ### Compatibility Flags
 
-`nodejs_compat_v2` is enabled. While the browser entrypoint doesn't require Node.js APIs, this tests that the browser entrypoint works correctly when Node.js compatibility is present. This is useful for scenarios where other parts of the application need Node.js APIs.
+`nodejs_compat_v2` is enabled to verify that the default Cloudflare Workers resolution still behaves correctly when Node.js compatibility is present.
 
 ### Expected Outcome
 
-All tests should pass. The browser entrypoint should work correctly regardless of whether Node.js compatibility is enabled or not.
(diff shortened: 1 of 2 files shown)


AbhiPrasad (commented):
This looks good to me (although I wish cloudflare made this easier on the runtime side). CI will probably fail on the fork, but we can cherry-pick the commit onto a new branch and PR it in.

PR #1617, opened 2026-03-20
fix(auto-instrumentation): Fully preserve returned promise in tracing channel hook

lforst:
Fix the auto-instrumentation node hook so Anthropic `messages.stream()` keeps working under `node --import braintrust/hook.mjs`.

- Preserve the original promise-like return value in the hook instead of returning a chained promise
- Keep Anthropic `APIPromise` helpers like `withResponse()` intact
- Remove the e2e workaround so the Anthropic node-hook scenario uses `messages.stream()` again
- Add regression coverage for promise subclass helper preservation

Fixes https://github.com/braintrustdata/braintrust-sdk-javascript/issues/1591


Qard (commented):
This change would be incorrect though. It also breaks behavioural compatibility with Node.js core.

PR #1618, opened 2026-03-20
fix: Skip promise chaining when original promise has additional fields

lforst:
OpenAI and Anthropic SDK functions sometimes return Promises (having Promise prototype) with additional fields. Whenever we wrap these Promises with a chain (`.then()`) the return value will not have the additional field, breaking API contracts.

To mitigate we have previously introduced a check to compare the prototype to `Promise` to make sure that we only chain actual promises, however that was only enough for OpenAI. Anthropic seems to return actual Promises with additional fields so we gotta add an additional check that checks for additional ownproperties.

### DIFF
--- a/js/src/auto-instrumentations/patch-tracing-channel.ts
+++ b/js/src/auto-instrumentations/patch-tracing-channel.ts
@@ -105,11 +105,11 @@ export function patchTracingChannel(
               isPlainNativePromiseWithoutHelpers(result)
             ) {
               return result.then(
-                (res) => {
+                (res: unknown) => {
                   publishResolved(res);
                   return res;
                 },
-                (err) => {
+                (err: unknown) => {
                   publishRejected(err);
                   return Promise.reject(err);
                 },


Qard:
We already addressed the thenables problem though, so that should be a non-issue. As for the async iterator thing, is the sync result itself directly an async iterator? Or is that wrapped in a promise? I'm working on some improvements to tracing channel in Node.js core to address the lack of async iterator support. It would be good if we try to align with that. Namely, we should still be producing the asyncStart and asyncEnd around the overall boundaries, and should have boundary events around each yield too ideally.

The benefit of conforming to the diagnostics_channel model is that you get discrete event pairs which can be understood more generically rather than having all this sort of code which ties deeply into the implementation details of the code and is therefore brittle to changes to the API surface. If the code changes with direct patching like this the patches could break. If the code changes with diagnostics_channel either the channel publishing changes with it or the channels simply don't emit those events anymore and so nothing happens.

What I had done for these things previously was just patch the received async iterators in the diagnostics_channel handlers we _did_ have provided as we do not have the async iterator mechanisms yet. But we _should_ still express the full async lifecycle though because that's what indicates when to stop a corresponding span, right? It's _supposed_ to be the case that we can simply use the diagnostics_channel events directly for starting/stopping spans and then we can augment that with additional channel events to capture other details like errors or logs.

While `tracePromise` might be insufficient for the async iterator part, I'm not seeing why it's not at least usable for the _rest_ of that and we can just deviate minimally for only the async iterator need?

PR #1640, opened 2026-03-23
chore(ci): Extend timeout to try and fix flakey ci

lforst:
Tries to fix random failures for the vitest tests.


lforst:
> Approving just to fix it for now, but we might want to look more deeply at why they take so long that they need the timeout extended. 🤔

I agree. I think that they're mostly slow due to the version matrix installs and not any majorly dumb reason.

PR #1696, opened 2026-03-27
fix(anthropic): enable messages.stream() in auto-instrumentation e2e tests

### LINKED ISSUE #1592: Auto-instrumentation via node hook breaks Anthropic messages.stream()
## Summary
Under `node --import braintrust/hook.mjs`, Anthropic `client.messages.stream(...)` breaks even though the equivalent `messages.create({ stream: true })` flow still works.

## What is broken
In auto-instrumented mode, `messages.stream(...)` fails because the helper path goes through an internal `create(...).withResponse()` flow that no longer behaves correctly once instrumented.

## Why this matters
This is an API-compatibility bug in the auto-instrumented path. Valid Anthropic SDK code works normally, but breaks when Braintrust auto-instrumentation is enabled.

## Repro
The aligned e2e coverage in `e2e/scenarios/anthropic-auto-instrumentation-node-hook` reproduces this. The failure we hit was:
- `messages.create(...).withResponse is not a function`

The current e2e workaround is to avoid `messages.stream(...)` in auto mode and use `messages.create({ stream: true })` instead.

## Expected
`client.messages.stream(...)` should continue to work under the node-hook auto-instrumented path without requiring a scenario-specific fallback.

Qard:
## Summary

Closes #1592.

The root cause of this issue was fixed by PR #1617, which patched `TracingChannel.prototype.tracePromise` to preserve the original `APIPromise` object rather than wrapping it in a plain `.then()` chain that strips `.withResponse()`.

This PR removes the e2e workaround that was preventing `client.messages.stream()` from being tested in the auto-hook instrumentation path.

## What was the workaround?

`runAutoAnthropicInstrumentation` was forcing `useMessagesStreamHelper: false`, which caused the `"anthropic-stream-with-response-operation"` scenario to use `client.messages.create({ stream: true })` instead of `client.messages.stream()`. This avoided the failure but left the real code path untested.

## Why is the fix correct?

`client.messages.stream()` internally calls `this.create({...}).withResponse()`. The instrumented `create` returns the original `APIPromise` (not a wrapped plain Promise), so `.withResponse()` is still available. The `APIPromise` constructor check (`result.constructor !== Promise`) correctly routes to the preserve-original branch in the patched `tracePromise`.

All 140 anthropic e2e tests pass with both wrapper and auto-hook instrumentation paths using `messages.stream()`.

## Test plan

- [x] All 140 anthropic e2e tests pass with `messages.stream()` enabled (`pnpm vitest run scenarios/anthropic-instrumentation/scenario.test.ts`)
- [ ] CI e2e and js-test checks pass


e2e/scenarios/anthropic-instrumentation/scenario.impl.mjs
```diff
@@ -251,7 +251,6 @@ export async function runWrappedAnthropicInstrumentation(Anthropic, options) {
 export async function runAutoAnthropicInstrumentation(Anthropic, options) {
   await runAnthropicInstrumentationScenario(Anthropic, {
     ...options,
-    useMessagesStreamHelper: false,
```
lforst: we should probably also remove the code that used this argument since it's only used once I think

PR #1761, opened 2026-04-08
feat: Add `@openrouter/agent` instrumentation

lforst:
Basically a carbon copy of the `@openrouter/sdk` instrumentation. We got unlucky with timing that instrumentation because it seems like they are re-architecting stuff.


AbhiPrasad (approved):
So do we deprecate `@openrouter/sdk` instrumentation?

PR #1763, opened 2026-04-08
chore: Pin models and increase max tokens in e2e/canary tests

lforst:
Closes https://github.com/braintrustdata/braintrust-sdk-javascript/issues/1760


AbhiPrasad (approved):
I assume we'll do a follow up to add a retry to `withScenarioHarness`? Then we can remove per assertion retries in favour of just retrying the entire scenario.

lforst:
> I assume we'll do a follow up to add a retry to `withScenarioHarness`? Then we can remove per assertion retries in favour of just retrying the entire scenario.

Yeah I am vibing on a more proper but elaborate fix here where we migrate to more granular testing, putting everything into the tests and not the beforeAll hook.

Secondary reason for that is that the tests are starting to take up 20 minutes, and I would like to shard them up and that only really makes sense if we do the whole scenario shebang in the tests and not the before all hook.

PR #1769, opened 2026-04-09
fix: Give AI SDK top-level api spans type `function`

lforst:
Ref: https://linear.app/braintrustdata/issue/BT-4717

Turns span type of top-level spans for ai sdk into `function` instead of `llm` to fix the product "double counting" the duration as LLM duration.


AbhiPrasad (approved):
Like you were saying, we probably need to write down the general rules for this somewhere. Even just a markdown in https://github.com/braintrustdata/braintrust-spec is a good start.

PR #1796, opened 2026-04-10
perf: Remove zod from deepCopyEvent

ankrgyl:
This is called on the hot path.


js/src/logger.ts
```diff
@@ -5466,22 +5467,61 @@ export function deepCopyEvent<T extends Partial<BackgroundLogEvent>>(
...
+  number
+> {
+  if (!value || typeof value !== "object" || Array.isArray(value)) {
+    return false;
+  }
+
+  const prototype = Object.getPrototypeOf(value);
+  if (prototype !== Object.prototype && prototype !== null) {
+    return false;
+  }
```
AbhiPrasad: the `JSON.parse` reviver will always get valid POJOs, so we don't need to include the prototype checks here

```suggestion
```

js/src/logger-misc.test.ts
```diff
@@ -180,6 +180,44 @@ test("deepCopyEvent basic", () => {
...
+  expect(copy.output.externalAttachment).toBe(externalAttachment);
+  expect(copy.output.nested[0]).toBe(attachment);
+  expect(copy.output.nested[1]).toBe(externalAttachment);
+  expect(copy.output.fake).toEqual({
+    _bt_internal_saved_attachment_idx: 0,
+    _bt_internal_saved_attachment_marker: 0,
+  });
+});
+
 test.skip("deepCopyEvent with attachments", () => {
```
AbhiPrasad: can we just delete this skipped test?

js/src/logger.ts
```diff
@@ -5466,22 +5467,61 @@ export function deepCopyEvent<T extends Partial<BackgroundLogEvent>>(
...
+  ) {
+    return false;
+  }
+
+  if (Object.keys(record).length !== 2) {
+    return false;
+  }
+
+  return (
+    Number.isInteger(record._bt_internal_saved_attachment_idx) &&
```
AbhiPrasad: unfortunately `Number.isInteger` doesn't narrow the type 😢, so you'll have to cast as a number manually.

js/src/logger.ts
ankrgyl: fixed

js/src/logger-misc.test.ts
ankrgyl: fixed

js/src/logger.ts
ankrgyl: fixed

AbhiPrasad (approved):

PR #1837, opened 2026-04-15
feat: dataset versioning

max-braintrust:
### Summary

This PR adds dataset snapshot and environment tag support to the JS SDK. See feature spec here: https://github.com/braintrustdata/braintrust-spec/pull/14

### Background

This change adds two friendlier ways to reference dataset versions:

- Snapshots, which are stable human-readable names for a specific dataset version
- Environment tags, which are movable aliases like ppe or production that can be repointed over time

These are still just ways of referring to a concrete dataset version (xact_id). The SDK resolves snapshot names and environment tags down to the underlying xact_id before experiment or eval registration, so we keep the existing reproducibility guarantees while making version selection much easier to use.

### This PR adds:
- SDK support for initializing datasets by:
  - explicit version (xact_id)
  - snapshot name
  - environment tag
- Resolution of snapshot and environment selectors to a concrete dataset version internally before eval / experiment registration
- SDK helpers for dataset snapshots, including:
  - create
  - list
  - update via register/upsert for the current dataset version
  - patch snapshot metadata by id
  - delete
  - restore and restore/preview to return the dataset head to the state at a particular version
- Dev server support for forwarding dataset version and environment when resolving datasets for remote evals
- Tests and example coverage for the new version-selection paths


js/tests/api-compatibility/api-compatibility.test.ts
max-braintrust, line 1540: This heuristic fails if you are adding substantially to the method bodies of a class - added validation through the TS parser as a fallback when this fails. Since this should be faster, keeping it for the normal case.

js/src/logger.ts
AbhiPrasad, line 1: @lforst we really need to split this file up 😭

js/src/logger.ts
```diff
@@ -3621,20 +3656,13 @@ export function init<IsOpen extends boolean = false>(
...
-          // Full Dataset object
-          args["dataset_id"] = await (dataset as AnyDataset).id;
-          args["dataset_version"] = await (dataset as AnyDataset).version();
+        const datasetSelection = await serializeDatasetForExperiment({
+          dataset,
+          state,
+        });
+        args["dataset_id"] = datasetSelection.datasetId;
+        if (datasetSelection.datasetVersion !== undefined) {
+          args["dataset_version"] = datasetSelection.datasetVersion;
```
AbhiPrasad, line 3696: This is a change to how this worked before because we had the `} else {` branch that would do `args["dataset_version"] = await (dataset as AnyDataset).version();`. Is that intentional? I guess we do save on having to do the `dataset.version()` call.

js/src/logger.ts
```diff
@@ -6896,12 +7181,59 @@ export class Dataset<
...
+      this.lazyPinnedVersion !== undefined &&
+      this.getPinnedVersion() === undefined
+    ) {
+      this.setPinnedVersion(await this.lazyPinnedVersion.get());
+    }
     return this.state;
   }
 
+  public override async version(options?: { batchSize?: number }) {
+    await this.getState();
```
AbhiPrasad, line 7293: by always call `await this.getState` we lose the fast path we had before.

https://github.com/braintrustdata/braintrust-sdk-javascript/blob/343634f7c6f2e6fc81a0d9039cc8017740edc10c/js/src/logger.ts#L6107-L6109

is there a way we can check for that?

js/src/logger.ts
```diff
@@ -3805,13 +3833,197 @@ export type FullInitDatasetOptions<IsLegacyDataset extends boolean> = {
...
+  state: BraintrustState;
+  datasetId: string;
+  environment: string;
+}): Promise<string> {
+  const response = await state
+    .apiConn()
+    .get_json(
+      `environment-object/dataset/${datasetId}/${encodeURIComponent(environment)}`,
+      {
+        org_name: state.orgName ?? undefined,
```
AbhiPrasad: I think we want to avoid setting undefined in the URI component

js/src/logger.ts
```diff
@@ -3805,13 +3833,197 @@ export type FullInitDatasetOptions<IsLegacyDataset extends boolean> = {
...
+  state,
+  datasetId,
+  snapshotName,
+}: {
+  state: BraintrustState;
+  datasetId: string;
+  snapshotName: string;
+}): Promise<string> {
+  const snapshots = await getDatasetSnapshots({ state, datasetId });
+  const match = snapshots.find((snapshot) => snapshot.name === snapshotName);
```
AbhiPrasad: can we add a backend endpoint to do this instead? Feels like a lot to scan through all the datasets client-side.

js/src/logger.ts
```diff
@@ -6896,12 +7181,59 @@ export class Dataset<
     return this.state;
   }
 
+  public async toEvalData(): Promise<{
+    dataset_id: string;
+    dataset_version?: string;
+    dataset_environment?: string;
+    dataset_snapshot_name?: string;
```
AbhiPrasad, line 7244: do we need to pass `dataset_snapshot_name` into the remote evals created in https://github.com/braintrustdata/braintrust-sdk-javascript/blob/343634f7c6f2e6fc81a0d9039cc8017740edc10c/js/dev/server.ts#L312

(I lack a lot of context w/ remote evals so lmk if I'm off the mark!)

js/src/logger.ts
max-braintrust: No that should probably get fixed so this still works for subclasses/cases where the version is pinned manually - updated serializeDatasetForExperiment() so it will always hit .version() if we don't resolve through one of the other selections.

js/src/logger.ts
max-braintrust: added the check here as well so we avoid the slow path when the version is pinned already.

js/src/logger.ts
max-braintrust: Yes this should already be supported - pulled apart listSnapshots() and getSnapshot() so this makes use of that properly now.

js/src/logger.ts
max-braintrust: Discussed a bit offline - the remote eval path needs more api changes before it gets added in the sdk.

js/src/logger.ts
```diff
@@ -3390,13 +3414,55 @@ type InitOpenOption<IsOpen extends boolean> = {
...
+  snapshotName: string;
+  xactId?: never;
+};
+
+type DatasetSnapshotXactLookup = {
+  snapshotName?: never;
+  xactId: string;
+};
+
+type DatasetSnapshotLookup =
```
AbhiPrasad: can we export this? it's used by `public async getSnapshot`

AbhiPrasad (approved):

.changeset/goofy-hotels-care.md
```diff
@@ -0,0 +1,6 @@
+---
+"braintrust": minor
+---
+
+- (feat) Add dataset snapshot/environment selection support to `init()` and `initDataset()`, including snapshot CRUD helpers and `DatasetSnapshot` type exports.
+- (feat) Update `braintrust/dev` to respect `dataset_version` and `dataset_environment` when resolving datasets for evals.
```
AbhiPrasad: feel free to also add a little extra detail here, like an example code snippet!

PR #1897, opened 2026-04-22
Add LangGraph auto-instrumentation for Braintrust hook imports

### LINKED ISSUE #1889: Add automatic instrumentation for the LangGraph SDK
https://www.braintrust.dev/docs/integrations/agent-frameworks/langgraph

Right now this can only be manually instrumented, let's make this work with orchestrion

We need to also figure out what to do with `@braintrust/langchain-js`

Qard:
## Summary
- Add LangChain/LangGraph auto-instrumentation to `braintrust/hook.mjs` so callback handlers are injected automatically without requiring `@braintrust/langchain-js` imports.
- Keep the manual `@braintrust/langchain-js` API working while reusing the same span semantics through a vendored callback-handler surface.
- Add config support for `langchain`, `langchain-js`, and `langgraph` disable aliases and export the new LangChain instrumentation configs.
- Add a real provider-backed e2e scenario that verifies `--import braintrust/hook.mjs` captures LangGraph task spans and nested OpenAI spans end to end.

## Testing
- Passed focused Vitest coverage for the new LangChain plugin, config mapping, registry integration, and existing `@braintrust/langchain-js` tests.
- Passed the new provider-backed LangGraph e2e scenario repeatedly with `OPENAI_API_KEY` set.
- Passed the repo build, and the touched files were formatted successfully.

Fixes #1889


lforst (approved):
Let's also add e2e test ci summary for this.

lforst:
For the scenario we have in the test, it seems a bit odd that we are creating two LLM spans:

Screenshot 2026-05-06 at 16 38 27

js/src/instrumentation/registry.ts
```diff
@@ -58,6 +58,9 @@ export interface InstrumentationConfig {
     genkit?: boolean;
     gitHubCopilot?: boolean;
     openaiCodexSDK?: boolean;
+    langchain?: boolean;
+    langchainJS?: boolean;
```
lforst: langchainJS is maybe a bit much, I would remove that. I am also not sure about having both langchain and langgraph because disabling langchain may imply disabling both langchain and langgraph and disabling langgraph may imply disabling langgraph but not langchain but I am not totally sure or have a hard opinion.

js/src/instrumentation/plugins/langchain-plugin.ts
```diff
@@ -0,0 +1,128 @@
...
+      unsubscribe();
+    }
+    this.unsubscribers = [];
+    this.injectedManagers = new WeakSet<object>();
+  }
+
+  private subscribeToConfigure(channel: LangChainConfigureChannel): void {
+    const tracingChannel = channel.tracingChannel() as IsoTracingChannel<
+      ChannelMessage<LangChainConfigureChannel>
+    >;
```
lforst: I feel like we should if possible use the typed channel handlers we use in the other plugins. Makes """type safety""" a bit easier to manage.

js/src/instrumentation/plugins/langchain-channels.ts
```diff
@@ -0,0 +1,13 @@
+import { channel, defineChannels } from "../core/channel-definitions";
+import type { LangChainCallbackManagerConfigureResult } from "../../vendor-sdk-types/langchain";
+
+export const langChainChannels = defineChannels("@langchain/core", {
+  configure: channel<unknown[], LangChainCallbackManagerConfigureResult>({
+    channelName: "CallbackManager.configure",
+    kind: "sync-stream",
+  }),
+  configureSync: channel<unknown[], LangChainCallbackManagerConfigureResult>({
+    channelName: "CallbackManager._configureSync",
```
lforst: Going a bit hand in hand with my other comment, we should probably type the input arguments of this tracing channel.

js/src/wrappers/langchain/callback-handler.ts
```diff
@@ -0,0 +1,538 @@
...
+  LangChainEndSpanArgs,
+  LangChainLLMResult,
+  LangChainSerialized,
+  LangChainStartSpanArgs,
+} from "../../vendor-sdk-types/langchain";
+
+export const BRAINTRUST_LANGCHAIN_CALLBACK_HANDLER_NAME =
+  "BraintrustCallbackHandler";
+
+export class BraintrustLangChainCallbackHandler<
```
lforst, line 14: I feel like since `braintrust` is needed as a peer dependency, instead of duplicating this whole thing we could extract it and use a shared version?

js/src/instrumentation/plugins/langchain-callback-handler.ts
```diff
@@ -0,0 +1,538 @@
...
+
+    args.event = {
+      ...args.event,
+      tags: undefined,
+      metadata: {
+        ...(tags ? { tags } : {}),
+        ...args.event?.metadata,
+        braintrust: {
+          integration_name: "langchain-js",
+          integration_version: "0.2.0",
```
lforst: `integration_version` does not really apply here. Maybe we also remove it from the other package because it already is outdated.

js/src/auto-instrumentations/configs/langchain.test.ts
```diff
@@ -0,0 +1,36 @@
+import { describe, expect, it } from "vitest";
+import { langChainChannels } from "../../instrumentation/plugins/langchain-channels";
+import { langchainConfigs } from "./langchain";
+
+describe("langchain auto-instrumentation configs", () => {
```
lforst: I honestly don't like these tests. They are legit just asserting the exact shape of the export. Would remove.

e2e/scenarios/langgraph-auto-instrumentation/scenario.mjs
```diff
@@ -0,0 +1,61 @@
...
+import { ChatOpenAI } from "@langchain/openai";
+import { runMain, runTracedScenario } from "../../helpers/provider-runtime.mjs";
+import { ROOT_NAME, SCENARIO_NAME } from "./constants.mjs";
+
+const OPENAI_MODEL = "gpt-4o-mini-2024-07-18";
+
+runMain(async () => {
+  if (!process.env.OPENAI_API_KEY) {
+    throw new Error("OPENAI_API_KEY is required for this e2e scenario");
+  }
```
lforst: I think with cassettes this is not the case!

js/src/instrumentation/registry.ts
Qard: Dropped `langchainJS` in 33ec0f3f — kept `langchain` and `langgraph` as separate keys. `langgraph: false` disables only LangGraph; `langchain: false` disables the whole plugin.

js/src/instrumentation/plugins/langchain-channels.ts
Qard: Typed in 33ec0f3f — added `LangChainCallbackManagerConfigureArgs` matching `CallbackManager.configure`'s signature in `@langchain/core`. `channel<unknown[], ...>` is gone.

js/src/instrumentation/plugins/langchain-plugin.ts
Qard: Now using a typed tracing channel — dropped the `as IsoTracingChannel<...>` cast in favor of a typed declaration. Done in 33ec0f3f.

js/src/wrappers/langchain/callback-handler.ts
Qard: Started the extraction in 33ec0f3f: the auto-instrumentation copy is now at `js/src/wrappers/langchain/callback-handler.ts` (out of `plugins/`, into the shared `wrappers/` tree), and the plugin imports it from there. I didn't fully unify with `@braintrust/langchain-js`'s `BraintrustCallbackHandler` in this PR — that one extends `BaseCallbackHandler` from `@langchain/core` (which we don't want to import statically in the auto-instrumentation path), and the two copies have diverged in places, so a full unification is more invasive than it looks. Happy to do that as a follow-up.

js/src/instrumentation/plugins/langchain-callback-handler.ts
Qard: Removed from both copies in 33ec0f3f.

js/src/auto-instrumentations/configs/langchain.test.ts
Qard: Deleted in 33ec0f3f.

e2e/scenarios/langgraph-auto-instrumentation/scenario.mjs
Qard: Right — dropped the OPENAI_API_KEY guard in 33ec0f3f.

Qard:
On the two LLM spans concern — current span tree after 33ec0f3f shows only a single `ChatOpenAI [llm]` span. If you saw two on an earlier revision, the cassette / snapshot changes since may have already fixed it. Worth a fresh look at the new `span-tree.txt`.

integrations/langchain-js/src/BraintrustCallbackHandler.ts
```diff
@@ -1,186 +1,36 @@
-import {
-  BaseCallbackHandler,
-  BaseCallbackHandlerInput,
-} from "@langchain/core/callbacks/base";
+import { BaseCallbackHandler } from "@langchain/core/callbacks/base";
```
lforst, line 1: Reminder that we need a changeset for the langchain package now.

lforst (approved):

integrations/langchain-js/src/BraintrustCallbackHandler.ts
Qard: Oh, we need separate changeset files for each changed package?

lforst: No but we need to list both packages in the changeset. `pnpm run changeset` walks you through it.

lforst (approved):

PR #1898, opened 2026-04-23
feat: Add experiment dataset filter to experiment metadata

max-braintrust:
Stores _internal_btql filters for experiment datasets in the experiment metadata.

Right now we don’t persist those filter options, so we lose the ability to reconstruct the exact subset of rows an experiment ran against. If we save them with the experiment, we can recreate the same row set later instead of having to guess.

This unlocks a few useful things:
- Re-running an experiment on the exact same data it originally saw.
- Showing the BTQL filter used by an experiment in the Braintrust UI.
- Anything else that depends on reconstructing the precise rows that were initially fed into an experiment.


AbhiPrasad (approved):

lforst (approved):
It would be great if you could still add to the PR description how a user would use this or how this is used in the product!

PR #1901, opened 2026-04-23
Add mastra auto-instrumentation

### LINKED ISSUE #1890: Add automatic instrumentation for Mastra
https://www.braintrust.dev/docs/integrations/agent-frameworks/mastra

add the ability to instrument mastra w/ orchestrion

Qard:
Fixes #1890

## Summary
- add Mastra auto-instrumentation configs and wire them into the hook, bundler plugin, webpack loader, and public exports
- add Mastra tracing channels and a runtime plugin for agent, tool, workflow run, and workflow step spans with current-span context binding
- add focused Mastra tests and an e2e scenario covering agent generate/stream, Tool.execute, workflow runs, and workflow-step tool nesting

## Testing
- `pnpm --dir js exec vitest run src/auto-instrumentations/configs/mastra.test.ts src/instrumentation/plugins/mastra-plugin.test.ts src/instrumentation/registry.test.ts`
- `pnpm run test:e2e -- mastra-instrumentation`

## Notes
- `pnpm --dir js test -- mastra` and `pnpm --dir js test -- auto-instrumentations` still hit unrelated existing network-dependent failures in `src/framework.test.ts` and `src/logger-misc.test.ts`
- `pnpm run formatting` still reports pre-existing unformatted fixture files under `js/tests/auto-instrumentations/fixtures/test-files/`

### DIFF
--- a/e2e/scenarios/mastra-instrumentation/__snapshots__/mastra-v1260.span-tree.txt
+++ b/e2e/scenarios/mastra-instrumentation/__snapshots__/mastra-v1260.span-tree.txt
@@ -1,0 +1,199 @@
+span_tree:
+└── mastra-instrumentation-root [task]
+    metadata: {
+      "scenario": "mastra-instrumentation"
+    }
+    ├── mastra-agent-generate-operation
+    │   metadata: {
+    │     "operation": "generate"
+    │   }
+    │   ├── agent run: 'weather-agent' [task]
+    │   │   input: "What is the weather in Paris?"
+    │   │   output: {
+    │   │     "files": [],
+    │   │     "text": "The forecast is sunny."
+    │   │   }
+    │   │   metadata: {
+    │   │     "entity_id": "weather-agent",
+    │   │     "entity_name": "Weather Agent",
+    │   │     "entity_type": "agent"
+    │   │   }
+    │   │   └── llm: 'mock-model-id' [llm]
+    │   │       input: {
+    │   │         "messages": [
+    │   │           {
+    │   │             "content": "Answer weather questions with the provided mock forecast.",
+    │   │             "role": "system"
+    │   │           },
+    │   │           {
+    │   │             "content": [
+    │   │               {
+    │   │                 "providerOptions": {
+    │   │                   "mastra": {
+    │   │                     "createdAt": 0
+    │   │                   }
+    │   │                 },
+    │   │                 "text": "What is the weather in Paris?",
+    │   │                 "type": "text"
+    │   │               }
+    │   │             ],
+    │   │             "role": "user"
+    │   │           }
+    │   │         ]
+    │   │       }
+    │   │       output: {
+    │   │         "files": [],
+    │   │         "reasoning": [],
+    │   │         "sources": [],
+    │   │         "text": "The forecast is sunny.",
+    │   │         "warnings": []
+    │   │       }
+    │   │       metadata: {
+    │   │         "entity_id": "weather-agent",
+    │   │         "entity_name": "Weather Agent",
+    │   │         "entity_type": "agent",
+    │   │         "model": "mock-model-id",
+    │   │         "provider": "mock-provider"
+    │   │       }
+    │   │       metric_keys: [
+    │   │         "completion_tokens",
+    │   │         "prompt_tokens",
+    │   │         "tokens"
+    │   │       ]
+    │   │       └── step: 0 [llm]
+    │   │           output: {
+    │   │             "text": "The forecast is sunny.",
+    │   │             "toolCalls": []
+    │   │           }
+    │   │           metadata: {
+    │   │             "entity_id": "weather-agent",
+    │   │             "entity_name": "Weather Agent",
+    │   │             "entity_type": "agent"
+    │   │           }
+    │   │           metric_keys: [
+    │   │             "completion_tokens",
+    │   │             "prompt_tokens",
+    │   │             "tokens"
+    │   │           ]
+    │   │           └── chunk: 'text' [llm]
+    │   │               output: {
+    │   │                 "text": "The forecast is sunny."
+    │   │               }
+    │   │               metadata: {
+    │   │                 "entity_id": "weather-agent",
+    │   │                 "entity_name": "Weather Agent",
+    │   │                 "entity_type": "agent"
+    │   │               }
+    │   ├── agent run: 'weather-agent' [task]
+    │   │   input: "Stream the Paris forecast."
+    │   │   output: {
+    │   │     "files": [],
+    │   │     "text": "The forecast is sunny."
+    │   │   }
+    │   │   metadata: {
+    │   │     "entity_id": "weather-agent",
+    │   │     "entity_name": "Weather Agent",
+    │   │     "entity_type": "agent"
+    │   │   }
+    │   │   └── llm: 'mock-model-id' [llm]
+    │   │       input: {
+    │   │         "messages": [
+    │   │           {
+    │   │             "content": "Answer weather questions with the provided mock forecast.",
+    │   │             "role": "system"
+    │   │           },
+    │   │           {
+    │   │             "content": [
+    │   │               {
+    │   │                 "providerOptions": {
+    │   │                   "mastra": {
+    │   │                     "createdAt": 0
+    │   │                   }
+    │   │                 },
+    │   │                 "text": "Stream the Paris forecast.",
+    │   │                 "type": "text"
+    │   │               }
+    │   │             ],
+    │   │             "role": "user"
+    │   │           }
+    │   │         ]
+    │   │       }
+    │   │       output: {
+    │   │         "files": [],
+    │   │         "reasoning": [],
+    │   │         "sources": [],
+    │   │         "text": "The forecast is sunny.",
+    │   │         "warnings": []
+    │   │       }
+    │   │       metadata: {
+    │   │         "entity_id": "weather-agent",
+    │   │         "entity_name": "Weather Agent",
+    │   │         "entity_type": "agent",
+    │   │         "model": "mock-model-id",
+    │   │         "provider": "mock-provider"
+    │   │       }
+    │   │       metric_keys: [
+    │   │         "completion_tokens",
+    │   │         "prompt_tokens",
+    │   │         "tokens"
+    │   │       ]
+    │   │       └── step: 0 [llm]
+    │   │           output: {
+    │   │             "text": "The forecast is sunny.",
+    │   │             "toolCalls": []
+    │   │           }
+    │   │           metadata: {
+    │   │             "entity_id": "weather-agent",
+    │   │             "entity_name": "Weather Agent",
+    │   │             "entity_type": "agent"
+    │   │           }
+    │   │           metric_keys: [
+    │   │             "completion_tokens",
+    │   │             "prompt_tokens",
+    │   │             "tokens"
+    │   │           ]
+    │   │           └── chunk: 'text' [llm]
+    │   │               output: {
+    │   │                 "text": "The forecast is sunny."
+    │   │               }
+    │   │               metadata: {
+    │   │                 "entity_id": "weather-agent",
+    │   │                 "entity_name": "Weather Agent",
+    │   │                 "entity_type": "agent"
+    │   │               }
+    │   └── workflow run: 'travel-flow' [task]
+    │       input: {
+    │         "city": "Berlin"
+    │       }
+    │       output: {
+    │         "forecast": "Sunny in Berlin"
+    │       }
+    │       metadata: {
+    │         "entity_id": "travel-flow",
+    │         "entity_name": "travel-flow",
+    │         "entity_type": "workflow_run"
+    │       }
+    │       └── workflow step: 'lookup-step' [function]
+    │           input: {
+    │             "city": "Berlin"
+    │           }
+    │           output: {
+    │             "forecast": "Sunny in Berlin"
+    │           }
+    │           metadata: {
+    │             "entity_id": "lookup-step",
+    │             "entity_name": "travel-flow",
+    │             "entity_type": "workflow_step"
+    │           }
+    ├── mastra-agent-stream-operation
+    │   metadata: {
+    │     "operation": "stream"
+    │   }
+    ├── mastra-tool-operation
+    │   metadata: {
+    │     "operation": "tool"
+    │   }
+    └── mastra-workflow-operation
+        metadata: {
+          "operation": "workflow"
+        }
--- a/js/src/auto-instrumentations/configs/all.ts
+++ b/js/src/auto-instrumentations/configs/all.ts
@@ -77,6 +77,13 @@ const defaultInstrumentationConfigGroups: readonly InstrumentationConfigGroup[] 
       integrations: ["gitHubCopilot"],
       configs: gitHubCopilotConfigs,
     },
+    // Note: `@mastra/core` is not listed here because its instrumentation
+    // doesn't go through the AST `code-transformer` matcher — Mastra's
+    // content-hashed chunks make `filePath`-based matching too brittle.
+    // Instead it's handled by the source-replacement entry in
+    // `loader/special-case-patches.ts`, which both the runtime loader
+    // (`hook.mjs` → `cjs-patch.ts`/`esm-hook.mts`) and the bundler plugin
+    // (`bundler/plugin.ts`) call. The `mastra` env-var disable still works.
   ];
 
 export function getDefaultInstrumentationConfigs({
--- a/js/src/auto-instrumentations/configs/mastra.ts
+++ b/js/src/auto-instrumentations/configs/mastra.ts
@@ -1,0 +1,152 @@
+import type { InstrumentationConfig } from "@apm-js-collab/code-transformer";
+import { mastraChannels } from "../../instrumentation/plugins/mastra-channels";
+import { mastraAssignedAsyncTransformName } from "../custom-transforms";
+
+const mastraPackageName = "@mastra/core";
+
+const stableVersionRange = "1.26.0";
+const alphaVersionRange = "1.26.1-alpha.0";
+
+const stableAgentWorkflowChunks = [
+  { filePath: "dist/chunk-CXW3Z2OL.js", moduleType: "esm" },
+  { filePath: "dist/chunk-PBGNXXU5.cjs", moduleType: "cjs" },
+] as const;
+
+const alphaAgentWorkflowChunks = [
+  { filePath: "dist/chunk-4QTY73BW.js", moduleType: "esm" },
+  { filePath: "dist/chunk-7432OHPH.cjs", moduleType: "cjs" },
+] as const;
+
+const toolChunks = [
+  { filePath: "dist/chunk-O3JJ5ZPY.js", moduleType: "esm" },
+  { filePath: "dist/chunk-U7Z7GCXY.cjs", moduleType: "cjs" },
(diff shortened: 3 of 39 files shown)


lforst (approved):
Can we also add the mastra e2e test scenario to the ci summary?

lforst (commented):
In the [e2e test summary](https://www.braintrust.dev/app/Braintrust%20SDKs/p/js-sdk-e2e-tests/logs?tvt=trace&search={%22filter%22:[{%22label%22:%22metadata.scenario%2520%253D%2520%2522mastra-instrumentation%2522%22,%22originType%22:%22btql%22,%22text%22:%22metadata.scenario%2520%253D%2520%2522mastra-instrumentation%2522%22},{%22label%22:%22metadata.testRunId%2520%253D%2520%2522e2e-f5a85bf4-5060-4f8e-a7f2-5356f851e2d6%2522%22,%22originType%22:%22btql%22,%22text%22:%22metadata.testRunId%2520%253D%2520%2522e2e-f5a85bf4-5060-4f8e-a7f2-5356f851e2d6%2522%22}]}&r=71f822bd-8db1-4f33-9a41-fbb5d6b61fb6&s=4fbf212d-64f4-419f-a612-92e9b545cacc) it looks like we don't capture LLM spans.

Image

 Do you think that is an issue?

js/src/instrumentation/registry.ts
```diff
@@ -189,26 +201,32 @@ class PluginRegistry {
...
+      return "gitHubCopilot";
+    case "openai-codex-sdk":
+      return "openaiCodexSDK";
+    case "@mastra/core":
+    case "mastra-core":
+      return "mastra";
+    default:
+      return name as IntegrationName;
+  }
+}
```
lforst: Did you find this necessary? It randomly widens our API scope and what we need to document.

js/src/instrumentation/plugins/mastra-channels.ts
```diff
@@ -0,0 +1,65 @@
...
+  MastraWorkflowStepParams,
+} from "../../vendor-sdk-types/mastra";
+
+export const mastraChannels = defineChannels("@mastra/core", {
+  agentExecute: channel<[MastraAgentExecuteOptions], unknown>({
+    channelName: "agent.execute",
+    kind: "async",
+  }),
+
+  agentNetwork: channel<[unknown, MastraAgentNetworkOptions?], unknown>({
```
lforst: I think we can improve the typing of these channels. I see that we are casting them in the implementation so gut feeling wise this should be possible.

e2e/scenarios/mastra-instrumentation/__snapshots__/mastra-v1260.span-tree.txt
```diff
@@ -0,0 +1,286 @@
+span_tree:
```
lforst, line 1: I don't see any type: llm spans here which is a bit worrying.

js/src/auto-instrumentations/custom-transforms.ts
```diff
@@ -0,0 +1,80 @@
+import type {
```
lforst: This entire thing is like so sketchy. I assume you did everything in your power trying different approaches?

js/src/auto-instrumentations/custom-transforms.ts
```diff
@@ -0,0 +1,80 @@
...
+      tracePromise?: CustomTransform;
+    }
+  ).tracePromise ??
+  (
+    codeTransformerTransforms as {
+      default?: {
+        tracePromise?: CustomTransform;
+      };
+    }
+  ).default?.tracePromise;
```
lforst: Can we add a comment explaining why this is necessary?

js/src/auto-instrumentations/custom-transforms.ts
```diff
@@ -0,0 +1,80 @@
...
+
+if (!tracePromiseTransform) {
+  throw new Error(
+    "Failed to resolve code-transformer tracePromise helper for Mastra instrumentation.",
```
lforst: I feel like this error message should not be mastra specific?

js/src/auto-instrumentations/configs/mastra.ts
```diff
@@ -0,0 +1,152 @@
...
+const alphaAgentWorkflowChunks = [
+  { filePath: "dist/chunk-4QTY73BW.js", moduleType: "esm" },
+  { filePath: "dist/chunk-7432OHPH.cjs", moduleType: "cjs" },
+] as const;
+
+const toolChunks = [
+  { filePath: "dist/chunk-O3JJ5ZPY.js", moduleType: "esm" },
+  { filePath: "dist/chunk-U7Z7GCXY.cjs", moduleType: "cjs" },
+] as const;
+
```
lforst: Are we really only instrumenting one specific version of the package?

I also have the gut feeling that the chunk hashes have a pretty high chance of updating between versions putting burden on us to update everything, and also our users being blocked from upgrading if they want to have tracing. While implementing, did you explore any other options of doing things? Maybe patching the `new Mastra()` constructor to pass in an observability object of some sort? https://mastra.ai/docs/observability/tracing/exporters/braintrust#zero-config-setup

js/src/instrumentation/registry.ts
Qard: Dropped the `@mastra/core` and `mastra-core` aliases in e9e32249 — `mastra` is the canonical key. The other aliases in `normalizeIntegrationName` (cursor-sdk → cursorSDK, github-copilot → gitHubCopilot, etc.) bridge case mismatches between env var conventions and the property names; `mastra` already matches both, so the extras were just noise. `normalizeIntegrationName` and `IntegrationName` are both module-local, not exported, so the public API surface is unchanged either way.

js/src/instrumentation/plugins/mastra-channels.ts
Qard: Tightened in e9e32249 — `agent.network` / `agent.resumeNetwork` now take `MastraAgentLike` as the first arg, and the legacy `unknown` slots in tool/step channels now have meaningful type aliases. The double `as unknown as IsoTracingChannel<...>` in `mastra-plugin.ts` is gone (single annotated cast). I left a few `unknown` for user-defined payloads (tool input, step input data) since those are arbitrary by definition — open to a more specific shape if you have one in mind.

e2e/scenarios/mastra-instrumentation/__snapshots__/mastra-v1260.span-tree.txt
Qard: Fixed in e9e32249 — `Mastra Workflow Step workflow.executionWorkflow.step.llm-execution` is now typed as `[llm]` (it's the synthetic step where the model call actually happens). The new `mastra-v1260.span-tree.txt` shows two `[llm]` spans (one per agent.generate/stream invocation).

js/src/auto-instrumentations/custom-transforms.ts
Qard: Yeah it's sketchy. Added explanatory comments at the top of `custom-transforms.ts` and inline at the @ts-expect-error in e9e32249. Short version: code-transformer doesn't have an ESM-friendly public export for `tracePromise` yet, so we reach into `lib/transforms.js` and tolerate the dual ESM/CJS shape. Alternatives I explored: (1) patching Mastra constructors at runtime, but the chunk-hash file naming makes orchestrion matching brittle; (2) re-implementing tracePromise locally — duplicates code-transformer's nontrivial AST transforms; (3) waiting for code-transformer to export `tracePromise` publicly — that's the right long-term answer. Happy to file an upstream issue if you want this resolved at the source.

js/src/auto-instrumentations/custom-transforms.ts
Qard: Added a header doc-comment in e9e32249 explaining the arrow-in-constructor problem the transform is solving.

js/src/auto-instrumentations/custom-transforms.ts
Qard: Generalized in e9e32249 — now reads `Failed to resolve the tracePromise helper from @apm-js-collab/code-transformer.`

js/src/auto-instrumentations/configs/mastra.ts
Qard: You're right that the chunk-hash approach is the wrong long-term answer — every Mastra release will produce new chunk filenames, so we'll be playing catch-up. I restructured the config in e9e32249 into a `mastraVersions` table that makes adding a new release mechanical (install, grep `dist/` for the agent/workflow/tool chunks, append an entry).

For the doc'd `new Mastra({ observability: ... })` exporter pattern: yes, that's the right design and I want to switch to it, but it's a meaningful refactor — we need an orchestrion match for the `Mastra` constructor (or a runtime-side patch in `hook.mjs` that wraps the imported class), plus a Braintrust exporter that conforms to Mastra's exporter interface. I'd rather land this PR with the per-version mapping documented and ship constructor patching as a focused follow-up than block on the redesign.

Versions outside the current `mastraVersions` entries simply no-op (no spans) — the user's code still runs fine — so the graceful-degradation cost is acceptable in the interim.

e2e/scenarios/mastra-instrumentation/__snapshots__/mastra-v1260.span-tree.txt
lforst: Can we loosen the normalization a bit on the scenario so we can see input & output of the llm spans?

js/src/auto-instrumentations/configs/mastra.ts
lforst: I think going with the current approach and merging this PR doesn't really make sense in its current state. This PR is about adding mastra auto instrumentation and if it only works for a single version (which isn't even the latest or most downloaded version) we are missing the goal no?

I think we should immediately try to find a good solution here.

Qard:
@lforst Getting some provider limit failures, but otherwise should be good for another review.

js/src/wrappers/mastra.ts
```diff
@@ -1,5 +1,383 @@
...
+
+function logExporterError(err: unknown): void {
+  debugLogger.warn("Mastra exporter failure:", err);
+}
+
+/**
+ * @deprecated `wrapMastraAgent` is a no-op kept for one minor cycle so older
+ * tutorials don't break the build. Use `BraintrustObservabilityExporter` via
+ * `new Mastra({ observability: ... })` or rely on
+ * `node --import braintrust/hook.mjs` for auto-instrumentation.
```
lforst: This deprecation message could be phrased more user facing.

js/src/wrappers/mastra.ts
```diff
@@ -1,5 +1,383 @@
...
+
+/** Global name the loader patch reads at runtime. Exported so the loader
+ *  patch and `hook.mjs` stay in sync — `hook.mjs` registers the factory on
+ *  `globalThis`, the Observability wrap consumes it. */
+export const MASTRA_EXPORTER_FACTORY_GLOBAL =
+  "__braintrustMastraExporterFactory";
+
+function logExporterError(err: unknown): void {
+  debugLogger.warn("Mastra exporter failure:", err);
+}
```
lforst, line 362: Bit of an overkill util.

js/src/instrumentation/registry.test.ts
```diff
@@ -156,6 +156,12 @@ describe("configureInstrumentation API", () => {
         cohere: false,
       },
     });
+
+    configureInstrumentation({
+      integrations: {
+        mastra: false,
+      },
+    } as any);
```
lforst: Why is this separate?

js/src/instrumentation/config.ts
```diff
@@ -52,6 +53,9 @@ const envIntegrationAliases: Record<
   cursorsdk: "cursorSDK",
   flue: "flue",
   "flue-runtime": "flue",
+  mastra: "mastra",
+  "@mastra/core": "mastra",
+  "mastra-core": "mastra",
```
lforst: Let's just do `mastra`. Everything else is a bit unconventional.

js/src/instrumentation/braintrust-plugin.ts
```diff
@@ -23,6 +23,8 @@ export interface BraintrustPluginConfig {
   integrations?: InstrumentationIntegrationsConfig;
 }
 
+type RuntimeBraintrustPluginConfig = BraintrustPluginConfig;
```
lforst: Why this alias?

js/src/instrumentation/braintrust-plugin.test.ts
```diff
@@ -529,7 +529,7 @@ describe("BraintrustPlugin", () => {
           groq: false,
           gitHubCopilot: false,
         },
-      });
+      } as any);
```
lforst: Sketchy/unnecessary change? Also above.

js/src/wrappers/mastra.ts
```diff
@@ -1,5 +1,383 @@
...
+/** Factory used by the loader patch. Decoupled from the class so the runtime
+ *  patch can stay a small string of JS that just calls a global function. */
+export function createBraintrustObservabilityExporter(): MastraObservabilityExporter {
+  return new BraintrustObservabilityExporter();
+}
+
+/** Global name the loader patch reads at runtime. Exported so the loader
+ *  patch and `hook.mjs` stay in sync — `hook.mjs` registers the factory on
+ *  `globalThis`, the Observability wrap consumes it. */
+export const MASTRA_EXPORTER_FACTORY_GLOBAL =
```
lforst: I find this an odd place to put this export when it is only used in auto instrumentation code.

js/src/auto-instrumentations/loader/cjs-patch.ts
```diff
@@ -48,13 +52,27 @@ export class ModulePatch {
...
+        // orchestrion matcher because Mastra's chunk filenames change every
+        // release, but the package-exports entry paths are stable.
+        const mastraTarget = classifyMastraTarget(
+          packageName,
+          normalizedModulePath,
+        );
+        if (mastraTarget) {
+          args[0] = patchMastraSource(String(content), mastraTarget, "cjs");
+          return self.originalCompile.apply(this, args);
+        }
```
lforst, line 66: Can we make it somehow more obvious that this is not "normal" code in this file? I also really don't like that we have openai specific code in in this function without any note that it is ideally temporary. I would recommend we put this into a separate function with a note that this is hopefully temporary somehow. Also so that agents see that this is actually an anti-pattern and are not tempted to add similar code as a shortcut.

js/src/auto-instrumentations/loader/mastra-observability-patch.ts
```diff
@@ -0,0 +1,268 @@
+/**
```
lforst, line 1: **This is the most important feedback.**

This goes more into the direction I find good. However, I find this approach still quite sketchy. I would find this acceptable if we were in a hurry to get this out asap, which is not the case.

I honestly think we should invest the time to use `require-in-the-middle` and `import-in-the-middle` whenever we are targeting top-level-exports (ie. stable API) which seems to be the case here. That way we don't need sketchy custom one-off instrumentation code for all packages.

Qard: IITM would be substantially worse. It's very unreliable and has been known-broken with half the major AI SDKs for a while now because there are a bunch of design issues which are simply not addressable via basic ESM loaders wrapping. I went with orchestrion in the first place because most of the modules we instrument would crash IITM. As for RITM, it's not particularly useful at this point either given many of the AI SDKs are ESM-only now.

There's talk of rewriting the two from scratch on the newer sync loader hooks API, but no one has really been putting any work into that. Unless we want to take on rewriting RITM + IITM ourselves, I don't think there's a viable path there yet.

lforst: > IITM would be substantially worse.

Can why it would be substantially worse for this particular case?

> It's very unreliable

From experience, IITM only has problems when there are actual problems. I have not seen it be unreliable. Can you elaborate what you mean by it being unreliable?

Its only downside that I know of you have already mentioned, which is that it does not work for very specific packages if they export things in a certain dodgy way.

Right now looking at the PR, probabilistically extracting a chunk seems more unreliable (ie. less future-proof) than using IITM to patch public API (which tends to be stable). To me, unless IITM/RITM are broken for Mastra, it seems like the best tool for the job.

Qard: When I initially built the auto-instrumentation a few months ago IITM was the first thing I tried because, despite knowing it had some problems, it overall had more adoption just from being around longer. But I quickly found that, since we were starting from scratch, building for the future seemed a better approach.

With IITM, both OpenAI and AI SDK crashed the process on load because of syntax support issues. It's also known to have issues with patching only applying externally and so code which calls other functions from the same file often don't work. Also IITM wraps literally every module, even ones not being patched, in proxies and injects a bunch of setter registration stuff which adds a ton of overhead to every access. That's not only a big performance hit but also a significant stability risk given its poor track record with module parsing, which it had to do whether or not it was actually patching the module or not, since it would not know the patches to apply until _after_ the load due to the ordering of things.

IITM itself is _much_ more hacky than what we're doing here. Yes, there's a bit of jank in what we're doing here, but we can smooth that over as our AST-based patching matures. Our options are: working but with a bit of our own jank to smooth out, and aligning with the direction of the segment leaders, or often broken and concealing a ton of jank within which provides a significant unaddressable risk. As one of the people responsible for creating IITM in the first place, I don't think the tradeoffs are worth it.

The core architecture of IITM trying to treat modules as load-time-patchable like RITM simply does not fit the reality that ESM just doesn't work the way CJS does, and so it accumulated a ton of hacks to try and work around the numerous sharp edges to the point that it's now even more of a mess than when it started. We built it at Datadog and then abandoned it not too long after handing it off to the Node.js foundation because it was just more trouble than it was worth. Datadog created Orchestrion shortly after specifically _because_ of this failure, and now there's several vendors migrating to that instead. Using IITM at this point feels to me like a step backwards.

lforst: > Also IITM wraps literally every module, even ones not being patched

I believe this no longer has to be the case since `createAddHookMessageChannel` was added: https://github.com/nodejs/import-in-the-middle#only-intercepting-hooked-modules I have had a nice experience using it in the past.

I also don't have anything against using orchestrion. It covers the cases that IITM/RITM can't. However, I don't understand you mentioning it because we are not using it in this PR.

In any case, I feel like I won't convince you on this so feel free to proceed.

js/src/auto-instrumentations/configs/all.ts
```diff
@@ -77,6 +77,10 @@ const defaultInstrumentationConfigGroups: readonly InstrumentationConfigGroup[]
       integrations: ["gitHubCopilot"],
       configs: gitHubCopilotConfigs,
     },
+    // Note: `@mastra/core` is handled by the loader's mastra-observability
+    // patch (see `loader/mastra-observability-patch.ts`), not via AST
+    // transforms — Mastra's content-hashed chunks make `filePath`-based
+    // matching too brittle.
```
lforst: This might be an issue, because it means the bundler plugins will not instrument the package. Either you can address this or please don't forget to sufficiently document it.

lforst (approved):

js/src/auto-instrumentations/configs/all.ts
Qard: I've addressed it.

PR #1920, opened 2026-04-29
feat(e2e): MSW cassette layer for hermetic e2e tests

Qard:
## Summary

Adds an inbound provider HTTP capture/replay layer (cassettes) to the e2e test suite so hermetic CI runs replay recorded traffic instead of hitting live provider APIs. Built on **MSW** (already in the workspace as a dev dep for `integrations/langchain-js` and `integrations/otel-js`) plus ~600 LoC of cassette-format/matcher/recorder glue.

E2E scenarios previously hit live provider APIs on every CI run. Flakiness sources: rate limits, transient 5xx, model-output drift breaking exact-string snapshot fields. With this layer:
- Replay mode (CI default) hits no network. Subprocess MSW interceptor returns canned responses from `__cassettes__/<variantKey>.json`.
- Recording is local-only via `BRAINTRUST_E2E_CASSETTE_MODE=record-missing`. CI never records.
- The existing outbound `mock-braintrust-server.ts` and `__snapshots__/` are untouched. Cassettes are the inbound mirror: provider→SDK, where snapshots are SDK→Braintrust.

**Status:** ~506 tests passing in hermetic mode across 25 scenarios, 2 of 3 consecutive runs deterministic (the third had an unrelated `turbopack-auto-instrumentation` Next.js compile timeout flake — not introduced by this PR).

## Architecture

| Layer | Direction | Captured by | Stored in |
|---|---|---|---|
| Existing | SDK → Braintrust | `mock-braintrust-server.ts` (parent process HTTP server) | `__snapshots__/*.json` |
| **New** | Provider → SDK | `cassette/preload.mjs` boots an MSW `setupServer()` in subprocess | `__cassettes__/<variantKey>.json` |

- `e2e/helpers/cassette/preload.mjs` — loaded into each scenario subprocess via `node --import=<preload>`. Boots MSW synchronously and intercepts provider HTTP traffic.
- Per-key call counter for retries; reuses single entry on every match if only one entry exists for a key.
- Recorder server runs in the parent vitest process; subprocess POSTs captured wire to it; parent merges entries across variant runs and writes once at scenario end.
- `cassetteTagsFor(import.meta.url, variantKey)` auto-tags scenarios with `hermetic` based on cassette file presence — opt-in is by committing the cassette.

## Cassette modes (`BRAINTRUST_E2E_CASSETTE_MODE`)

- `replay` (default in CI): match or throw `CassetteMissError`.
- `record`: overwrite cassette fresh.
- `record-missing`: match if possible, else live + record. Standard re-record loop.
- `passthrough`: bypass cassettes entirely (local debugging).

## Recording safeguards

- **Skip-list filters transient/auth failures** so they don't poison the cassette: 400/401/403/429/5xx are not persisted. Caught one real bug here — google-genai cassettes had been recorded with a bad key, capturing `400 API_KEY_INVALID` responses. The 400 case was added after.
- **Automatic retry-with-backoff for 429/5xx** during recording (2 attempts, 5s/10s, abort-aware to respect SDK timeouts).
- **Volatile-header scrub** strips `authorization`, `x-api-key`, `api-key`, `x-goog-api-key`, `cohere-api-key`, cookies, request IDs, rate-limit windows, `content-encoding`, etc. before persisting. (Caught a near-miss on this PR — the initial commit leaked `x-api-key` for Anthropic; volatile-header set has been broadened and a scrub run removed leaked values.)
- **Headers forwarding fix** — `new Headers(request.headers)` silently drops most headers when the source is an MSW-intercepted request (Authorization included). The forwarder copies via `forEach` instead. This one bug was responsible for the bulk of the recording failures during initial migration (every Mistral 401, plenty of others).

## Scenarios with complete cassettes (hermetic green)

- `anthropic-instrumentation` (6 variants)
- `openai-instrumentation` (3 variants)
- `claude-agent-sdk-instrumentation`
- `openrouter-instrumentation` (2 variants)
- `ai-sdk-instrumentation` (4 variants)
- `ai-sdk-otel-export` (2 variants)
- `groq-instrumentation` (2 variants)
- `huggingface-instrumentation` (3 variants)
- `openrouter-agent-instrumentation`
- `wrap-langchain-js-traces`
- `cassette-replay` (meta-scenario validating record→replay loop end-to-end)
- `cohere-instrumentation` v7-14-0 (1 of 5 variants — see below)

## Scenarios still missing cassettes (auto-skipped in hermetic mode)

These auto-skip cleanly because `cassetteTagsFor` only applies the `hermetic` tag when the cassette file is present. CI does not fail on them today; they need a follow-up record run with working credentials.

### `mistral-instrumentation` — needs re-record after rebase
- This branch was originally based on PR #1891 (OpenAI Agents auto-instrumentation). After being rebased onto `main`, the existing mistral cassettes no longer match: main extended the mistral scenario with new thinking/reasoning model coverage (`NATIVE_REASONING_MODEL`, `ADJUSTABLE_REASONING_MODEL`) that wasn't in the older scenario shape the cassettes were recorded against.
- Re-recording also hit aggressive provider rate limits on the new reasoning models — the existing cassette layer's retry-with-backoff (5s/10s, 2 attempts) is not sufficient for that endpoint, so re-recording stalled mid-suite. Solving this likely requires either a longer per-call throttle (mirroring what `cohere-instrumentation` already does — `COHERE_RECORD_THROTTLE_MS = 60_000`) or running mistral variants serially with longer waits between calls.
- Cassettes for the older scenario shape were dropped to keep the suite green; mistral now auto-skips in hermetic mode until the follow-up re-record lands.
- **Action needed:** add a record-time throttle to mistral's `scenario.impl.mjs` and run `BRAINTRUST_E2E_CASSETTE_MODE=record-missing pnpm --filter=@braintrust/js-e2e-tests vitest run scenarios/mistral-instrumentation`.

### `google-genai-instrumentation` — Gemini quota exhausted
- Quota on the recording key is fully consumed (`RESOURCE_EXHAUSTED 429`).
- Initial cassettes had been recorded back when the key was invalid and captured `400 API_KEY_INVALID` responses; those were detected and deleted in this PR. The skip-list now rejects 400 to prevent recurrence.
- **Action needed:** rotate to a working Gemini key (or wait for quota reset) and run `BRAINTRUST_E2E_CASSETTE_MODE=record-missing pnpm --filter=@braintrust/js-e2e-tests vitest run scenarios/google-genai-instrumentation`.

### `cohere-instrumentation` — per-MONTH quota exhausted (4 of 5 variants)
- Cohere account is past the per-MONTH request limit on the chat models. Quote from API: `"You are past the per-month request limit for this model, please wait and try again later."` This is monthly, not daily — recovers on the next billing cycle.
- The `v7-14-0` variant is fully recorded (chat + chat-stream + embed + rerank) and replays green. The 4 remaining variants (`v7-20-0`, `v7-21-0`, `v7` default, `v8`) auto-skip until re-recorded.
- The scenario impl now includes a 60s throttle between calls during recording (`COHERE_RECORD_THROTTLE_MS`) to land each call in a fresh budget window once quota is restored — but the throttle can't help with monthly exhaustion.
- **Action needed:** wait for the next billing cycle (or upgrade Cohere plan), then run the same record command for the cohere scenarios.

### `google-adk-instrumentation` — model-behavior drift, unrelated to cassette layer
- ADK does not have any `__cassettes__/` files in this PR, so it auto-skips in hermetic mode. There is pre-existing snapshot drift unrelated to this PR which should be triaged independently.

## Risks / things to watch

- **Cassette files are large and committed.** `claude-agent-sdk-instrumentation` and `ai-sdk-instrumentation` have the largest cassettes (long transcripts). This is intentional — diff-ability matters for review, and we want byte-identical replay.
- **Re-record workflow is documented** in `e2e/README.md` ("Cassettes" section) and `.agents/skills/e2e-tests/SKILL.md`.
- **Day-one scope excluded** Deno scenarios, `nextjs-instrumentation`, `turbopack-auto-instrumentation`, and OTEL-only scenarios. Those need separate preload mechanisms; they continue running as before (or are already hermetic-ish via different machinery).

## Test plan

- [x] `pnpm --filter=@braintrust/js-e2e-tests exec vitest run --tags-filter=hermetic` — green (506 passed, 396 skipped, 0 failed)
- [x] Same command run consecutively for determinism — green twice in a row (one intermediate run hit an unrelated turbopack flake)
- [x] `pnpm run formatting` clean
- [x] `pnpm run lint` — 0 errors
- [x] No leaked credentials in committed cassettes (`x-api-key`, `api-key`, `x-goog-api-key`, `authorization`, etc. all stripped)
- [ ] CI hermetic suite green
- [ ] Re-record blocked scenarios (mistral, google-genai, remaining cohere variants) once credentials/quota are available — follow-up PR


e2e/vitest.setup.ts
```diff
@@ -1,3 +1,19 @@
...
+// shell. Existing env values are preserved (override: false).
+const setupDir = path.dirname(fileURLToPath(import.meta.url));
+const repoRoot = path.resolve(setupDir, "..");
+loadDotenv({ path: path.join(repoRoot, ".env"), override: false, quiet: true });
+loadDotenv({
+  path: path.join(repoRoot, ".env.local"),
+  override: false,
+  quiet: true,
+});
+
```
lforst, line 18: mise should be taking care of loading .env

e2e/scripts/record-cassettes.mjs
```diff
@@ -0,0 +1,39 @@
...
+const SCRIPT_DIR = path.dirname(fileURLToPath(import.meta.url));
+const E2E_ROOT = path.resolve(SCRIPT_DIR, "..");
+
+const args = process.argv.slice(2);
+
+const PNPM = process.platform === "win32" ? "pnpm.cmd" : "pnpm";
+
+const env = {
+  ...process.env,
+  BRAINTRUST_E2E_CASSETTE_MODE: "record",
```
lforst, line 24: I feel like instead of having this entire file we can just set the env var in the `test:e2e:record` script?

.agents/skills/e2e-tests/SKILL.md
```diff
@@ -33,10 +33,27 @@ Run workspace scripts from the repo root when you want the standard e2e entrypoi
...
+- To re-record after changing a scenario:
+
+  ```bash
+  ANTHROPIC_API_KEY=... OPENAI_API_KEY=... \
+    pnpm --filter=@braintrust/js-e2e-tests run test:e2e:record scenarios/<name>/scenario.test.ts
+  ```
+
+  Then run again in `BRAINTRUST_E2E_CASSETTE_MODE=replay` with no provider keys to confirm the cassette is sufficient.
+
+- Volatile fields in request bodies (e.g. AI-SDK `experimental_generateMessageId`) need a per-scenario filter. Add the scenario name and a `FilterSpec` to `e2e/helpers/cassette-filters.mjs`. The cassette layer is backed by `@braintrust/seinfeld` (`dev-packages/seinfeld`); the preload entry point is `e2e/helpers/cassette-preload.mjs`.
```
lforst, line 55: Can we add to the skill that the agent should always check the cassettes for potentially leaked API keys when generating them and redact them if they would be leaked if committed?

dev-packages/seinfeld/src/format/v1.ts
```diff
@@ -0,0 +1,69 @@
+import { z } from "zod";
+
+/**
+ * Zod schema for cassette format version 1.
```
lforst, line 4: Is there a reason for us to have versions?

dev-packages/seinfeld/scripts/migrate-from-legacy.mjs
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env node
```
lforst, line 1: Do we still need this file?

dev-packages/seinfeld/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Stephen Belanger
```
lforst, line 3: I feel like we don't need a license here because we already have one in the repo and this should most likely be Braintrust LLC if anything

dev-packages/seinfeld/scripts/migrate-from-legacy.mjs
AbhiPrasad, line 1: can we remove this file now that we've migrated this?

e2e/package.json
```diff
@@ -7,16 +7,20 @@
...
     "test:e2e:hermetic": "vitest run --tags-filter=hermetic",
     "test:e2e:canary": "node ./scripts/run-canary-tests-docker.mjs",
-    "test:e2e:update": "vitest run --update"
+    "test:e2e:update": "vitest run --update",
+    "test:e2e:record": "node ./scripts/record-cassettes.mjs"
   },
   "devDependencies": {
     "@braintrust/langchain-js": "workspace:^",
     "@braintrust/otel": "workspace:^",
+    "@braintrust/seinfeld": "workspace:^",
```
lforst, line 16: I don't know how I feel about having to build that package before being able to run e2e tests but I won't block on this now.

lforst (approved):
A few things I think we could clean up but otherwise lgtm

e2e/helpers/scenario-harness.ts
```diff
@@ -471,6 +641,57 @@ export async function withScenarioHarness(
...
+      cassetteOpt === true || typeof cassetteOpt === "object";
+    const fileExists = existsSync(cassettePath);
+    if (!explicitOptIn && !fileExists && !isRecordingMode) {
+      return {};
+    }
+
+    const normalizerName = config.normalizerName ?? scenarioName;
+
+    return {
+      ...getProviderKeyPlaceholders(),
```
AbhiPrasad, line 685: I think we need to gate this on `cassetteMode === "replay"`

e2e/helpers/cassette-filters.mjs
```diff
@@ -0,0 +1,94 @@
...
+
+export const CASSETTE_FILTERS = {
+  default: "default",
+  "ai-sdk": ["default", AI_SDK_VOLATILE_FIELDS],
+  "ai-sdk-instrumentation": ["default", AI_SDK_VOLATILE_FIELDS],
+  "ai-sdk-otel-export": ["default", AI_SDK_VOLATILE_FIELDS],
+  "mistral-instrumentation": ["default", MISTRAL_VOLATILE_FIELDS],
+  "openrouter-agent-instrumentation": ["default", OPENROUTER_VOLATILE_FIELDS],
+  "openrouter-instrumentation": ["default", OPENROUTER_VOLATILE_FIELDS],
+};
```
AbhiPrasad, line 94: can we make these filters configured in the scenario? So they are closer to the integration they are testing instead of being global like this?

dev-packages/seinfeld/README.md
```diff
@@ -0,0 +1,302 @@
...
+
+For cassettes committed to version control, use the `'paranoid'` preset, which covers all three paths:
+
+```ts
+createCassette({ name: "demo", redact: "paranoid" });
+```
+
+`'paranoid'` redacts credential headers, common credential field names at any JSON depth (`apiKey`, `token`, `secret`, `password`, `authorization`), and Bearer / `sk-` style tokens in text bodies.
+
+To detect misconfigurations at record time, add `strict: true`:
```
AbhiPrasad, line 28: this feels like it should just be the default behaviour. Also feels like paranoid should be the default behaviour. I would only change redact if I'm removing something imo.

PR #1933, opened 2026-05-03
feat: Add Dataset pipelines (experimental)

ankrgyl:
Internal design doc: https://www.notion.so/braintrustdata/Dataset-pipelines-34df78580289803e801eef0ea689bc44


js/src/dataset-pipeline.ts
```diff
@@ -0,0 +1,101 @@
...
+function registry(): DatasetPipelineDefinition[] {
+  if (!globalThis.__braintrust_dataset_pipelines) {
+    globalThis.__braintrust_dataset_pipelines = [];
+  }
+  return globalThis.__braintrust_dataset_pipelines;
+}
+
+export function getRegisteredDatasetPipelines(): DatasetPipelineDefinition[] {
+  return [...registry()];
+}
```
AbhiPrasad: why do we have this? I think we can remove it, plus it's associated functionality.

js/src/trace.ts
AbhiPrasad, line 1: I think we need to update https://github.com/braintrustdata/braintrust-sdk-javascript/blob/eebc78d9e46c6c4fa29bd189250d8151ff3940f5/js/src/trace.ts#L276 here as well

js/src/dataset-pipeline.ts
ankrgyl: we use this so that you don't have to export each dataset pipeline. it's an analogous thing to how we extract `Eval` body definitions

js/src/trace.ts
ankrgyl: fixed

AbhiPrasad (approved):

lforst:
Gonna take this over making sure that it's clear to consumers that this is experimental/beta for now.

ankrgyl:
> Gonna take this over making sure that it's clear to consumers that this is experimental/beta for now.

Oh great. I landed the python change before seeing this. Do you mind marking it as such in python too, and then whenever you are comfortable, we can land this change as well?

js/src/dataset-pipeline.ts
lforst: what is the exact use-case? I think I kinda get it but it honestly doesn't feel very JSy

lforst: Oh i see. This is used for `bt`. Honestly, to avoid API creep (and reduce confusion for SDK users/agents), I would probably just have the global registry and just have `bt` read off it.

lforst (approved):

PR #1956, opened 2026-05-06
perf(e2e): parallelize scenario variants within test files

Qard:
## Summary

- Wraps the `for`-loop in each multi-variant e2e scenario file in `describe.concurrent` so all SDK versions run their `beforeAll` subprocess hooks in parallel, bounded by `maxConcurrency: 5` in `vitest.config.mts`
- Uses `describe.sequential` for the per-version inner block to keep "wrapped" and "auto-hook" variants sequential within each version, preventing snapshot write races on `--update`
- Adds `js/scripts/run-parallel.mjs`, a cross-platform Node.js script that runs the isolated provider test suites (`ai-sdk-v5`, `ai-sdk-v6`, `claude-agent-sdk`) concurrently; the shared-workspace suites (openai, anthropic, google-genai) stay sequential since they call `pnpm prune` on a shared `node_modules`

**Expected CI wins:**
- `anthropic-instrumentation`: 6 SDK versions now run in parallel (was sequential)
- `ai-sdk-instrumentation`: 4 SDK versions in parallel
- `openai-instrumentation`: 3 versions in parallel
- `claude-agent-sdk-instrumentation`: 4 versions in parallel within each of the two outer groups
- `test:external` provider suite: ai-sdk v5/v6 and claude-agent-sdk now overlap with each other instead of chaining

All within the existing CI job structure — no new GitHub Actions jobs.

## Test plan

- [x] Hermetic e2e tests pass locally (`pnpm test:e2e:hermetic` — 13 passed, 908 skipped awaiting API keys)
- [ ] Full e2e CI run to confirm cassette-backed tests pass concurrently
- [ ] `js-provider-tests` CI run to confirm `run-parallel.mjs` exit propagation works on both ubuntu and windows


.changeset/parallelize-e2e-scenarios.md
```diff
@@ -0,0 +1,5 @@
+---
+"braintrust": patch
+---
+
+perf(e2e): parallelize scenario variants within test files
```
lforst: I don't think we need a changeset for this. Maybe you can add to the agents md that only user-facing changes need a changeset.

lforst (approved):

.changeset/parallelize-e2e-scenarios.md
Qard: Our CI currently requires it though. Maybe we need a specific label to explicitly opt-out of that or something?

lforst: We have a label! The ci job should also print that you can circumvent the failure with a label.

PR #1969, opened 2026-05-08
fix OpenAI body already red

viadezo1er:
OpenAI's `chat.completions.parse()` calls `create()._thenUnwrap(...)`, which produces two `APIPromise` instances sharing the same responsePromise.

Our SDK called `.then()` on each of them, triggered a `parseResponse`, which calls `response.json()` on the HTTP response body. However a `Response` body can only be read once.

Might fix a customer issue https://braintrustdata.slack.com/archives/C0B27UX9UDR/p1778223683547239

### DIFF
--- a/.changeset/dirty-hotels-show.md
+++ b/.changeset/dirty-hotels-show.md
@@ -1,0 +1,9 @@
+---
+"braintrust": patch
+---
+
+OpenAI's `chat.completions.parse()` calls `create()._thenUnwrap(...)`, which produces two `APIPromise` instances sharing the same responsePromise.
+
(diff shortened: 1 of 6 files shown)


viadezo1er:
CI e2e fails due to invalid COPILOT_API_KEY/BRAINTRUST_E2E_MODEL_BASE_URL in github secrets.
e2e-canary also happens on the main branch.

Qard (approved):
This feels a bit hack-ish, but the APIPromise stuff has been a pain...I'm not sure if there's a non-hacky solution. 😐

PR #1972, opened 2026-05-08
More explicit default git settings

ankrgyl:


js/src/logger.ts
```diff
@@ -5699,7 +5698,8 @@ function _saveOrgInfo(
       state.orgName = org.name;
       state.apiUrl = iso.getEnv("BRAINTRUST_API_URL") ?? org.api_url;
       state.proxyUrl = iso.getEnv("BRAINTRUST_PROXY_URL") ?? org.proxy_url;
-      state.gitMetadataSettings = org.git_metadata || undefined;
+      state.gitMetadataSettings =
+        org.git_metadata || defaultGitMetadataSettings();
```
manugoyal: Do we really need to default these settings in both the SDKs and the control
plane? I worry about maintining duplicate copies of defaults across our control
plane and SDKs in case we want to revise this.

If we are always defaulting in the control plane now, maybe we can defensively
switch this to not collecting anything if the `apikey/login` endpoint returns
nothing?

manugoyal (commented):
Wondering if we really need the defaults in the SDK level too? Same applies to the python sdk

js/src/logger.ts
alexr17: @manugoyal given that this PR merged with the same behavior: https://github.com/braintrustdata/braintrust-sdk-python/pull/408

Should we change that one as well or keep what's currently in this patch?

AbhiPrasad: It doesn't feel too bad to duplicate the defaults here, especially since I can't see this changing too much after now that it's an explicit allowlist. Collecting info about the branch/commit/author feels like as a reasonable default, and I can see it be valuable for users.

Maintaining SDK consistency for this should be fine, we can just add an entry to https://github.com/braintrustdata/braintrust-spec to validate this.

manugoyal: Maybe it is unlikely to change. But it also feels unnecessary to duplicate the defaults if we're modifying the control plane to always provide the explicit allowlist to the sdks. So the default won't even be used outside of edge cases I'm unaware of.

PR #1991, opened 2026-05-12
chore(deps): align bundler and test tooling versions

AbhiPrasad:
Upgrade bundler-related dependencies and pin Vitest across workspace packages so pnpm resolves a single version for the shared test toolchain. Align esbuild through a pnpm override because Vite and tsup otherwise resolve different compatible versions.


dev-packages/seinfeld/package.json
```diff
@@ -50,8 +50,8 @@
   },
   "devDependencies": {
     "@types/node": "^22.10.0",
-    "tsup": "^8.3.5",
+    "tsup": "^8.5.1",
     "typescript": "^5.7.2",
-    "vitest": "^2.1.8"
+    "vitest": "4.1.5"
```
realark, line 55: should these be variables or somehow consolidated somewhere?

realark (approved):

dev-packages/seinfeld/package.json
AbhiPrasad: they should, let me see if I can leverage pnpm to make this better.

AbhiPrasad: I tried to use the pnpm feature for this, catelogs, but it didn't really work well. Gonna merge this in and re-visit afterwards.

PR #1993, opened 2026-05-12
feat(vitest): Support projectId in wrapVitest config

evanmkeith:
## Summary

Adds an optional `projectId` field to the `wrapVitest` configuration, letting users target a specific Braintrust project by its stable ID instead of by name. When `projectId` is set, it takes precedence over `projectName`.

## Motivation

Project names can be renamed or collide across orgs. Users who already know the project ID (e.g. from CI config or another integration) had no way to bind Vitest experiments to that exact project. This brings the wrapper in line with `init` / `initExperiment`, which both accept `projectId`.

## Changes

- `js/src/wrappers/vitest/types.ts` — Add `projectId?: string` to `WrapperConfig` with a docstring noting precedence over `projectName`.
- `js/src/wrappers/vitest/wrapper.ts` — In `wrapDescribe`'s lazy `getOrCreateContext`, branch on `config.projectId`:
  - If set, call `initExperiment({ projectId, experiment })`.
  - Otherwise, preserve the existing `initExperiment(projectName ?? suiteName, { experiment })` call (no behavior change for existing users).
- `js/src/wrappers/vitest/README.md` — Document the new `projectId` parameter on `wrapVitest`.
- `js/src/wrappers/vitest/vitest-wrapper.test.ts`:
  - Update the `initExperiment` mock to handle both overloads (positional `projectName` and options object).
  - Add `configureNode()` at module load so AsyncLocalStorage actually propagates context — required for the new tests' synchronous fake `describe` factories to exercise experiment creation.
  - Add a `Project selection` describe block with four tests:
    1. `projectId` is forwarded to `initExperiment`.
    2. `projectId` takes precedence when both `projectId` and `projectName` are set.
    3. `projectName`-only path is unchanged.
    4. Suite-name fallback (no project config) is unchanged.

## Usage

```ts
import * as vitest from "vitest";
import { wrapVitest } from "braintrust";

const bt = wrapVitest(vitest, {
  projectId: "abc-123-stable-project-id",
});

bt.describe("my suite", () => {
  bt.test("logs to the project by id", async () => {
    /* ... */
  });
});

### DIFF
--- a/.changeset/curvy-grapes-admire.md
+++ b/.changeset/curvy-grapes-admire.md
@@ -1,0 +1,5 @@
+---
+"braintrust": minor
+---
+
+feat(vitest): Support projectId in wrapVitest config
--- a/js/src/wrappers/vitest/README.md
+++ b/js/src/wrappers/vitest/README.md
@@ -314,6 +314,7 @@ wrapVitest(
   },
   config?: {
     projectName?: string;        // Project name (defaults to suite name)
+    projectId?: string;          // Stable project ID; takes precedence over projectName
     displaySummary?: boolean;    // Show summary after tests (default: true)
     onProgress?: (event: ProgressEvent) => void;  // Progress callback
   }
--- a/js/src/wrappers/vitest/types.ts
+++ b/js/src/wrappers/vitest/types.ts
@@ -154,6 +154,11 @@ export interface BraintrustVitest<
 export interface WrapperConfig {
   projectName?: string;
   /**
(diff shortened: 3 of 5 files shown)


cpinn (approved):
Thank you for making the change.

AbhiPrasad (commented):
needs a changeset and we can merge this in!

AbhiPrasad:
released with https://github.com/braintrustdata/braintrust-sdk-javascript/releases/tag/braintrust%403.11.0!

PR #2003, opened 2026-05-13
chore(cli): replace chalk and boxen with native formatting

AbhiPrasad:
Use Node's native styleText for CLI colors and a small local formatter for the experiment summary box. Also pin @braintrust/browser to the workspace braintrust package so local dependency changes are reflected during installs.

Our SDK is Node 20+, so these changes are safe.


viadezo1er (approved):
stamp

PR #2011, opened 2026-05-18
chore: Record stable span tree snapshots instead of heavily normalized payloads

lforst:
Replaces our admittedly random snapshots with received span tree snapshots that should make it better to reason about changes.


Qard (commented):
Why use this _mostly_ JSON text format rather than just using _actual_ JSON with just a stable structure so we can actually do structural assertions on it? That way when something goes wrong we can better understand _why_ and not just get some indecipherable string match failure. 🤔

lforst:
> Why use this _mostly_ JSON text format rather than just using _actual_ JSON with just a stable structure so we can actually do structural assertions on it? That way when something goes wrong we can better understand _why_ and not just get some indecipherable string match failure. 🤔

A big motivation behind this change was to quickly understand what the instrumentation spits out. While a JSON gets us there, I found this format better for mentally parsing things like span type and especially hierarchy/relationships between spans. While getting this PR cleaned up, I haven't found diffs to be too hard to understand, however I see your point.

Want me to change it?

Qard:
Can we have that as a debug output or something? I can see some value to that too, but I think for what actually gets asserted against I would slightly prefer something more structural.

lforst:
@Qard changed it so we are emitting both.

I think the txt files are great for PR reviews so I still wanted it to be tracked via a file and not just process output.
Additionally, I wanted the json and the txt to always be in sync - which technically means that we also assert on the txt.
I am happy with this, even though it is a bit verbose.

AbhiPrasad (approved):

lforst:
@Qard fyi, I made Abhi approve the PR to unblock me. We can rip out either text or json pretty easily.

Qard:
No worries. I tried to review, but 120k+ lines takes a _long_ time to review properly. 😐

PR #2032, opened 2026-05-20
feat(bundler-plugins): Add `braintrust{Bundler}Plugin` aliases for bundler plugins and deprecate rest

lforst:
Reason: In the context of a vite config, it makes no sense to have a vite plugin named `vitePlugin`. There is an argument to be made around calling it `braintrustPlugin`, however I imagine agents have a better time understanding and importing stuff when the different plugins are actually named differently.

Also the default value of the `browser` option is insane since we don't even have a working browser sdk.


js/src/wrappers/vitest/package.json
```diff
@@ -1,6 +1,5 @@
 {
   "name": "@braintrust/vitest-wrapper-tests",
-  "version": "0.0.1",
```
lforst, line 3: Snuck these changes in because these packages were polluting `pnpm run changeset` output and the only way to prevent that is by not giving the packages version numbers.

Also it's one extra safeguard to not accidentally publish these packages I guess.

dev-packages/seinfeld/src/cassette.ts
```diff
@@ -75,8 +75,6 @@ export interface CassetteEntry {
 export interface CassetteMeta {
   /** ISO-8601 timestamp when the cassette was first created. */
   createdAt: string;
-  /** The seinfeld version that produced the cassette. */
-  seinfeldVersion: string;
```
AbhiPrasad, line 79: we also gotta update `dev-packages/seinfeld/src/format/v1.ts`

.changeset/tangy-waves-find.md
```diff
@@ -0,0 +1,5 @@
+---
+"braintrust": minor
+---
+
+feat(bundler-plugins): Add `braintrust{Bundler}Plugin` aliases for bundler plugins and deprecate rest
```
AbhiPrasad: I'd be more explicit here and list all the exports that were renamed.

js/src/auto-instrumentations/bundler/esbuild.ts
```diff
@@ -1,24 +1,26 @@
...
+}
 
-export type EsbuildPluginOptions = BundlerPluginOptions;
+export type EsbuildPluginOptions = LegacyBundlerPluginOptions;
 
+/**
+ * @deprecated Use {@link braintrustEsbuildPlugin} instead. This legacy export
+ * defaults to browser-compatible diagnostics channel shimming when `browser`
+ * is omitted; `braintrustEsbuildPlugin` defaults to Node.js diagnostics_channel
+ * unless `useDiagnosticChannelCompatShim` is set to `true`.
```
AbhiPrasad: feel like this stuff doesn't really matter to the user. I'd just remove the diagnostics channel note all together. Or word it to be more clear about user impact

js/src/auto-instrumentations/bundler/plugin.ts
```diff
@@ -57,6 +44,26 @@ export interface BundlerPluginOptions {
...
+   * of Node.js's built-in `diagnostics_channel` module.
+   *
+   * Enable this for browser, edge, or worker bundles where Node's
+   * `diagnostics_channel` module is unavailable. Leave it disabled for Node.js
+   * bundles so transformed SDK code publishes on the native `diagnostics_channel`
+   * registry.
+   *
+   * @default false
+   */
+  useDiagnosticChannelCompatShim?: boolean;
```
AbhiPrasad, line 55: should this be a top level option? Or can we prefix with internal or similar to make it more clear that most users shouldn't really mess with this?

lforst: This is not an internal option, also I don't fully like steering users away from an option. Technically very valid to use if they are on cloudflare or whatever.

dev-packages/seinfeld/src/cassette.ts
lforst: Good catch, the rabbit hole went even deeper cad5d4af4fe8145b759a731b89299c3e698fe846

.changeset/tangy-waves-find.md
lforst: 203906c182a96952eb2b7853629028df7de97fb0

js/src/auto-instrumentations/bundler/esbuild.ts
lforst: Yeah, agree. Will remove.

lforst: 6a5d12c836603b6d4472f2c9d7c3aa4ec47b3859

AbhiPrasad (approved):

PR #2033, opened 2026-05-20
feat(nextjs): Add `wrapNextjsConfigWithBraintrust` as canonical setup utility instead of webpack loader/plugin

lforst:
Adds a new canonical way of setting up auto-instrumentation for Next.js: `wrapNextjsConfigWithBraintrust`

It takes care of automatically setting up custom loaders depending on nextjs version and bundler used (webpack, turbopack).


AbhiPrasad (approved):
somethings never change

lforst:
I wonder how long I can base my career on functions that wrap the Next.js config

PR #2066, opened 2026-05-28
chore: declare node engine requirements

AbhiPrasad:
Make it clear that we support Node 20+ in the JS SDK

### DIFF
--- a/.changeset/green-nodes-serve.md
+++ b/.changeset/green-nodes-serve.md
@@ -1,0 +1,12 @@
+---
+"braintrust": patch
+"@braintrust/browser": patch
+"@braintrust/langchain-js": patch
+"@braintrust/openai-agents": patch
+"@braintrust/otel": patch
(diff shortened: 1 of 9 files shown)


lforst (commented):
This change gives me a feel of dread - sorry that it comes so late. Every time we did this in the past, even if we thought it should be fine, people ended up complaining.

I am wondering whether for now we should just document a minimum version requirement and only add the engine with the next major?

Qard (approved):
Regardless of potential user impact, I would generally always consider the change of a minimum installable version to be a major change. Is there any reason not to just do a major _now_ or soon-ish? The change itself LGTM, just a question of when to ship it?

AbhiPrasad:
> I am wondering whether for now we should just document a minimum version requirement and only add the engine with the next major?

yeah I'm fine with doing that. I updated https://github.com/braintrustdata/braintrust-sdk-javascript/issues/1841

PR #2067, opened 2026-05-28
feat: Include `bt` cli as part of `braintrust` package

viadezo1er:
~Depends on https://github.com/braintrustdata/bt/pull/208.~ Merged


lforst (dismissed):
This is not what we want. Updates should be manual as a minor in bt is a minor in sdk, and a major in bt, is a major in sdk.

js/bin/bt.cjs
lforst, line 1: This should not be a .cjs file I think. In general, let's follow https://github.com/getsentry/sentry-cli/blob/master/scripts/install.js or https://blog.sentry.io/publishing-binaries-on-npm/ to the tee. The logs are very intentional there and we should log similar things to ensure people are not confused when installing the package and things fail.

I also have an example repo on how this should look like: https://github.com/lforst/npm-binary-example

viadezo1er:
We'll stick with manual version update for now.

viadezo1er:
Forgot we need to change the `package.json` to expose the `bt` binaries so we still need this PR.

js/scripts/bt-helper.js
```diff
@@ -0,0 +1,139 @@
...
+
+const BINARY_DISTRIBUTIONS = [
+  { packageName: "@braintrust/bt-darwin-arm64", subpath: "bin/bt" },
+  { packageName: "@braintrust/bt-darwin-x64", subpath: "bin/bt" },
+  { packageName: "@braintrust/bt-linux-arm64", subpath: "bin/bt" },
+  { packageName: "@braintrust/bt-linux-x64", subpath: "bin/bt" },
+  { packageName: "@braintrust/bt-linux-x64-musl", subpath: "bin/bt" },
+  { packageName: "@braintrust/bt-win32-arm64", subpath: "bin/bt.exe" },
+  { packageName: "@braintrust/bt-win32-x64", subpath: "bin/bt.exe" },
+];
```
lforst, line 19: we need to update these to reflect the actual packages

js/scripts/bt-helper.js
```diff
@@ -0,0 +1,139 @@
...
+  } catch {
+    return "glibc";
+  }
+}
+
+function binaryName() {
+  return process.platform === "win32" ? "bt.exe" : "bt";
+}
+
+function getDistributionForThisPlatform() {
```
choochootrain, line 38: slop but looks correct

lforst: bro what do you mean slop. I wrote this code before Slop was a thing https://github.com/getsentry/sentry-cli/blame/8c2c0e9a3cc0807e9fc8cd14fbc8456dbd1fd612/lib/helper.ts#L37

your comment is going on a list

lforst (approved):

PR #2069, opened 2026-05-29
ci: Update actions to use node 24

lforst:
Github is warning in CI jobs with:

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Let's upgrade to 24 except for where we test specific versions and dd the `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION` version until we move off it as minimum version. Then we can remove the tests.

### DIFF
--- a/.github/workflows/changeset.yaml
+++ b/.github/workflows/changeset.yaml
@@ -21,10 +21,10 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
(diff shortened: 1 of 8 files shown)


lforst:
Screenshot 2026-05-29 at 12 40 25

Note also fixing this I guess

Qard (approved):

PR #2091, opened 2026-06-03
fix(stream): preserve multi-byte UTF-8 split across chunk boundaries

stretpjc:
## Summary

`BraintrustStream` corrupted multi-byte UTF-8 characters when an SSE chunk boundary landed mid-character (e.g. Cloudflare `workerd` flushing a small first chunk), emitting `U+FFFD` instead of the decoded character. 

Root cause: `btStreamParser` called `decoder.decode(chunk)` without `{ stream: true }`, so the `TextDecoder` flushed its internal buffer on every chunk and dropped any trailing partial UTF-8 sequence. 

## Fix

Pass `{ stream: true }` when decoding each `Uint8Array` chunk so partial sequences are held until the next chunk arrives, and flush the decoder once in the transform's `flush` handler to handle a truly-truncated tail. No behavior change for `string` or `BraintrustStreamChunk` inputs, and single-chunk UTF-8 input is byte-identical to before.

## Tests

Added a regression test in `js/src/functions/stream.test.ts` that splits an SSE `text_delta` event inside the 3-byte sequence for `U+201C` and asserts a clean round-trip with no `U+FFFD`. Fails on `main`, passes here.

### DIFF
--- a/.changeset/bright-falcons-carry.md
+++ b/.changeset/bright-falcons-carry.md
@@ -1,0 +1,5 @@
+---
+"braintrust": patch
+---
+
+fix(stream): Preserve multi-byte UTF-8 split across chunk boundaries
--- a/js/src/functions/stream.test.ts
+++ b/js/src/functions/stream.test.ts
@@ -55,6 +55,38 @@ test("final value passthrough", async () => {
   }
 });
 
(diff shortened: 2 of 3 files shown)


lforst (approved):
Thanks! <3

LiuLawrence45:
GOAT

PR #2094, opened 2026-06-04
Parameters update values

j13huang:
When building parameters, this line https://github.com/braintrustdata/braintrust-sdk-javascript/blob/dfb709b80db0a2b48fb36eea7451f0d58b1fce8c/js/src/framework2.ts#L633

Means that the parameter data we upload to braintrust via `bt functions push` will always replace existing parameter data even if the user has made updates to the parameters via the UI or API. That also means setting default values in the parameters schema is the only way to update the underlying parameter data via the sdk. I think default values and actual stored data should be decoupled.

This PR adds `parametersObj.addValues({})` which allows users to use `bt functions push` to update existing fields in a PATCH/merge style update rather than a PUT/replace update.

Aligns with https://github.com/braintrustdata/bt/pull/225. Comments on API/interface welcome

### DIFF
--- a/js/src/eval-parameters.ts
+++ b/js/src/eval-parameters.ts
@@ -7,23 +7,35 @@ import {
 } from "./prompt-schemas";
 import { PromptData as promptDataSchema } from "./generated_types";
 
+// Schema for a prompt-typed evaluation parameter
+export const promptParameterSchema = z.object({
+  type: z.literal("prompt"),
+  default: promptDefinitionWithToolsSchema.optional(),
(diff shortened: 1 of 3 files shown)


AbhiPrasad (commented):
It seems like `updateValues` doesn't merge, it just replaces. I would expect that it would have merge semantics.

what if we expose:

- `clearAndSetValues`: basically what `updateValues` does here right now (can also be called `replaceValues`, but I prefer to be super explicit)
- `updateValues`: which has merge semantics

PR #2095, opened 2026-06-05
Add classifier function push builder

Qard:
## Summary
- add a project-level classifier builder for code-backed saved functions
- register classifiers through the existing function push registry with `type: "classifier"`
- add focused coverage for normal and lazy-load function registry registration

## Validation
- `./node_modules/.bin/vitest run src/framework.test.ts -t "classifier registers|lazy classifier"`
- `./node_modules/.bin/tsc --noEmit` was attempted, but the local Node/type environment reports existing `node:util.styleText` and `diagnostics_channel.tracingChannel` type errors unrelated to this change

### DIFF
--- a/js/src/exports.ts
+++ b/js/src/exports.ts
@@ -231,7 +231,6 @@ export {
 export { DatasetPipeline } from "./dataset-pipeline";
 
 export type {
-  ClassifierOpts,
   CodeOpts,
   CreateProjectOpts,
   FunctionEvent,
@@ -240,7 +239,6 @@ export type {
 } from "./framework2";
 
 export {
-  ClassifierBuilder,
   CodeFunction,
   CodePrompt,
   Project,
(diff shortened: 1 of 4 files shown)


js/src/exports.ts
```diff
@@ -231,6 +231,7 @@ export {
 export { DatasetPipeline } from "./dataset-pipeline";
 
 export type {
+  ClassifierOpts,
```
lforst: I don't think we should export the options. Just unnecessary API surface for very little gain.

js/src/exports.ts
```diff
@@ -239,6 +240,7 @@ export type {
 } from "./framework2";
 
 export {
+  ClassifierBuilder,
```
lforst: I also don't think we should export this.

lforst (commented):
Discussion about general API design offline. My current state of mind is that I am not sure we should introduce `Project.classifiers.create()` because technically it would be a scorer and we already have `Project.scorers.create()`.

lforst (approved):
If we rm the exports this is good.

lforst (approved):

PR #2104, opened 2026-06-09
Allow inline eval cases to carry explicit origin metadata.

max-braintrust:
Previously, eval span origins were inferred only when running directly from a `Dataset`, using the dataset row id/xact metadata on each datum. Callers that resolve or transform rows before invoking `Eval` as inline data had no generic way to preserve the original source-row association.

This change lets an `EvalCase` provide an explicit `origin`, and keeps the existing dataset-derived origin inference as a fallback. That enables batched inline eval flows to preserve per-row provenance without changing behavior for normal dataset-backed evals.

### DIFF
--- a/js/src/framework.ts
+++ b/js/src/framework.ts
@@ -1169,6 +1169,17 @@ async function runEvaluatorInternal(
           : Dataset.isDataset(evaluator.data)
             ? evaluator.data
             : undefined;
+        const origin =
+          datum.origin ??
+          (eventDataset && datum.id && datum._xact_id
+            ? {
+                object_type: "dataset",
+                object_id: await eventDataset.id,
+                id: datum.id,
+                created: datum.created,
+                _xact_id: datum._xact_id,
+              }
+            : undefined);
 
         const baseEvent: StartSpanArgs = {
           name: "eval",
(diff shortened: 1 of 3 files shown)


js/src/framework.ts
```diff
@@ -1169,6 +1169,17 @@ async function runEvaluatorInternal(
           : Dataset.isDataset(evaluator.data)
             ? evaluator.data
             : undefined;
+        const origin =
+          datum.origin ??
```
j13huang, line 1173: one question I have here:
is there some way we could do something like only take the object_type and object_id from the base row rather than the whole origin - since technically the origin id should always be the row id? That way we don't need to change the interface

Something like that could convert dataset_id on the base row -> object_type: dataset, object_id: dataset_id

max-braintrust: I think we'd still have to change the API surface to do something like this - right now we are passing these rows inline, so there is no dataset_id in the eval input: https://github.com/braintrustdata/braintrust-sdk-javascript/blob/ece3b320535fec47c436a1001f8364b2f7865bd5/js/src/generated_types.ts#L1875 

I don't think there is any reference to the base dataset in the existing shape that we could use - I could add that instead of the origin change, but then I think we are still basically implicitly changing the contract.

j13huang (approved):

PR #2109, opened 2026-06-10
fix(anthropic): Log Anthropic system message first in span input

joshua-tj:
## Problem

The Anthropic instrumentation merges the top-level `system` param into the logged `input` messages array (since the Anthropic API takes it separately rather than as a message). But `coalesceInput` appends it **after** the conversation messages, so every `anthropic.messages.create` span logs:

```json
[
  { "role": "user", "content": "..." },
  { "role": "system", "content": "..." }
]
```

This doesn't match what the model actually receives (the API applies the system prompt first), and it makes logged calls awkward to reuse: opening a logged span in the playground requires manually reordering the messages before rerunning, since chat-message convention puts system first.

## Fix

`coalesceInput` now prepends the system message instead of appending it. Logged inputs read `[system, user, ...]`.

- Exported `coalesceInput` and added unit tests (system-first ordering, system as text-block array, no system prompt, no mutation of the caller's array) — it previously had no direct coverage.
- Regenerated the anthropic-instrumentation e2e span-tree snapshots via `test:e2e:update`; the diffs are purely the system message moving from last to first.

## Verification

- `js/` core suite: 1070 passed
- anthropic-instrumentation e2e scenario run twice: 184 passed both times
- `tsc --noEmit` and prettier clean

Note: this changes the logged `input` shape, so anything downstream that assumes the system message is last (e.g. saved playground sessions or scorers parsing `input`) will see the new order. The in-repo e2e assertions already look up the system message by role rather than position, so none needed changes.


lforst (approved):
Thank you!

joshua-tj:
> Thank you!

No worries! Thanks for merging!

PR #2140, opened 2026-06-18
Fix dataset-backed eval origin for copied dataset rows

evanmkeith:
## Summary

  Fix dataset-backed eval/playground origin handling so copied dataset rows use the active dataset row id instead of
  preserved source provenance.

  ## Details

  Dataset rows copied from logs can carry origin pointing at the original project_logs row. runEvaluator previously
  preferred that datum.origin over the synthesized dataset-row origin, causing playground output/progress events to be
  keyed by the source log row id instead of the selected dataset row id.

  This changes dataset-backed evals to prefer the active dataset row origin whenever eventDataset and datum.id are
  present. Inline/non-dataset evals still preserve explicit datum.origin.

  ## Tests

  - Added regression coverage for log-derived dataset rows using dataset row origin
  - Added coverage for dataset rows without _xact_id
  - Added coverage for streamed progress events carrying dataset row origin
  - Ran focused Vitest origin tests successfully via local Vitest binary

### DIFF
--- a/.changeset/beige-ducks-sneeze.md
+++ b/.changeset/beige-ducks-sneeze.md
@@ -1,0 +1,5 @@
+---
+"braintrust": patch
+---
+
+Fix: dataset-backed eval origin for copied dataset rows
--- a/js/src/framework.test.ts
+++ b/js/src/framework.test.ts
@@ -346,7 +346,7 @@ describe("runEvaluator", () => {
     ]);
   });
 
-  test("rejects an invalid inline origin", async () => {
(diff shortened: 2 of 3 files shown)


max-braintrust (approved):

j13huang:
updated the code a little bit. Feel free to merge after tests. Thanks again @evanmkeith!

max-braintrust:
Thanks for catching+fixing this! Added the changeset so it will pass - lmk if I can help FI.

evanmkeith:
Thank you both!!

PR #2141, opened 2026-06-18
feat: Add otel compatible distributed tracing

realark:
implements: https://github.com/braintrustdata/braintrust-spec/blob/main/docs/features/distributed-tracing.md

- use otel-friendly IDs by default in the SDK
- add inject/extract methods for distributed tracing instead of parent slug methods
    - implements: https://www.w3.org/TR/trace-context/

OLD WAY: https://www.braintrust.dev/docs/instrument/advanced-tracing#trace-distributed-systems
```typescript
import { currentSpan, initLogger, startSpan, traced } from "braintrust";

const logger = initLogger({ projectName: "my-project" });

// Client: Export the span
const processRequest = (request: Request) =>
  traced(
    async (span) =>
      fetch("/api/process", {
        method: "POST",
        body: JSON.stringify(request),
        headers: { "X-Trace-ID": await currentSpan().export() },
      }),
  );

// Server: Resume the trace
async function handleRequest(req: Request) {
  const traceId = req.headers.get("X-Trace-ID") ?? undefined;

  return traced(
    async (span) => {
      const result = await processData(req.body);
      span.log({ input: req.body, output: result });
      return result;
    },
    { parent: traceId },
  );
}
```

NOTE: the "old way" is still supported and backwards-compatible. Native SDKs can continue to link to each other in the manner, even if one is upgraded and the other is not. However if users with to link to otel sdks, they will have to change their code to use the new apis.

NEW WAY:
```typescript
import {
  currentSpan,
  extractTraceContext,
  initLogger,
  startSpan,
  traced,
} from "braintrust";

const logger = initLogger({ projectName: "my-project" });

// Client: Export the span
const processRequest = (request: Request) =>
  traced(
    async (span) =>
      fetch("/api/process", {
        method: "POST",
        body: JSON.stringify(request),
        headers: currentSpan().inject({}), // <--- new inject api
      }),
  );

// Server: Resume the trace
async function handleRequest(req: Request) {
  return traced(
    async (span) => {
      const result = await processData(req.body);
      span.log({ input: req.body, output: result });
      return result;
    },
    { parent: extractTraceContext(req.headers) }, // <--- new extract api
  );
}
```

# Other notes

- confusing naming: this has the unfortunate effect of the `root_span_id` field essentially becoming the trace id rather than the id of the root span (i.e. `root_span_id` != `rootSpan.id`). This sucks, but it's already the case for otel sdks and py otel compat mode. So at least we're consistent.
- legacy id flag: the sdk currently maintains two paths for id generation. it defaults to UUIDs and does hex IDs when we're in in otel compat mode. This changes the default to use hex IDs and has a flag to revert to the old behavior (`BRAINTRUST_LEGACY_IDS`). We could just rip out the old uuid logic entirely, but this approach seems like the least risky
- to support passing through tracestate and flags, an additional field was added to SpanImpl (propagatedState)

# examples

- https://www.braintrust.dev/app/Braintrust%20SDKs/p/distributed-tracing-test/trace?object_type=project_logs&object_id=d010dc4b-b4f2-421e-a165-232a8e6f062f&r=c7681ae1d7754e3fd27a06fb53c6fc29&s=decc9abf92ddeae5
- https://www.braintrust.dev/app/Braintrust%20SDKs/p/distributed-tracing-test/logs?r=6449255f3c9a8f7121cf9ce04c6d73f6&s=4180f4889809461e

### DIFF
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -5530,6 +5530,45 @@ function setHeader(
   headerBag[name] = value;
 }
 
+function deleteHeader(carrier: TraceContextCarrier, name: string): void {
+  const lowered = name.toLowerCase();
+  if (isMutableHeaderTupleArray(carrier)) {
+    for (let i = carrier.length - 1; i >= 0; i--) {
+      if (carrier[i][0].toLowerCase() === lowered) {
+        carrier.splice(i, 1);
+      }
+    }
+    return;
+  }
+
+  const deleter = (carrier as { delete?: unknown }).delete;
+  if (typeof deleter === "function") {
+    try {
+      deleter.call(carrier, name);
+      return;
+    } catch {
+      // Fall through to other header carrier styles.
+    }
+  }
+
+  const remover = (carrier as { removeHeader?: unknown }).removeHeader;
+  if (typeof remover === "function") {
+    try {
+      remover.call(carrier, name);
+      return;
+    } catch {
+      // Fall through to plain object cleanup.
+    }
+  }
+
+  const headerBag = carrier as { [name: string]: string };
+  for (const key of Object.keys(headerBag)) {
+    if (key.toLowerCase() === lowered) {
+      delete headerBag[key];
+    }
+  }
+}
+
 /**
  * Inject W3C trace-context headers into a carrier (in place).
  *
@@ -5573,6 +5612,13 @@ export function _injectIntoCarrier(
   const baggageValue = mergeBaggage(existing, args.braintrustParent);
   if (baggageValue !== undefined) {
     setHeader(carrier, BAGGAGE_HEADER, baggageValue);
+  } else if (existing !== undefined) {
+    // mergeBaggage returns undefined when the existing header only had a stale
+    // braintrust.parent (or otherwise no members worth forwarding), so remove
+    // the old header instead of letting a reused carrier propagate bad routing.
+    // If left in place, a downstream service could attach this trace to the
+    // previous project/experiment, or leak that parent to an unrelated request.
+    deleteHeader(carrier, BAGGAGE_HEADER);
   }
 }
 
--- a/js/src/logger.ts
+++ b/js/src/logger.ts
@@ -2359,6 +2359,12 @@ function startSpanParentArgs(args: {
       );
     }
 
+    // For logger/span method calls that receive a W3C parent, the inbound
+    // context links this span to the upstream trace ids, but the span still
+    // belongs to the current Braintrust project/experiment and should propagate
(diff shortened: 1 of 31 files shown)


js/src/propagation.ts
```diff
@@ -0,0 +1,395 @@
...
+/**
+ * Case-insensitive header lookup.
+ *
+ * Some frameworks normalize header names to title case (e.g. `Traceparent`)
+ * while the W3C keys are lowercase. Returns the first matching value or
+ * undefined. Array-valued headers (e.g. Node's `IncomingHttpHeaders`) are
+ * reduced to their first element, since these headers are single-valued.
+ */
+export function getHeader(
+  headers: Record<string, unknown> | null | undefined,
```
AbhiPrasad: we should also try to parse https://developer.mozilla.org/en-US/docs/Web/API/Headers here (we can duck type it).

js/src/propagation.ts
```diff
@@ -0,0 +1,395 @@
...
+      const cost = utf8ByteLength(member) + (relayed.length ? 1 : 0);
+      if (length + cost > byteBudget) {
+        break;
+      }
+      relayed.push(member);
+      length += cost;
+    }
+  }
+
+  const members = btMember !== undefined ? [...relayed, btMember] : relayed;
```
AbhiPrasad, line 509: looks like `btMember` is always added. Shouldn't we check to make sure it is not too large?

js/src/logger.ts
```diff
@@ -5153,38 +5244,410 @@ export function currentSpan(options?: OptionalStateArg): Span {
...
+
+  const context: PropagationContext = { [TRACEPARENT_HEADER]: traceparent };
+  const baggageValue = getHeader(headers, BAGGAGE_HEADER);
+  if (baggageValue) {
+    context[BAGGAGE_HEADER] = baggageValue;
+  }
+  const tracestate = getHeader(headers, TRACESTATE_HEADER);
+  if (tracestate) {
+    context[TRACESTATE_HEADER] = tracestate;
+  }
```
AbhiPrasad, line 5719: should we do validation on `tracestate` here?

js/src/propagation.ts
realark: @lforst took care of this (thanks Luca!)

js/src/propagation.ts
realark: @lforst took care of this (thanks Luca!)

js/src/logger.ts
realark: @lforst started this (thanks Luca!) and I just finished it up with some tests

AbhiPrasad (approved):

PR #2177, opened 2026-07-01
feat(ai-sdk): Add Workflow Agent instrumentation

### LINKED ISSUE #2167: Support AI SDK v7 (WorkflowAgent)
Similar to supporting v6 and the ToolLoopAgent with `wrapAiSDK` it would be amazing to support it for the new v7 SDK

lforst:
Fixes https://github.com/braintrustdata/braintrust-sdk-javascript/issues/2167


js/src/instrumentation/plugins/ai-sdk-plugin.ts
```diff
@@ -2049,6 +2847,31 @@ function isAsyncGenerator(value: unknown): value is AsyncGenerator {
...
+  }
+
+  const prototype = Object.getPrototypeOf(value);
+  if (prototype !== Object.prototype && prototype !== null) {
+    return value;
+  }
+
+  const sanitized: Record<string, unknown> = {};
+  for (const [key, nested] of Object.entries(value)) {
+    if (key.toLowerCase() === "headers") {
```
AbhiPrasad: this seems pretty aggressive. Can't we do similar logic to what we have in `sanitizeAISDKMetadataValue`?

e2e/scenarios/ai-sdk-instrumentation/scenario.impl.mjs
```diff
@@ -201,6 +202,51 @@ function createWeatherTool(ai, schemaKey) {
...
+              condition: "sunny",
+              location,
+              temperatureC: 22,
+            }),
+          }),
+        },
+      });
+
+      await agent.stream({
+        messages: [
```
AbhiPrasad, line 234: we should also add a test for something like

```ts
await agent.stream({                                                                                      
   system: "You are a helpful weather assistant.",                                                         
   prompt: "What's the weather in Paris?",                                                                 
 })
```

From what I can see I don't think we extract the prompt properly for that.

AbhiPrasad (approved):

PR #2181, opened 2026-07-01
Fix LangChain JS token total metrics

Qard:
## Summary

- Emit canonical `metrics.tokens` for LangChain JS LLM spans.
- Prefer explicit LangChain totals when present.
- Synthesize `tokens` from prompt and completion counts when LangChain omits an explicit total.
- Preserve existing cache token mappings.

## Root Cause

LangChain JS callback spans were already logging prompt and completion token metrics, but the callback handler only emitted the legacy `total_tokens` field for totals. Braintrust cost/token insights expect canonical `metrics.tokens`, so spans with complete prompt/completion data could still look like they were missing total token usage.

## Validation

- `mise exec -- pnpm --dir js exec vitest run src/wrappers/langchain/callback-handler.test.ts`
- `mise exec -- pnpm --dir js exec vitest run src/instrumentation/plugins/langchain-plugin.test.ts`
- `mise exec -- pnpm exec prettier --check js/src/wrappers/langchain/callback-handler.ts js/src/wrappers/langchain/callback-handler.test.ts .changeset/langchain-js-token-total-metrics.md`
- `git diff --check`


e2e/scenarios/wrap-langchain-js-traces/__snapshots__/span-tree.json
```diff
@@ -1259,6 +1263,7 @@
                 "completion_tokens": 21,
                 "prompt_cached_tokens": 0,
                 "prompt_tokens": 76,
+                "tokens": 97,
                 "total_tokens": 97
```
lforst: What's the actual diff between tokens and total_tokens?

lforst (approved):

PR #2187, opened 2026-07-01
feat: Expose flag for toggling realtime mode in span fetcher

choochootrain:


lforst:
For those curious, real time mode also reads uncompressed WAL entries in brainstore (fresher data but more expensive).

lforst (approved):

alexr17 (approved):

PR #2203, opened 2026-07-06
feat: Add first party eve instrumentation

lforst:
Ok this instrumentation is a bit interesting:
- Eve already has docs for braintrust but the traces it generates are not great. They're otel traces with wild disconnection issues.
- Eve is most likely deployed to Vercel workflows very often which means each step runs in it's own "function invocation", meaning we cannot rely on ALS.
- To combat the "can't use ALS" situation, we use the session/turn-IDs that eve gives us to produce hashes that we then use for our trace and span IDs.
- Also, because the turns are detached runtime-wise, we have a hard time creating a trace for each turn, which is arguably not what we want in any case (?) - so instead we attach all turns, tools, subagents, llm calls to a singular trace as they happen.

The way someone instruments their app is via:

```ts
// agent/instrumentation.ts
import { braintrustEveInstrumentation, initLogger } from "braintrust";
import { defineState } from "eve/context";
import { defineInstrumentation } from "eve/instrumentation";

export default defineInstrumentation(
  braintrustEveInstrumentation({
    defineState,
    setup: ({ agentName }) => {
      initLogger({
        projectName: 'project',
      });
    },
  }),
);

```

plus 

```ts
// agent/hooks/braintrust.ts
import { braintrustEveHook } from "braintrust";
import { defineHook } from "eve/hooks";

export default defineHook(braintrustEveHook());
```


js/src/instrumentation/plugins/eve-plugin.ts
```diff
@@ -0,0 +1,1978 @@
...
+      "step.started": (input: EveInstrumentationStepStartedEventInput) => {
+        try {
+          captureEveModelInput(state, input);
+        } catch (error) {
+          debugLogger.warn("Error in Eve LLM input capture:", error);
+        }
+      },
+    },
+    recordInputs: false,
+    recordOutputs: false,
```
AbhiPrasad, line 145: why are these false? Should we make them user configurable?

js/src/instrumentation/plugins/eve-plugin.ts
```diff
@@ -0,0 +1,1978 @@
...
+  }
+}
+
+class EveBridge {
+  constructor(private readonly state: EveStateHandle<EveTraceState>) {}
+
+  private sessionsById = new Map<string, SessionState>();
+  private completedToolKeys = new Set<string>();
+  private toolsByCallKey = new Map<string, ToolState>();
+  private turnsByKey = new Map<string, TurnState>();
```
AbhiPrasad: should we bound? Like use an LRU to manage memory?

js/src/instrumentation/plugins/eve-plugin.ts
```diff
@@ -0,0 +1,1978 @@
...
+      return previous.ordinal;
+    }
+
+    const ordinal = state.stepStarts.filter(
+      (entry) => entry.turnId === event.data.turnId,
+    ).length;
+    this.state.update((current) => ({
+      ...normalizeEveTraceState(current),
+      stepStarts: [
+        ...state.stepStarts,
```
AbhiPrasad: should we use current here? Ditto with some other places where we reference state instead of current.

js/src/instrumentation/plugins/eve-plugin.ts
```diff
@@ -0,0 +1,1978 @@
...
+    event: unknown,
+    ctx: unknown,
+    hookMetadata?: Record<string, unknown>,
+  ): Promise<void> {
+    if (!isEveHandleMessageStreamEvent(event)) {
+      return;
+    }
+    try {
+      await this.handleEvent(event, ctx, hookMetadata);
+      await this.flushInstrumentation();
```
AbhiPrasad: should we only call `flushInstrumentation` for events we handle?

.changeset/eve-instrumentation.md
```diff
@@ -0,0 +1,5 @@
+---
+"braintrust": patch
```
AbhiPrasad: should be a minor!

js/src/instrumentation/plugins/eve-plugin.ts
```diff
@@ -0,0 +1,1978 @@
...
+      updateSpan({
+        exported: this.reference.exported,
+        metrics: { end: this.endTime },
+      });
+    }
+    return this.endTime;
+  }
+}
+
+class EveBridge {
```
AbhiPrasad, line 194: kind feels weird there is no `createTurn` and `endTurn` methods on the bridge here.

js/src/instrumentation/plugins/eve-plugin.ts
```diff
@@ -0,0 +1,1978 @@
...
+): Promise<string> {
+  return deterministicEveId("eve:row:turn", sessionId, turnId);
+}
+
+async function spanIdForStep(
+  sessionId: string,
+  turnId: string,
+  stepIndex: number,
+): Promise<string> {
+  return deterministicEveId("eve:step", sessionId, turnId, String(stepIndex));
```
AbhiPrasad: I think we can have 

```ts
type EveEntityKind = "session" | "step" | "subagent" | "tool" | "turn";

async function generateEveIds(
  kind: EveEntityKind,
  ...parts: string[]
): Promise<{ rowId: string; spanId: string }> {
  const [rowId, spanId] = await Promise.all([
    deterministicEveId(`eve:row:${kind}`, ...parts),
    deterministicEveId(`eve:${kind}`, ...parts),
  ]);
  return { rowId, spanId };
}
```

and take care of most of the use cases, and then we just need `rootSpanIdForSession` and `spanIdForSubagent`?

js/src/instrumentation/plugins/eve-plugin.ts
```diff
@@ -0,0 +1,1978 @@
...
+  output?: unknown;
+  stepsByIndex: Map<number, StepState>;
+};
+
+type ToolState = SpanState & {
+  endedByTurn?: boolean;
+  kind: "subagent" | "tool";
+  stepIndex?: number;
+  turnKey: string;
+};
```
AbhiPrasad, line 72: clanker tells me `ToolState.kind and ToolState.stepIndex were written at every creation site but never read.`

AbhiPrasad (commented):
no orchestrion config? Or iitm setup?

js/src/instrumentation/plugins/eve-plugin.ts
lforst: These are options that `defineInstrumentation()` uses. When they are enabled, the otel data when turned on contains input and output data. Since awe are grabbing the input and output data from hook calls instead of instrumentation it doesn't matter.

js/src/instrumentation/plugins/eve-plugin.ts
lforst: We have handleTurnStarted and handleTurnCompleted and handleTurnFailed. Is that sufficient?

AbhiPrasad: Hmm in general the lifecycle here felt hard for me to follow. I didn't think too deeply about it though.

AbhiPrasad (approved):

PR #2214, opened 2026-07-08
fix(mastra): record time_to_first_token for streaming model spans

viadezo1er:
The Mastra observability exporter now derives time_to_first_token (in
seconds) from a model span's completionStartTime attribute and its start
time — the metric Braintrust surfaces for streaming LLM calls — matching
how our other wrappers emit TTFT. The raw completionStartTime is kept in
span metadata as before, so no consumer-visible field is removed
(earlier releases surfaced it there).

Token usage mapping (prompt/completion/total, cached, cache-creation,
and reasoning tokens) was already at parity with @mastra/braintrust and
is unchanged.

Ports the TTFT half of mastra-ai/mastra#11029 (fixes #10174, #9821,
#9853). Adds js/src/wrappers/mastra.test.ts and regenerates the
mastra-instrumentation e2e snapshots (streaming model spans now carry
time_to_first_token).

### DIFF
--- a/e2e/scenarios/mastra-instrumentation/__snapshots__/mastra-v1260.span-tree.json
+++ b/e2e/scenarios/mastra-instrumentation/__snapshots__/mastra-v1260.span-tree.json
@@ -88,6 +88,7 @@
                   "metric_keys": [
                     "completion_tokens",
                     "prompt_tokens",
+                    "time_to_first_token",
                     "tokens"
                   ]
                 }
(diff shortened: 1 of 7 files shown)


e2e/scenarios/mastra-instrumentation/__snapshots__/mastra-v1260.span-tree.json
```diff
@@ -184,6 +185,7 @@
                   "metric_keys": [
                     "completion_tokens",
                     "prompt_tokens",
+                    "time_to_first_token",
                     "tokens"
```
lforst: I know this is slightly unrelated but it would be sick if we could log the actual metrics here instead of the keys. We fumbled this earlier but would be a nice lil related cleanup.

lforst (approved):

PR #2218, opened 2026-07-09
test: Bump e2e test package versions

lforst:


lforst:
@AbhiPrasad @Qard I know this is a bit ridiculous to review. I honestly did not look at all the outputs myself. This is purely additive and adding/bumping versions, so we have more visibility. We just gotta make sure that the adjustments we did in the scenarios are sane.

AbhiPrasad (approved):
trust the process

Qard (commented):
I have very little trust in any system which results in 200k line PRs. Nobody is going to review that fully, it would take _days_ to do so. 😬

lforst:
> I have very little trust in any system which results in 200k line PRs. Nobody is going to review that fully, it would take _days_ to do so. 😬 

@Qard what's your suggestion to proceed?

Qard:
I won't block merging, because being up-to-date is better than not, I guess. But I definitely think we should rethink the value of this e2e set up. It's way too big to be able to update it with any confidence of correctness. At the least we should probably have a conversation about it at a future team sync to think about what we can do to refactor or replace this with something we can actually _trust_.

lforst:
> I won't block merging, because being up-to-date is better than not, I guess. But I definitely think we should rethink the value of this e2e set up. It's way too big to be able to update it with any confidence of correctness. At the least we should probably have a conversation about it at a future team sync to think about what we can do to refactor or replace this with something we can actually _trust_.

sgtm! This is the best that I could come up with. Feel free to share any thoughts on what you think is optimal.

PR #2226, opened 2026-07-13
feat: Add span origin provenance

Qard:
## Summary

Adds span origin provenance to the TypeScript SDK direct logging and OpenTelemetry paths, including .env.braintrust environment detection.

## Validation

- ./node_modules/.bin/vitest run in integrations/otel-js/otel-v1
- ./node_modules/.bin/vitest run in integrations/otel-js/otel-v2
- aggregate git diff --check

### DIFF
--- a/e2e/scenarios/test-framework-evals-vitest/vitest.runner-v4.config.mts
+++ b/e2e/scenarios/test-framework-evals-vitest/vitest.runner-v4.config.mts
@@ -1,8 +1,23 @@
+import { readFileSync } from "node:fs";
+import * as path from "node:path";
+
 const vitestConfigPackageName =
   process.env.VITEST_CONFIG_PACKAGE_NAME ?? "vitest-v4-latest";
 const { defineConfig } = await import(`${vitestConfigPackageName}/config`);
 
+const repoRoot = process.env.BRAINTRUST_E2E_REPO_ROOT;
+if (!repoRoot) {
+  throw new Error("BRAINTRUST_E2E_REPO_ROOT is not set");
+}
+
+const packageJson = JSON.parse(
+  readFileSync(path.join(repoRoot, "js/package.json"), "utf8"),
+);
+
 export default defineConfig({
+  define: {
+    __BRAINTRUST_SDK_VERSION__: JSON.stringify(packageJson.version),
+  },
   test: {
     hookTimeout: 30_000,
     include: ["runner.case.ts"],
--- a/e2e/scenarios/test-framework-evals-vitest/vitest.runner-v4.config.mts
+++ b/e2e/scenarios/test-framework-evals-vitest/vitest.runner-v4.config.mts
@@ -1,23 +1,8 @@
-import { readFileSync } from "node:fs";
-import * as path from "node:path";
-
 const vitestConfigPackageName =
   process.env.VITEST_CONFIG_PACKAGE_NAME ?? "vitest-v4-latest";
 const { defineConfig } = await import(`${vitestConfigPackageName}/config`);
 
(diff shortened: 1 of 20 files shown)


js/src/span-origin.ts
```diff
@@ -0,0 +1,130 @@
+import iso from "./isomorph";
+import packageJson from "../package.json";
```
lforst: I just checked and this bundles in an odd way pulling in the entire json. Nothing bad but maybe instead of importing package json, we could use the tsup define option?

```
import { defineConfig } from "tsup"; 

export default defineConfig({
  define: {
    __SDK_VERSION__: JSON.stringify("1.2.3"),
  },
});
```

Just requires global definition of `__SDK_VERSION__` somewhere.

Could become relevant if we ever publish for browsers bundle size wise because we don't wanna bundle the entire package json.

js/src/span-origin.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (!value) return undefined;
+  const normalized = value.toLowerCase();
+  if (normalized === "production" || normalized === "staging") {
+    return { type: "server", name: normalized };
+  }
+  if (normalized === "development" || normalized === "local") {
+    return { type: "local", name: normalized };
+  }
+  return undefined;
+}
```
lforst, line 130: Same thought as my (sadly late) feedback here just for Node_env instead: https://github.com/braintrustdata/braintrust-sdk-python/pull/570#discussion_r3593673202

e2e/scenarios/test-framework-evals-vitest/vitest.runner-v4.config.mts
```diff
@@ -1,8 +1,23 @@
...
+}
+
+const packageJson = JSON.parse(
+  readFileSync(path.join(repoRoot, "js/package.json"), "utf8"),
+);
+
 export default defineConfig({
+  define: {
+    __BRAINTRUST_SDK_VERSION__: JSON.stringify(packageJson.version),
+  },
```
lforst: This almost looks like your agent went overboard a bit. Not sure we need to add this to any vitest config in our repo.

lforst: Ah actually, sorry I realize what's going on. The tests would fail because we haven't replaced `__BRAINTRUST_SDK_VERSION__` yet.

Can we maybe define it as so: `const SDK_VERSION = typeof __BRAINTRUST_SDK_VERSION__ !== 'undefined' ? __BRAINTRUST_SDK_VERSION__ : '0.0.0'` -> that way we would still be properly replacing but have a consistent thing in tests.

js/src/span-origin.ts
```diff
@@ -0,0 +1,131 @@
...
+};
+
+type SpanOrigin = {
+  name: string;
+  version: string;
+  instrumentation: { name: string };
+  environment?: SpanOriginEnvironment;
+};
+
+const SDK_VERSION = __BRAINTRUST_SDK_VERSION__;
```
lforst:
```suggestion
const SDK_VERSION = typeof __BRAINTRUST_SDK_VERSION__ !== 'undefined' ? __BRAINTRUST_SDK_VERSION__ : '0.0.0';
```

lforst: would allow us to get rid of the defines in all the vitest configs

e2e/scenarios/test-framework-evals-vitest/vitest.runner-v4.config.mts
Qard: Sure, I think that should work. 🙂

lforst (approved):

PR #2230, opened 2026-07-14
feat: Track eve turns as individual traces

lforst:
Currently we attach turns to existing traces as a "session". Going forward we generally want to have a trace per turn.

### DIFF
--- a/.changeset/heavy-horses-tickle.md
+++ b/.changeset/heavy-horses-tickle.md
@@ -1,0 +1,5 @@
+---
+"braintrust": patch
+---
+
+feat: Track eve turns as individual traces
--- a/.changeset/heavy-horses-tickle.md
+++ b/.changeset/heavy-horses-tickle.md
@@ -1,5 +1,5 @@
 ---
-"braintrust": patch
+"braintrust": minor
 ---
 
-feat: Track eve turns as individual traces
+feat: Track Eve turns as individual traces and record Eve session IDs on all spans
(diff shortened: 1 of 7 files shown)


.changeset/heavy-horses-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+"braintrust": patch
```
AbhiPrasad:
```suggestion
"braintrust": minor
```

I noticed my clanker loves to do this too.

AbhiPrasad (commented):
Do we not want to attach session id then?

.changeset/heavy-horses-tickle.md
lforst: That was actually me haha - whenever it has a generated name, I ran the changeset command myself.

I thought this was no API change and soon enough after introduction so that we can act like this is a patch - not sure? What would be the arguments for minor? I guess any sort of scorers/processors that depend on this shape. I'll change it.

lforst:
> Do we not want to attach session id then?

Can add (albeit it is like very useless imo cause it is very much not exposed - only useful for cross referencing I guess)

AbhiPrasad:
if we attach session id, we can run trace group scorers via aggregating on session id. It can also be used for to make [topic facets](https://www.braintrust.dev/docs/observe/topics/manage#group-traces-into-conversations) better.

AbhiPrasad (approved):

PR #2233, opened 2026-07-14
fix(eve): Don't report fallback model when dynamic model is used

lforst:
Eve has the concept of "dynamic models" and "fallback models". Fallback models are used when the chosen dynamic model doesn't work. Currently eve always reports the fallback model with a `dynamic:` prefix when a dynamic model is used. We should not pick this up because it causes completely wrong pricing and model data.


Qard (approved):
If I'm understanding correctly, this places the model metadata on every child span and explicitly _not_ the root? Does our backend understand this currently for cost computation? If so, LGTM.

lforst:
> If I'm understanding correctly, this places the model metadata on every child span and explicitly _not_ the root? Does our backend understand this currently for cost computation? If so, LGTM.

yup as per my last knowledge, cost is aggregated through llm spans

PR #2236, opened 2026-07-15
feat(ai-sdk): Add `HarnessAgent` tracing support

lforst:
- Adds HarnessAgent tracing support which is actually quite simple
- The added instrumentation basically just adds spans around the HarnessAgent top level methods: generate, stream, continueGenerate, or continueStream
- When these apis are called, we inject `options: { telemetry: [braintrustAISDKTelemetry()] }`, reusing our ai sdk handler, which will make the harness agent supply telemetry events to it that we trace


js/src/wrappers/ai-sdk/ai-sdk.ts
```diff
@@ -166,6 +173,48 @@ export const wrapAgentClass = (
         get(instanceTarget, prop, instanceReceiver) {
           const original = Reflect.get(instanceTarget, prop, instanceTarget);
 
+          if (
+            instanceTarget.constructor.name === "HarnessAgent" &&
```
AbhiPrasad: is this robust enough? What if the user extends the `HarnessAgent` class?

AbhiPrasad (approved):

PR #2257, opened 2026-07-21
feat: Add `agents` instrumentation

lforst:


e2e/scenarios/cloudflare-agents-instrumentation/scenario.ts
```diff
@@ -0,0 +1,168 @@
...
+    stdio: ["ignore", "pipe", "pipe"],
+  });
+  const output = captureOutput(build);
+  const [exitCode] = (await once(build, "close")) as [number | null];
+  if (exitCode !== 0) {
+    throw new Error(`Vite build failed with code ${exitCode}\n${output()}`);
+  }
+}
+
+async function getFreePort(): Promise<number> {
```
Qard, line 102: minor nit: this approach could be flaky because you're cunning multiple of these concurrently and the port you claim and release to identify as available could then be take by another instance before this one can use it.

Qard (approved):

e2e/scenarios/cloudflare-agents-instrumentation/scenario.ts
lforst: True. This is annoying to work around. I'll see how flakey it is and if it becomes an issue I'll fix.

PR #2265, opened 2026-07-22
deps: Bump deps

lforst:
bumps deps which fixes somedecurity alerts that should not have any runtime exploitation paths


Qard (approved):
nit: Why did we gain the body-parser and brace-expansion dependencies in all these? The brace-expansion dependencies were previously only on the otel-js package and the body-parser dependency was not present at all. 🤔

PR #2318, opened 2026-07-31
feat(vitest-evals): Add `meta.eval.input`

aislinnnnn:
## What this PR does

Adds `meta.eval.input` to the `vitest-evals` reporter, as a direct passthrough for an eval row's `input` field — the same way `meta.eval.output` already works.

## Why

Today the eval-level root span's `input` can *only* be populated via `harness.run.session.messages`, specifically the first entry with `role: "user"` (`firstUserMessageContent()`). That's a reasonable default for conversational harnesses, but it forces any harness whose input isn't a conversation (e.g. a structured payload, a single scalar) to fabricate a synthetic `{ role: "user", content: ... }` message just to get anything into `input` at all.

This is inconsistent with the rest of this same file:
- `meta.eval.output` already supports a direct passthrough (`input.output !== undefined ? { output: input.output } : {}` in `readEvalMeta`) — there's no equivalent for `input`.
- `logToolCallSpans` in this file sets `event: { input: call.arguments }` directly, no message-shape required.
- The classic `Eval()` framework (`js/src/framework.ts`) does `event: { input: datum.input }` — a direct, arbitrary passthrough, which is the norm for the rest of the SDK.

## Changes

- `EvalMeta` gains an `input?: unknown` field, alongside the existing `output?: unknown`.
- `readEvalMeta` reads `input.input` the same way it already reads `input.output`.
- `logEvalTest` computes `const input = meta.eval?.input ?? firstUserMessageContent(run);` — `meta.eval.input` takes precedence when present, falling back to the existing session-derived behavior when it's not. Fully backward compatible: nothing changes for consumers that only set `harness.run.session.messages`.
- Adds a test covering both the direct-passthrough case and the fallback case.

## Testing

- `pnpm vitest run src/wrappers/vitest-evals/reporter.test.ts` — all 6 tests pass (5 existing + 1 new).
- `pnpm run fix:formatting` — clean.
- Added a changeset (`minor`, per the existing convention for new fields).

### DIFF
--- a/.changeset/vitest-evals-direct-input.md
+++ b/.changeset/vitest-evals-direct-input.md
@@ -2,7 +2,7 @@
 "braintrust": minor
 ---
 
-feat(vitest-evals): support a direct `input` on eval task meta, mirroring `output`
+feat: `BraintrustVitestEvalsReporter` accepts a direct `input` on eval task meta
 
 Previously the only way to populate an eval row's `input` field was via
 `harness.run.session.messages` (the first message with `role: "user"`), which
--- a/.changeset/vitest-evals-direct-input.md
+++ b/.changeset/vitest-evals-direct-input.md
@@ -2,12 +2,4 @@
 "braintrust": minor
 ---
 
-feat: `BraintrustVitestEvalsReporter` accepts a direct `input` on eval task meta
-
-Previously the only way to populate an eval row's `input` field was via
-`harness.run.session.messages` (the first message with `role: "user"`), which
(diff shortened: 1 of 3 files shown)


AbhiPrasad:
thanks for the PR @aislinnnnn. We'll take a look at this on monday!

aislinnnnn:
sounds good, thanks @AbhiPrasad !

lforst (approved):

lforst:
Thank you!

PR #2333, opened 2026-08-03
feat: Add instrumentation for `anthropic.beta.sessions`

lforst:
Adds instrumentation for `anthropic.beta.sessions`.

The anthropic API is either `sessions.event.stream()` or `sessions.threads.events.stream()`. You are reading events from something that is happening on anthropic infra - that's why I didn't want to make these apis auto instrumented, instead, the user should be manually collecting the streams with `collectAnthropicSession()`:

```ts
import Anthropic from "@anthropic-ai/sdk";
import {
  collectAnthropicSession,
  wrapAnthropic,
} from "braintrust";

const anthropic = wrapAnthropic(new Anthropic());

const stream = collectAnthropicSession(
  await anthropic.beta.sessions.events.stream("session-id", {
    event_deltas: ["agent.message"],
  }),
);

for await (const event of stream) {
  console.log(event);
}
```


AbhiPrasad (commented):
Ah interesting. I decided to auto-instrument this in python iirc. Maybe I should revert that there too. If we decide this is how we want the sdks to work, I'd add a spec doc change for this.

lforst:
> Ah interesting. I decided to auto-instrument this in python iirc. Maybe I should revert that there too. If we decide this is how we want the sdks to work, I'd add a spec doc change for this.

@AbhiPrasad yeah the main reason I decided to do it like this is that you may have braintrust tracing be "unwanted collateral damage" if we auto-instrument such APIs. I don't think there is a clear right or wrong here, but since getting the stream is mostly a manual decision someone has to make, I thought it would also be fine to only trace manually. Like I would be kinda surprised if I had something like analytics on the sessions, and suddenly braintrust traces my analytics service. Or maybe even emits duplicates traces if I retrieve the session events from multiple places. Makes sense?

AbhiPrasad:
I think you're right here. I'm going to update the python SDK accordingly.

AbhiPrasad (approved):

PR #2370, opened 2026-08-17
feat: Add DeepSeek Harness plugin

lforst:
Adds a `@braintrust/deepseek-harness` package.

The package can be installe as a plugin via the deepseek harness cli (seems to be the canonical way for now) to trace everything that the deepseek harness does.

Reports each user turn as a separate trace with nested spans for model calls, tool executions, and delegated child sessions. It captures model inputs and outputs, token usage, reasoning, tool calls, errors, and relevant Harness metadata.

## Installation

```bash
dsh plugin --profile web add @braintrust/deepseek-harness
```

Optional settings can be configured through the Harness plugin UI or `cordis.patch.yml`:

```yaml
- id: braintrust
  config:
    apiKey: <your-api-key>
    projectName: DeepSeek Harness
```

The API key is declared as a secret setting. When omitted, the plugin uses `BRAINTRUST_API_KEY`.

### DIFF
--- a/integrations/deepseek-harness/package.json
+++ b/integrations/deepseek-harness/package.json
@@ -1,0 +1,68 @@
+{
+  "name": "@braintrust/deepseek-harness",
+  "version": "0.1.0",
+  "description": "Braintrust tracing plugin for DeepSeek Harness",
+  "main": "./dist/index.js",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    "./package.json": "./package.json",
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs",
+      "module": "./dist/index.mjs",
+      "require": "./dist/index.js"
+    },
+    "./cordis.patch.yml": "./cordis.patch.yml"
+  },
+  "files": [
+    "dist/**/*",
+    "cordis.patch.yml",
+    "README.md"
+  ],
+  "scripts": {
+    "build": "tsup",
+    "check:typings": "tsc --noEmit",
+    "watch": "tsup --watch",
+    "clean": "rm -r dist/*",
+    "test": "vitest run",
+    "yalc:publish": "yalc publish"
+  },
+  "dsh": {
+    "bundle": {
+      "patch": "./cordis.patch.yml"
+    }
+  },
+  "license": "MIT",
+  "peerDependencies": {
+    "@deepseek-ai/cordis": ">=4.0.1",
+    "@deepseek-ai/dsh-llm": ">=0.1.0-rc.6 <0.2.0",
+    "@deepseek-ai/dsh-session": ">=0.1.0-rc.6 <0.2.0",
+    "@deepseek-ai/dsh-tools": ">=0.1.0-rc.6 <0.2.0",
+    "@deepseek-ai/schemastery": ">=3.18.1",
+    "braintrust": ">=3.0.0"
+  },
+  "devDependencies": {
+    "@deepseek-ai/cordis": "4.0.1",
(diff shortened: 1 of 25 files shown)


integrations/deepseek-harness/package.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "@braintrust/deepseek-harness",
```
AbhiPrasad, line 2: why are we publishing a new package?

lforst: It's deepseek harnesses recommended way of distributing things. You're supposed to run `dsh plugin --profile web add @braintrust/deepseek-harness` and it will install the package from npm.

AbhiPrasad: Ah would be useful to have in PR description!

integrations/deepseek-harness/src/index.ts
```diff
@@ -0,0 +1,709 @@
+import { AsyncLocalStorage } from "node:async_hooks";
+import { randomUUID } from "node:crypto";
+import type { Context } from "@deepseek-ai/cordis";
+import type {} from "@deepseek-ai/dsh-llm";
+import type {} from "@deepseek-ai/dsh-session";
+import type {} from "@deepseek-ai/dsh-tools";
```
AbhiPrasad, line 7: why these blank imports?

integrations/deepseek-harness/package.json
lforst: sorry very true!

integrations/deepseek-harness/src/index.ts
```diff
@@ -0,0 +1,709 @@
...
+          : {}),
+        ...(options.maxTokens !== undefined
+          ? { max_tokens: options.maxTokens }
+          : {}),
+        ...(options.stop !== undefined ? { stop: options.stop } : {}),
+        ...(options.tools !== undefined
+          ? {
+              tools: options.tools.map((tool) => ({
+                type: "function",
+                function: tool,
```
AbhiPrasad: doesn't this blindly forward the tool? Is that okay?

integrations/deepseek-harness/src/index.ts
```diff
@@ -0,0 +1,709 @@
...
+        };
+      case "tool-result":
+        return {
+          type: "tool_result",
+          tool_call_id: block.toolCallId,
+          content: normalizeBlocks(block.content ?? []),
+          ...(block.isError ? { is_error: true } : {}),
+        };
+      case "image":
+        return { type: "image", attachment: block.attachment };
```
AbhiPrasad: should we make into a braintrust attachment?

integrations/deepseek-harness/package.json
lforst: updated

integrations/deepseek-harness/src/index.ts
```diff
@@ -0,0 +1,709 @@
...
+}
+
+export const Config: z<Config> = z.object({
+  apiKey: z.string().role("secret"),
+  projectName: z.string().default(DEFAULT_PROJECT_NAME),
+  orgName: z.string(),
+  appUrl: z.string(),
+});
+
+interface BraintrustEvent {
```
AbhiPrasad: can't we get these types from the `braintrust` package?

integrations/deepseek-harness/src/index.ts
lforst: they're loading .d.ts files which augment the cordis Events

integrations/deepseek-harness/src/index.ts
lforst: good catch actually, we should narrow down

AbhiPrasad (approved):

PR #2392, opened 2026-08-20
fix(pi-coding-agent): support streamFn -> streamFunction rename

justcodebruh:
## Summary
- `@earendil-works/pi-coding-agent` renamed `agent.streamFn` to `agent.streamFunction` starting in v0.81.0.
- `wrapPiCodingAgentSDK` and the Pi Coding Agent auto-instrumentation plugin only detected/patched `streamFn`, so on pi >= 0.81  no spans were emitted.

- Resolves whichever property actually exists on the agent (`streamFn` or `streamFunction`) and patches that one, so both pre- and post-rename shapes work.

## Context
Surfaced via the customer, Unit, running `@earendil-works/pi-coding-agent@0.84.2` with `braintrust@3.27.0` no logs were reaching Braintrust at all. 

Verified end-to-end locally against pi 0.84.2 with a one-off patched build before opening this PR.

## Test plan
- [ ] Added a regression test (`patches agent.streamFunction when streamFn is absent (pi >= 0.81)`) in `pi-coding-agent-plugin.test.ts` covering the new shape.

### DIFF
--- a/js/src/instrumentation/plugins/pi-coding-agent-plugin.test.ts
+++ b/js/src/instrumentation/plugins/pi-coding-agent-plugin.test.ts
@@ -427,6 +427,41 @@ describe("PiCodingAgentPlugin", () => {
     await result;
     expect(taskSpan?.end).toHaveBeenCalledTimes(1);
   });
+
+  it("patches agent.streamFunction when streamFn is absent (pi >= 0.81)", async () => {
+    const interceptor = promptInterceptor(enablePlugin(plugins));
+    const finalMessage = makeAssistantMessage("done");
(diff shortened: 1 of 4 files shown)


js/src/instrumentation/plugins/pi-coding-agent-plugin.test.ts
justcodebruh: I am not confident at all in this test and would love a review on how we should test this at all.

lforst:
Superseded by https://github.com/braintrustdata/braintrust-sdk-javascript/pull/2400

PR #2410, opened 2026-08-28
ref: Remove deprecated packages

lforst:


AbhiPrasad (commented):
shouldn't we only do this with the next major?

PR #2470, opened 2026-09-11
fix(eve): Use `any` as return type to ensure eve version compat

lforst:


AbhiPrasad (approved):
annoying but what can you do

